PR #145, opened 2014-10-03
Handle empty array when encoding URL parameters

stuntdouble:
Alamofire was not including parameters whose value was an empty array in the URL request. I've added a check to handle an empty array, as well as two tests to ensure consistent behavior between URL and JSON parameter encoding.


asasdasasd:
here is what chrome does,  `bike=1231231&content=121212&images=` in post request form data, `images` is an empty array, but chrome still pass it to the server

Issue #141, opened 2014-10-01
Resume data not exposed in cancel API


zuschlag:
Although Alamofire provides a download api with a resume data block for resuming a previous canceled download, it seems that an API for obtaining the data block is missing. The resume data block provided by cancelByProducingResumeData is currently stored in TaskDelegate which is considering private to Request. 

A possible solution could be to extend the cancel API with a completion handler providing the caller with the generated resume data block:

    public func cancel(completionHandler: (NSData!) -> Void) {
        if let downloadDelegate = self.delegate as? DownloadTaskDelegate {
            downloadDelegate.downloadTask.cancelByProducingResumeData { (data) in
                downloadDelegate.resumeData = data
                completionHandler(data)
            }
        } else {
            self.task.cancel()         
        }
    }

cnoon:
Hey guys,

You can actually do this quite easily without modifying Alamofire at all. Your response serializer will receive the resume data that you need to resume the download. Here's a quick little example that I put together.

```swift
import Alamofire

class Downloader {
    var cancelledData: NSData?
    var downloadRequest: Request?
    
    func startDownload() {
        let request = Alamofire.download(URLRequest(.GET, "http://some_big_file"), Request.suggestedDownloadDestination())
        request.response { request, response, data, error in
            if let error = error {
                if error.code == NSURLErrorCancelled {
                    self.cancelledData = data as? NSData
                } else {
                    println("Failed to download file: \(response) \(error)")
                }
            } else {
                println("Successfully downloaded file: \(response)")
            }
        }
        
        self.downloadRequest = request
    }
    
    func cancelDownload() {
        self.downloadRequest?.cancel()
    }
    
    func resumeDownload() {
        if let cancelledData = self.cancelledData {
            let request = Alamofire.download(resumeData: cancelledData, Request.suggestedDownloadDestination())
            
            request.response { request, response, data, error in
                if let error = error {
                    println("Failed to resume file download: \(response) \(error)")
                } else {
                    println("Successfully downloaded file: \(response)")
                }
            }
        }
    }
}
```

I hope that helps!

vittoriom:
Isn't there a way to get the path to the temporary file so that on the next run we can read the <code>NSData</code> from there and easily resume the download? As far as I can see, the path is only provided after the download is done, to move it to a permanent folder.

vittoriom:
[...]

PR #190, opened 2014-10-22
[Issue #141] Allow obtaining of resume data

ivanmoskalev:
A fix for Issue #141

There are some inconsistencies with resume data closure parameter being exposed in non-download request. See issue thread for more detail.

The thing is that we have to be sure we will be OK with such interface for the first version at least, because if we decide to change it later, it will likely be a breaking change.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -677,10 +677,13 @@ public class Request {
     /**
         Cancels the request.
     */
-    public func cancel() {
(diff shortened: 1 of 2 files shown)


cnoon:
Thank you for your pull request @ivanmoskalev. Apologies for not getting back to you sooner. You can actually get the resume data from the code sample I posted in #141. Also, we're going to add better support for resume data in the 1.3.0 release. Please follow the progress of #141 for future updates.

PR #228, opened 2014-11-21
Carthage support

ishkawa:
According to their documentation, Carthage needs shared scheme to build dynamic framework.
> Carthage will only build Xcode schemes that are shared from your .xcodeproj. You can see if all of your intended schemes build successfully by running carthage build --no-skip-current, then checking the Carthage.build folder.

Alamofire has scheme that builds dynamic framework, but it is not shared. So I made it shared.
I tested Carthage builds Alamofire.framework correctly on the [forked repository](https://github.com/ishkawa/Alamofire).

NOTE: new tag is required to notify update of this repository to Carthage.


ianp:
+1

esummers:
+1

mattt:
Thanks for this, @ishkawa. Apologies for the delay in getting this integrated.

PR #258, opened 2014-12-14
Fixed Xcode error from documentation example

philipheinser:


mattt:
Thanks, @philipheinser!

PR #278, opened 2014-12-29
Change internal progress state before calling delegates

astrabot:
To maintain logical data integrity, the progress state of Request object should be changed before it will be available for delegates.


mattt:
Looks good. Thanks so much for this, @astrabot.

PR #287, opened 2015-01-04
bump podspec to 1.1.3

rynecheow:
Bump podspec Alamofire version to latest release

### DIFF
--- a/Alamofire.podspec
+++ b/Alamofire.podspec
@@ -1,12 +1,12 @@
 Pod::Spec.new do |s|
   s.name = 'Alamofire'
-  s.version = '1.1.0'
+  s.version = '1.1.3'
   s.license = 'MIT'
   s.summary = 'Elegant HTTP Networking in Swift'
   s.homepage = 'https://github.com/Alamofire/Alamofire'
   s.social_media_url = 'http://twitter.com/mattt'
   s.authors = { 'Mattt Thompson' => 'm@mattt.me' }
-  s.source = { :git => 'https://github.com/Alamofire/Alamofire.git', :tag => '1.1.0' }
+  s.source = { :git => 'https://github.com/Alamofire/Alamofire.git', :tag => '1.1.3' }
 
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'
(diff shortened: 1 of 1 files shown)


Alamofire.podspec
```diff
@@ -1,12 +1,12 @@
...
   s.name = 'Alamofire'
-  s.version = '1.1.0'
+  s.version = '1.1.3'
   s.license = 'MIT'
   s.summary = 'Elegant HTTP Networking in Swift'
   s.homepage = 'https://github.com/Alamofire/Alamofire'
   s.social_media_url = 'http://twitter.com/mattt'
   s.authors = { 'Mattt Thompson' => 'm@mattt.me' }
-  s.source = { :git => 'https://github.com/Alamofire/Alamofire.git', :tag => '1.1.0' }
+  s.source = { :git => 'https://github.com/Alamofire/Alamofire.git', :tag => '1.1.3' }
```
kylef: We can make this `:tag => s.version` so it doesn't need to be updated in two places each time.

iasandcb: s.ios.deployment_target = '7.0'

is also needed.

kylef: Minimum deployment target for iOS using Swift in CocoaPods is iOS 8.0, so it should stay as is.

iasandcb: @kylef could you let me know a link (of a web page or a blog) on specifying the requirement (iOS 8)?

Thanks.

kylef: iOS only supports frameworks on iOS 8 and onwards, you cannot use Swift with static libraries. Therefore it's only possible to use Alamofire with iOS 8 and onwards with CocoaPods.

If you was to lint the pod spec with iOS 7, the linker itself will mention this:

```bash
$ pod lib lint

 -> URITemplate (1.0.1)
    - NOTE  |  ld: warning: embedded dylibs/frameworks only run on iOS 8 or later
```

This will fail lint on the final release of CocoaPods 0.36 (https://github.com/CocoaPods/CocoaPods/issues/2963).

mattt:
Equivalent changes have since been applied. Thanks for the PR, @rynecheow.

PR #212, opened 2014-11-06
OS X target support

carloslozano:
* Added OSX target - generate framework for OSX 10.10 target.
* Tweaked DownloadTests.swift to avoid deleting entire OSX Documents directory during tearDown()
* OSX Tests passing with the exception of AlamofireDownloadResponseTestCase.testDownloadRequest()
   * Downloaded file is saved with **.json** extension instead of expected **\\(numberOfLines) lastPathComponent**. Dunno... anyone?

Please excuse the two previous pull requests I closed them myself... :blush:

PR #293, opened 2015-01-08
Add OS X target for Carthage

mk:
This is my attempt at #212 and adds a new shared target for OS X 'Alamofire OS X' which uses the same Info.plist as the iOS target. I've renamed iOS target to 'Alamofire iOS'. I've verified in an app that it works on both platforms.

Given the following `Cartfile`:

    github "mk/Alamofire"

Running `carthage bootstrap` against this outputs:

    *** Checking out Alamofire at "1.1.6"
    *** xcodebuild output can be found in /var/folders/lk/q34wnz8j0h93pc7rlck4stnc0000gn/T/carthage-xcodebuild.it0PZY.log
    *** Building scheme "Alamofire iOS" in Alamofire.xcworkspace
    *** Building scheme "Alamofire OSX" in Alamofire.xcworkspace

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


mattt:
Awesome. Excellent work, @mk. I really appreciate you taking the time to do this.

PR #290, opened 2015-01-07
Created an image response serializer

cnoon:
Hey @mattt,

We need this functionality for a networking library we're building internally. The more I worked through it, I began to think this made more sense to live directly inside `Alamofire` rather than be an outside extension. I'm very interested to see if you agree. The only downside is that UIKit ends up being a dependency and I'm not sure how you feel about that. Either way, I based the inflation logic off of AFNetworking and ported it to Swift. I have it built inside a test app and it works flawlessly. I've also profiled it to make sure the inflation logic is running on the proper Alamofire queue.

Cheers,
Christian

PR #294, opened 2015-01-09
Added UIImageView extension to support remote and placeholder images

cnoon:
NOTE: This PR first requires PR #290 to get merged.

I built a `UIImageView` extension with a supported `ImageCache` based off the `AFNetworking` `UIImageView` category. There are certainly some modifications, but the general idea and interface is the same. In Swift, we can't support a custom image response serializer unless we wrap the object inside some `NSObject` subclass in order to attach it to the `UIImageView` as an associated property. Curious to see what your thoughts are there.

I also made the assumption that you would like to start splitting `UIKit` extensions into separate Swift files and placing them in the iOS target. I could certainly modify this approach and wrap the entire set of logic in `#if os(iOS)` blocks if you would rather use that approach. This would also allow us to share the `ImageCache` logic with `OSX` if you wanted to. Had to start somewhere...

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


cnoon:
Just as a heads up, I've started working on a separate library `AlamofireImage` to contain all this logic. If you're interested, see #290 for more details.

cnoon:
Closing out this PR since it won't ever be merged into Alamofire core. Please see #290 for more details.

PR #298, opened 2015-01-13
Fixed ios deployment target (reflecting the README)

Reflejo:
I know embedded frameworks only work on ios 8+ but I think the podspec should reflect the minimum needed target. 

More info: In my case, I'm targeting ios7 and I'm still using embedded frameworks since it works on dev. For releasing I'll probably create a script later that moves the .swift files from the embedded framework to the target project.

### DIFF
--- a/Alamofire.podspec
+++ b/Alamofire.podspec
@@ -8,7 +8,7 @@ Pod::Spec.new do |s|
   s.authors = { 'Mattt Thompson' => 'm@mattt.me' }
   s.source = { :git => 'https://github.com/Alamofire/Alamofire.git', :tag => s.version }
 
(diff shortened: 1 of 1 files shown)


kylef:
@Reflejo This will cause the pod spec to fail validation with CocoaPods 0.36.0. See https://github.com/CocoaPods/CocoaPods/issues/2963

mattt:
Thanks, @kylef.

Reflejo:
@kylef Right, I guess it makes sense for now. Until there is another way on cocoapods to allow private builds targeting iOS7. For anyone reading this: [This is relevant](https://github.com/CocoaPods/CocoaPods/issues/1241)

PR #302, opened 2015-01-14
Add test target that runs tests on OS X

mk:
Hey Mattt,

Following #293 I've created a second test target that runs the test on OS X as well.

The `testDownloadRequest` failed on OS X because `NSURLRequest.suggestedFilename` returns `100` and `100.json` on iOS and OS X respectively.

Note that running the test on OS X creates a `100.json` file in `~/Documents` since its not running sandboxed which is later deleted in `tearDown()`, not sure if this is a problem in practice.


mk:
The tests do pass locally and should pass on travis once Xcode 6.1.1 is available there.

mk:
I can't get the OS X tests to pass on travis, removing them again for now to show that 640f8921fcef1afdd58744ac6544c8d8b90b399e actually fixes the travis build for the iOS test target.

mattt:
Thanks for your work on this, @mk. I cherry-picked 855e9c1 and setup my own OS X target, which is now on master.

mk:
Great, thank you!

PR #308, opened 2015-01-18
allow using custom dispatch queue for response serializers

max-potapov:
to avoid extra main-thread switching on each request


mattt:
See #215.

Issue #194, opened 2014-10-24
Background downloads


robertmryan:
Digging around in Alamofire, trying to get background sessions to work, there seem to be a couple of issues: For background downloads to work, it must support the situation in which the app has been terminated while background tasks proceed in the background, but where the app is fired up again when all the background tasks are done. Unfortunately, `didFinishDownloadingToURL` appears to:

 - look for the task delegate and if found, call the task's `didFinishDownloadingToURL` (which, if found, calls the task's `downloadTaskDidFinishDownloadingToURL` closure); and 

 - regardless of whether it found it or not, it then calls the session delegate's `downloadTaskDidFinishDownloadingToURL` closure.

So, there are a couple of issues here:

1. The `Manager` object doesn't expose the `delegate`, much less the `SessionDelegate` class itself. I see no way, when reinstantiating the session manager (presumably in the app delegate's `handleEventsForBackgroundURLSession`) to specify the session's `downloadTaskDidFinishDownloadingToURL` or the `sessionDidFinishEventsForBackgroundURLSession` closures. We need some way to tell the session delegate what to do upon these two events in the absence of task-level delegates.

 I exposed these two closures at the `Manager` level, and that seems to remedy the issue.

2. There would appear to be a similar issue with `sessionDidReceiveChallenge`: You have no assurances that when the app is fired up again (having possibly previously been terminated) that the request will be found, and thus the credentials provided during the request might be lost. And the session-level `sessionDidReceiveChallenge` closure is specified at the `delegate` object (which is private), so I see no way to set this.

 I created a variable at the `Manager` level that exposed this closure, and it now works.

3. The task's `downloadTaskDidFinishDownloadingToURL` closure asks us to return a `NSURL` (because the delegate method will kindly move the file for us). Great. But the session's `downloadTaskDidFinishDownloadingToURL` also asks us to return a `NSURL`, but the closure's return value does not appear to be used. Either the session delegate method should use this return value to move the file for us (and by the way, it should presumably only call this if it didn't find the task-level closure of the same name) or this session level delegate should not ask us to return a `NSURL` so that we're not mislead as to its intent and we understand that we have to move the file ourselves.

 The session level closure should also probably provide us some indication as to whether it successfully moved the file already as a result of the task-level closure (if one was found). Right now, we don't have a reliable way of knowing whether the task level delegate was found and succeeded or not.

[...]

Issue #314, opened 2015-01-25
Follow HTTP redirects (or not) per request


jhersh:
I know this isn't a support forum, and I did unsuccessfully [ask this on StackOverflow](http://stackoverflow.com/questions/27759420/alamofire-follow-http-redirects-or-not); I'm posting here because I'm wondering if this is a small oversight.

I'm trying to configure Alamofire to follow redirects (or not) on a per-request basis.

Alamofire has a private internal class `SessionDelegate` which serves as the `NSURLSessionTaskDelegate` for the current URL session. `SessionDelegate` does implement the relevant delegate method, `URLSession(session:, task:, willPerformHTTPRedirection response:, request:, completionHandler:)` which is exactly what I want. Even better, the delegate's implementation consults a custom variable closure named `taskWillPerformHTTPRedirection` to determine how to handle the redirect - again, exactly what I want! And as far as I can tell, that closure is always nil by default -- it is not assigned to internally by Alamofire -- which suggests that it is intended to let the user assign a closure to it.

I cannot access the `SessionDelegate` class to assign a closure to its `taskWillPerformHTTPRedirection` variable. It is a private class and it is not visible to my Swift files. What is the proper means of configuring an Alamofire request to (not) follow redirects?

developerJohn:
+1

jhersh:
This is resolved by #317 and #424.

PR #317, opened 2015-01-28
Support for Background Sessions as well as SessionDelegate override closures

cnoon:
Hey @mattt,

I've been working on a library built ontop of Alamofire to handle oauth refresh as well as some really interesting data validation and parsing. Overall we have been extremely pleased with Alamofire. The design is fantastic and it has truly been a pleasure to get to work with it. I can't wait to see how this evolves over the coming months!

The latest feature I've been working on is background session support. When digging into how it should integrate into Alamofire, I've hit some roadblocks. A really great discussion took place a couple months ago in #194 which documents some of the challenges. After reading through that thread multiple times, I started down the path of adding support for background sessions in Alamofire.

The initial problem I ran into was that I could not set the `SessionDelegate` `sessionDidFinishEventsForBackgroundURLSession:` closure. I couldn't even get the `delegate` property on the `SessionManager`. After digging in deeper, I realized that all the session delegate closures you have added are basically placeholders. None of them are being set to do anything. All the implementation is built into the task delegates. Once I understood the challenges, I decided to crack the nut wide open to "attempt" to understand what your placeholder closures "could" be used for.

#### Internal Session Delegate

The first thing that had to change was to make the `SessionDelegate` public along with the SessionManager `delegate` property. For the future iterations of Alamofire, I'm not sure you can possibly keep this internal. The reason is that by locking down those delegate callbacks, we can't get in to override the default behavior, even if we have to. While I think it is awesome to get so much default behavior for free, it is certainly necessary in certain cases for us to be able to access the session delegate callbacks (background session finishing) through mechanisms other than the response Serializers.

#### Overriding Session Delegate Closures

The next step was to make all the `SessionDelegate` closures public as well so we can actually set them. That part was easy, but the more I dug into how those closures were used, the more problematic it became. I realized that the logic for calling those closures was only partially implemented. Not all of the closures were being called in the delegate methods.

What makes the most sense to me is to have the closures be public allowing users to override any of the `SessionDelegate` methods if they need to. It also seems that the `SessionDelegate` should always prioritize the closures over the internal Alamofire implementation. This is why I make sure to always call the closure before the internal implementation if it is not nil.

There were also some small bugs in the closure declarations (81c69be0, bbf16e17), a missing closure (8b6af2fc) and a missing data task delegate call (dfab3f3f) that I fixed up.

#### Background Completion Handler

The last change that I made (582267b9) was to actually add a `backgroundCompletionHandler` closure property to the `SessionManager`. I'm still a bit on the fence about this change. It's certainly not necessary, but is a nice-to-have. By adding the property, we can automatically set up the `SessionDelegate` `sessionDidFinishEventsForBackgroundURLSession ` closure to call the handler if it exists.

```swift
required public init(configuration: NSURLSessionConfiguration? = nil) {
    self.delegate = SessionDelegate()
    self.session = NSURLSession(configuration: configuration, delegate: delegate, delegateQueue: nil)

    self.delegate.sessionDidFinishEventsForBackgroundURLSession = { [weak self] session in
        if let strongSelf = self {
            strongSelf.backgroundCompletionHandler?()
        }
    }
}
```

The only issue with the default implementation is that it doesn't give you the chance to handle your events before the handler is called. To do that you'd need to override the `sessionDidFinishEventsForBackgroundURLSession` closure directly. I think that's what most people will need to do when using background session downloads / uploads. That's why I'm a bit on-the-fence about this change.

#### Closures in the Session Manager

Another approach to this would be to expose the delegate closures directly in the `SessionManager`. This would allow you to keep the `SessionDelegate` internal, yet allow users to override those methods if they need to. This just seems completely wrong though from a design standpoint which is why I didn't go with this approach.

#### Summary

I'd love to get your thoughts around this approach and to see if I interpreted your placeholder closures correctly. IMO, I love this approach because it exposes the session delegate callbacks while still hiding all the internals of all the task delegates which is where all your default logic resides. I think this strikes a really nice balance. Most users will never have a need to override any of the behavior. In the rare case that you need to and really understand what you're doing, you can.


hborders:
I need these features as well. I want to be able to change redirect behavior like https://github.com/Alamofire/Alamofire/issues/314, and I also want to change authentication challenge behavior. This seems like it would enable both of my requirements.

cnoon:
Hi @hborders, I just pulled the upstream master commits into my background_sessions branch. Feel free to fork it or work off it for now until @mattt gets a chance to review this PR and #350. We've already moved to the fork for an internal library we've been building b/c we really need the background session support as well.

hborders:
Thanks. I don't need it immediately, but I'll need it in a few months. Just trying to get out ahead of things. Not trying to pressure @mattt, just upvoting the feature.

mattt:
Thanks so much for your hard work on this, @cnoon. I merged everything into master, and am excited to have this all for the 1.2.0 release. :bread:

cnoon:
Awesome, thanks @mattt!

PR #318, opened 2015-01-28
Use defined type for clojure 'destination' in download functions

astrabot:
Simple improvement of code


mattt:
:+1:

Thanks, @astrabot.

PR #321, opened 2015-01-30
Share framework build schemes for carthage support

yasuoza:
related to #293, #228

Carthage requires shared schemes.
https://github.com/Carthage/Carthage#share-your-xcode-schemes

There were shared schemes, but deleted in c148f7286930d2712ff65777031e4df1ca91bb95 :sob:

This PR re-enable carthage support.

What I did 

screen shot 2015-01-30 at 13 00 49


yasuoza:
Close in favor of #322

PR #326, opened 2015-02-02
Convert Alamofire to a universal framework

kylef:
This pull request removes the two frameworks `iOS Alamofire` and `OS X Alamofire` and replaces them with a single `Alamofire` framework that can work for both platforms. It makes use of configuration files which allow per-platform build settings.

```
FRAMEWORK_SEARCH_PATHS[sdk=iphoneos*]        = $(inherited) '$(SDKROOT)/Developer/Library/Frameworks'
FRAMEWORK_SEARCH_PATHS[sdk=iphonesimulator*] = $(inherited) '$(SDKROOT)/Developer/Library/Frameworks'
FRAMEWORK_SEARCH_PATHS[sdk=macosx*]          = $(inherited) '$(DEVELOPER_FRAMEWORKS_DIR)'
```

I think this is far more elegant than having two frameworks called `iOS Something` and `OS X Something` and it makes integration simpler as you only have a single framework to worry about, not having a bunch of conditional (if iOS link `iOS Alamofire`, if OS X link `OS X Alamofire`).


mattt:
I like this pattern a lot, and wasn't even fully aware that it was possible. Thanks so much for this, @kylef.

norio-nomura:
This single scheme and multi platform configuration is interesting.
But this make Alamofire inconvenient with multi platform projects using Carthage for managing frameworks.
I got answer from Carthage project that will not support this configuration.
Carthage/Carthage#307

mattt:
Based on what @norio-nomura reports and the feedback from the Carthage folks, it looks like a universal framework is not the right solution for now. Thanks for your help though, @kylef. Reverted with 36499d8f3d5e5ecf9c612a301c819c10baa402c2.

PR #333, opened 2015-02-03
Added UIImageView and UIButton extension to support async images download

Reflejo:
Hey @mattt, here is my first version of UIKit categories (UIImageView and UIButton so far) for Alamofire. I'm not sure if you were intending to include it eventually or you don't want this categories at all in Alamofire. 

If the case is the latter ignore the PR.


mattt:
Thanks for your PR, @Reflejo, but there are no current plans to include UIKit extensions in the base project. AFNetworking's UIKit extensions work fine with Swift.

PR #334, opened 2015-02-05
Convenience upload functions in Manager

obohrer:
After switching from using Alamofire.xxx to using a Manager instance i realized that some functions signatures were missing for upload.

Those new convenience functions allow us to use Manager.upload functions directly without having our own
```swift
private func URLRequest(method: Method, URL: URLStringConvertible) -> NSURLRequest {
    let mutableURLRequest = NSMutableURLRequest(URL: NSURL(string: URL.URLString)!)
    mutableURLRequest.HTTPMethod = method.rawValue

    return mutableURLRequest
}
```


mattt:
Thanks, @obohrer. I'd been meaning to do an audit of the top-level and manager APIs. I appreciate you taking the time to make this patch.

PR #335, opened 2015-02-05
Set IPHONEOS_DEPLOYMENT_TARGET to 8.0

lowell:
IPHONEOS_DEPLOYMENT_TARGET was set to 7.0 at the project level. Framework bundles require iOS 8 or later.

This change will not affect iOS 7 usage as that requires the direct inclusion of the source file.


mattt:
Framework support is orthogonal to the base SDK API requirements, so I'm not sure what this really accomplishes. That iOS 7 targets cannot use dynamic frameworks should not have any impact on the declared SDK dependencies.

PR #338, opened 2015-02-09
Fixed compilation errors when using Swift 1.2

boek:
Went through and fixed all the compilation errors caused by the upgrade to Swift 1.2

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -79,7 +79,7 @@ public enum ParameterEncoding {
             return (URLRequest.URLRequest, nil)
         }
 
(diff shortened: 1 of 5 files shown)


moreindirection:
This is really impressive turnaround time!

davejlong:
I'm not sure if this is related to the updates in your PR or not, but, with Carthage, Alamofire won't build for iOS. Only the Mac framework is built in the builds directory.

mattt:
Thanks for the PR, @boek.

I ended up going through and porting the code to 1.2 making a few more idiomatic changes with d337018.

For anyone targeting Swift 1.2, you can now use the [xcode-6.3 branch](https://github.com/Alamofire/Alamofire/tree/xcode-6.3).

PR #346, opened 2015-02-14
Explicitly set iOS deployment target to 7.0

indragiek:
After the switch to a universal framework to in #326, Xcode automatically chooses the latest version of iOS as the deployment target when compiling the framework for iOS (i.e. compiling with Xcode 6.3b1 results in iOS 8.3 being chosen as the deployment target).

This change explicitly sets the deployment target to iOS 7.0 to avoid this issue.


mattt:
Universal frameworks were an interesting idea, but due to incompatibility with other build tools, they were reverted with 36499d8f3d5e5ecf9c612a301c819c10baa402c2.

indragiek:
@mattt Could you pick that commit onto the `xcode-6.3` branch as well? Thanks.

mattt:
@indragiek Yessir. 36499d8f3d5e5ecf9c612a301c819c10baa402c2 is on xcode-6.3 as well now.

indragiek:
@mattt :+1: Thank you

PR #350, opened 2015-02-17
Public SessionDelegate + initial redirect support

### LINKED ISSUE #314: Follow HTTP redirects (or not) per request

jhersh:
Resolves #314.

My goal here is twofold: 

- Allow for per-request HTTP redirect control, by means of
- Making `public` the internal `SessionDelegate` class and its various closure variables for fine-grained request control across numerous `NSURLSessionTaskDelegate` methods.

I also changed the block signature of `taskWillPerformHTTPRedirection` to make its return value an optional `NSURLRequest?` (it was previously `NSURLRequest`) because of this note regarding the `completionBlock` from the [`NSURLSessionTaskDelegate` documentation](https://developer.apple.com/library/ios/documentation/Foundation/Reference/NSURLSessionTaskDelegate_protocol/index.html#//apple_ref/occ/intfm/NSURLSessionTaskDelegate/URLSession:task:willPerformHTTPRedirection:newRequest:completionHandler:):

> A block that your handler should call with either the value of the `request` parameter, a modified URL request object, or `NULL` to refuse the redirect and return the body of the redirect response.

However, my attempts to return `nil` here cause a deadlock/timeout in the test. Instead, I've performed the redirect test by returning a new `NSURLRequest` for the original URL, which also effectively refuses the HTTP redirect. I'd welcome input or suggestions here!

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -280,7 +280,7 @@ public class Manager {
                 "User-Agent": userAgent]
     }
 
-    private let delegate: SessionDelegate
+    public let delegate: SessionDelegate
 
     private let queue = dispatch_queue_create(nil, DISPATCH_QUEUE_SERIAL)
 
@@ -340,7 +340,7 @@ public class Manager {
         return request
     }
 
-    class SessionDelegate: NSObject, NSURLSessionDelegate, NSURLSessionTaskDelegate, NSURLSessionDataDelegate, NSURLSessionDownloadDelegate {
+    public class SessionDelegate: NSObject, NSURLSessionDelegate, NSURLSessionTaskDelegate, NSURLSessionDataDelegate, NSURLSessionDownloadDelegate {
         private var subdelegates: [Int: Request.TaskDelegate]
         private let subdelegateQueue = dispatch_queue_create(nil, DISPATCH_QUEUE_CONCURRENT)
         private subscript(task: NSURLSessionTask) -> Request.TaskDelegate? {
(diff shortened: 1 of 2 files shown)


jhersh:
Re: Travis failure, see #351

cnoon:
You're facing the same issues with the closures and `SessionDelegate` being internal as I was for background session support in #317.

cnoon:
Hi @jhersh, thanks for your PR. A majority of your changes here were merged in #317 which are now part of the `1.2.0` release. If you would like to submit a new PR with the redirect tests, I'm sure they would certainly be welcome.

jhersh:
Closing in favor of #424 and #317. Thanks @cnoon!

PR #353, opened 2015-02-20
Fix Swift 1.2 ambiguity error in responseString.

kevboh:
Calling responseString without an encoding yielded an error: Ambiguous use of 'responseString'.

This change shifts the default into the reponseString(completionHandler) function and makes the encoding argument explicit to match default argument behavior.


kevboh:
Looks like Travis is using Xcode 6.1.

nickvelloff:
I don’t know if Travis supports Xcode 6.3 beta yet but this is how I set it in the .yml file:

language: objective-c
osx_image: xcode611

so try

language: objective-c
osx_image: xcode63

cnoon:
Thanks for your PR @kevboh.

This issue has been resolved in the latest `1.2.0` release.

PR #356, opened 2015-02-21
TLS Key/Certificate pinning

briv:
Thanks for another great library @mattt !
The code is far from finished, but I was wondering if you agreed with the general direction ?
It adds a **ServerTrustPolicy** enum, inspired by the **ParameterEncodig** one, as well as a chainable *serverTrustPolicy()* function for requests.

I was not sure how to integrate globally into the Manager - if that's needed. I saw that for client credentials, the underlying NSURLSessionConfiguration can [provide them nicely](https://github.com/Alamofire/Alamofire/blob/master/Source/Alamofire.swift#L750). 
Would a new serverTrustPolicy property on the Manager, which, when set, assigns a policy to every created request be in the Alamofire spirit ?


briv:
Just to note that after some further testing, the actual implementation is neither faithful to its documentation nor very secure, so this is very much a question on the high level API.

For future work, I guess I'm wondering what potential users expect when pinning either certificates or public keys. For example, should the default case expect the server to present the entire chain rather than a single leaf certificate potentially ? I guess I'm saying that after using AFNetwoking in just one TLS situation, I'd really appreciate guidance on what would make an easily usable and simple-to-get-right API for the majority case.

briv:
This comment started as an attempt to recap some of the dicussion around this API, from blog posts, or AFNetworking issues such as:
* AFNetworking/AFNetworking#1906: request for pinning just root and/or intermediate cert.
* AFNetworking/AFNetworking#1852: public key pinnning discussion, [reedloden's comment](https://github.com/AFNetworking/AFNetworking/issues/1852#issuecomment-36199389) highlights best-pratice and usability while [pjjw](https://github.com/AFNetworking/AFNetworking/issues/1852#issuecomment-70788346) talks about his 2 root cert use-case

The sections below try to outline how the proposed `ServerTrustPolicy` enum would behave, but also why.

#### Certificates
The `.PinCertificates` case could work as follows. It would take a set of certificates to pin, whether roots or leaves, and would evaluate trust by calling out to the `SecTrustSetAnchorCertificates()` and `SecEvaluateTrust()` functions - a thin, straightforward layer. A host would therefore be trusted if **any** of the pinned certificates were within the chain it presents.
This hopefully makes the following use-cases straightforward:
* Pinning a parent (root/intermediate) certificate, with the host presenting a leaf certificate issued by this parent
* Pinning a self-signed non-CA certificate with the host presenting this same certificate.

Potential pitfalls:
* Pinning an entire chain - or any number of certificates - does not require all those certificates to be in the host chain but rather increases the pool of totally trusted certificates - regardless of their root/leaf status.
How prevalent is this use-case though ? I would argue that for certificate pinning, the catered use-case should be to pin root/intermediate certificate(s). Otherwise, when pinning a leaf, might as well use public-key pinning (described below) to ensure both the same security and independence from the issuer, rather than just removing the potential benefit of certificate rotation.
* Different from AFNetworking


#### Public keys
For the `.PinPublicKeys` case, I am less sure how to proceed, in part because I do not know what the dominant use-case is. Should it support multiple keys ? Should it then verify that the pinned keys are a subset of the host chain's keys ? Or should it be similar to `.PinCertificates` and create a pool of trusted keys ? I personally like the simplicity of HPKP, detailed below.
Another peril [Adam Langley](https://www.imperialviolet.org/2011/05/04/pinning.html) mentions when hashing the public key and not the certificate's SubjectPublicKeyInfo:
> This is important because just hashing the public key leaves one open to misinterpretation attacks. Consider a Diffie-Hellman public key: if one only hashes the public key, not the full SPKI, then an attacker can use the same public key but make the client interpret it in a different group. Likewise one could force an RSA key to be interpreted as a DSA key etc.

On the one hand, it's possible to generate multiple types of keys within the opaque `SecKeyRef`, which gives hope that it holds on to this type info however there's really only one kind - [RSA](https://developer.apple.com/library/ios/documentation/Security/Reference/keychainservices/index.html#//apple_ref/doc/constant_group/Key_Type_Values) - defined in iOS. I suppose this unicity side steps the problem - it wouldn't be possible to re-interpret the key any other way, but I would love feedback on this.

###### HPKP
Another idea might be to follow the web's [HPKP](https://developer.mozilla.org/en-US/docs/Web/Security/Public_Key_Pinning), also summarized [here](https://projects.dm.id.lv/Public-Key-Pins_calculator#Public_Key_Pinning_Extension_for_HTTP_.28HPKP.29). This includes affordances for backup keys which might not be in the scope of a default Alamofire policy. However the basic principle is to pin a key and then check if it exists at any level in the host's chain.

briv:
Certificate and public key pinning are implemented by adding a `ServerTrustPolicy` enum.
The API currently has 2 granularities. They are incompatible as the global level always has precedence if set. Perhaps it might be better to remove the request level method altogether ?
- at the `Manager` level, all requests will be affected when using `Alamofire.serverTrustPolicy =`
- at the request level, enabling code like `request(.GET, "https://example.com").trustPolicy(.InsecureTrustAny)`

The pre-existing closures in the session/task delegates are still present and can override this enum if needed.
A server is trusted if:
- its certificate chain is valid and *any* certicate in this chain is within the set of pinned certificates. The host name is also always checked (though this might need to be changed)
- *any* of its keys is contained within the set of pinned keys.

The `ServerTrustPolicy` enum provides easy-access vars to any certificates (or public keys of certificates) contained in the app bundle.

A simple example usage would be along the lines of:
```
Alamofire.serverTrustPolicy = .PinCertificates(ServerTrustPolicy.bundledCertificates)
Alamofire.request(.GET, "https://example.com", parameters: ["foo": "bar"])
```

or per request:
```
Alamofire.request(.GET, "https://example.com", parameters: ["foo": "bar"])
	.trustPolicy(.InsecureTrustAny)
```

mthole:
@briv It looks like you're on the right track here! I haven't reviewed the code but your summaries above sound like the correct/desired behavior to me (and I've done this cert pinning dance a few times). Thanks

cnoon:
Hi @briv, thank you for all your work you've put in here! We have SSL/TLS Key/Certificate pinning slated for our 1.3.0 release. We're starting to discuss this now and will keep you posted. Thanks again for all your hard work here! 👍🏻

briv:
Thank you for the kind words. I'll send them right back for how sweet Alamofire is.

Since the global policy was over-riding request-level policies in a silent way, that is now out. It is definitely do-able but I didn't want to reorganize too much without consulting. Also to simplify using any code from these patches.
I'm not sure how much this TLS security policy should take away from those who would want total configurability - I suppose this is every API's trade-off =). My instinct says "very little" as Alamofire wants to cater to all use cases ? At the same time, I'm thinking that a TLS pinning API that makes it easy to shoot yourself in the foot is not the best.

To go on about this point, the `SessionDelegate` exposes 2 ways to potentially override challenge responses and the underlying `TaskDelegate`, which is however private to the external world, exposes another one. I wasn't confident to reorganize those out without any discussion, so this last set of changes just tries to do the minimum while making it (hopefully) easy to reuse those pieces in a different way if need be.

In summary, the latest changes are rebased and squashed to give:
- A `TLSTrustPolicy` enum which evaluates a server's TLS credentials. I think the `.Custom` case might not make as much sense anymore since the user might want to implement the delegate closure `sessionDidReceiveChallenge` at that point? Unless this enum takes over completely.
- A `TLSPolicy` var on the `SessionDelegate` which only acts as a convenience to set the `sessionDidReceiveChallenge` closure.
- A global `TLSPolicy` var to act on `Manager.sharedInstance`
- Some tests

An example usage:
```swift
Alamofire.TLSPolicy = .PinCertificates(TLSTrustPolicy.bundledCertificates)
Alamofire.request(.GET, "https://example.com", parameters: ["foo": "bar"])
```

I'm more than happy to discuss, implement, re-juggle, or rework whatever you think is best, or just take a step back and wait for 1.3.0.

cnoon:
At this point, I think it would be best to hold off. The technical committee (TC) has already decided that we need to build have one of us build out this feature. It is one of the most critical pieces of the library to get right and be able to provide top notch support for at all times moving forwards.

With that said, we greatly value your efforts here and this will certainly give us a head start when we get started!

cnoon:
Hi @briv, I just pushed up a PR that implements TLS public key and certificate pinning. Thank you for all the hard work you put into this PR. It really helped get me going in the right direction. If you could please take a look at #581 I'd really appreciate it. I want to get everyone involved to make this feature rock solid.

I'm going to close out this PR. Please direct all further comments to #581.

Issue #358, opened 2015-02-24
is iso-8859-1 encoding supported by Alamofire?


jigneshsheth:
I am getting unexpectedly error sometimes in my response.

fatal error: unexpectedly found nil while unwrapping an Optional value
Printing description of serializationError:
Error Domain=NSCocoaErrorDomain Code=3840 "The operation couldn’t be completed. (Cocoa error 3840.)" (Unable to convert data to string around character 23447.) UserInfo=0x171668ac0 {NSDebugDescription=Unable to convert data to string around character 23447.}

Any suggestions on it. How do I resolve this issue other then change the response encoding to UTF-8

kylef:
ISO-8859-1 is not a valid encoding for JSON. As per the [JSON specification](http://www.ietf.org/rfc/rfc4627.txt) (3. encoding).

```
3.  Encoding

   JSON text SHALL be encoded in Unicode.  The default encoding is
   UTF-8.

   Since the first two characters of a JSON text will always be ASCII
   characters [RFC0020], it is possible to determine whether an octet
   stream is UTF-8, UTF-16 (BE or LE), or UTF-32 (BE or LE) by looking
   at the pattern of nulls in the first four octets.

           00 00 00 xx  UTF-32BE
           00 xx 00 xx  UTF-16BE
           xx 00 00 00  UTF-32LE
           xx 00 xx 00  UTF-16LE
           xx xx xx xx  UTF-8
```

Please correct me if this error isn't from the JSON deserialisation, but I can't find any other code-paths that result in a serialization error and you didn't explicitly state what you was doing. So I'm assuming JSON.

Apple's JSON API also only allows Unicode:

```
The data must be in one of the 5 supported encodings listed in the JSON specification: UTF-8, UTF-16LE, UTF-16BE, UTF-32LE, UTF-32BE.
```

kylef:
Unrelated for JSON, but I've made #359 which will cause Alamofire to follow the server's encoding for `responseString`.

jigneshsheth:
Cool, Thanks for your response Kylef. I was doing JSON deserialization and I would make my back-end changed to be UTF-8 formatted response.

mattt:
Fixed by 6cffddc. See #359.

staticdreams:
[...]

PR #359, opened 2015-02-24
[String Response] Respect remote server's content encoding

kylef:
We should always follow the content encoding sent from the server instead of assuming UTF-8 or whatever the user has configured.

Note, `iso-8859-1` is the default encoding for HTTP 1.1. As such that is the default if a server didn't explicitly set it to utf-8.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -1376,12 +1376,13 @@ extension Request {
     /**
         Creates a response serializer that returns a string initialized from the response data with the specified string encoding.
 
-        :param: encoding The string encoding. `NSUTF8StringEncoding` by default.
-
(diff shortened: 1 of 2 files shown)


mattt:
Thanks for your PR, @kylef. I really appreciate you following up on #358 to find the underlying problem here.

We should definitely respect the server's character encoding, when provided. We should definitely follow the spec to use ISO-8859-1 by default. However, we should also provide a way to override this behavior for when the server is doing something wrong, and it's not possible to fix.

I implemented this with 6cffddcf150539daea674fa9367b63a1d46b0519, which should resolve the original issues. Let me know if I missed anything here.

Thanks again, @kylef.

PR #361, opened 2015-02-25
Resolves explicit self requirement in Swift 1.2

winzig:
Resolves issue I encountered trying to build in Swift 1.2 beta 2: 

    Call to method 'queryComponents' in closure requires explicit 'self.' to make capture semantics explicit.

Not sure if this is a new requirement in beta 2 of Swift 1.2 or not, this is my first time using Alamofire.

PR #362, opened 2015-02-25
Fixed Xcode 6.3 Beta2 complaining about explicit call to 'self'

csotiriou:


csotiriou:
Sorry, I didn't see that this is a duplicate of #361

Issue #206, opened 2014-11-05
escape function crashes when escaping long Chinese strings


PrideChung:
This problem would be easier to reproduce on iOS devices with smaller RAM, I first found it on an iPod Touch 5 with 512MB RAM. It crashes everytime when I try to encode a POST parameter with more than 500 Chinese characters, and it only gives me an EXC_BAD_ACCESS with no further information. On iPhone 5 with 1GB RAM it can handle about 1300 Chinese characters but still crash if the limitation is exceeded. 

For quick reference, this is the `escape` function I copied from `Alamofire.swift` line 154

``` swift
func escape(string: String) -> String {
    let allowedCharacters =  NSCharacterSet(charactersInString:" =\"#%/<>?@\\^`{}[]|&+").invertedSet
    return string.stringByAddingPercentEncodingWithAllowedCharacters(allowedCharacters) ?? string
}
```

I can almost sure the reason of the crash is `stringByAddingPercentEncodingWithAllowedCharacters` takes too much memory, this is particularly obvious when it's encoding Chinese or Japanese strings, maybe because the encoding process is more complex for these languages.

I discovered that if I slice the long string into small pieces and call `stringByAddingPercentEncodingWithAllowedCharacters` in batches like below will fix the memory peak problem, 

``` swift
func escape(string: String) -> String {
    let allowedCharacters =  NSCharacterSet(charactersInString:" =\"#%/<>?@\\^`{}[]|&+").invertedSet
    let batchSize = 100
    var escapedString = ""
    let stringLength = countElements(string)
    for var i = 0; i < stringLength; i += batchSize {
        let rangeLength = i + batchSize > stringLength ? stringLength - i : batchSize;
        let slicedString = (string as NSString).substringWithRange(NSMakeRange(i, rangeLength))
        escapedString += slicedString.stringByAddingPercentEncodingWithAllowedCharacters(allowedCharacters) ?? slicedString
    }
    return escapedString
}
```

I don't even need to place `autoreleasepool` between each for loops, I suspect there is already a autoreleasepool inside the implementation of `stringByAddingPercentEncodingWithAllowedCharacters`

[...]

PR #369, opened 2015-03-04
fix escaping query parameters

slizeray:
For escaping the query parameters I propose to use the iOS 7 API and above
stringByAddingPercentEncodingWithAllowedCharacters(NSCharacterSet.URLQueryAllowedCharacterSet())

The result is not the same as the current implementation which for instances escapes the / parameter whereas iOS does NOT

### DIFF
--- a/Example/DetailViewController.swift
+++ b/Example/DetailViewController.swift
@@ -68,7 +68,7 @@ class DetailViewController: UITableViewController {
         self.refreshControl?.beginRefreshing()
 
         let start = CACurrentMediaTime()
-        self.request?.responseString { (request, response, body, error) in
+        self.request?.responseString( completionHandler: { (request, response, body, error) -> Void in
             let end = CACurrentMediaTime()
             self.elapsedTime = end - start
 
@@ -80,7 +80,7 @@ class DetailViewController: UITableViewController {
 
             self.tableView.reloadData()
             self.refreshControl?.endRefreshing()
-        }
+        })
     }
 
     // MARK: UITableViewDataSource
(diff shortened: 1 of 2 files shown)


mattt:
`stringByAddingPercentEncodingWithAllowedCharacters:` suffers from excessive memory consumption. See #206.

slizeray:
@mattt OK, then you need to change the characters to escape because the current implementation escapes "/" and it should NOT

mattt:
@slizeray This is incorrect. According to [RFC 3986 § 2.2](http://tools.ietf.org/html/rfc3986#section-2.2), `/` is a general delimiter, and needs to be escaped.

slizeray:
@mattt This is perfectly correct. According to the [RFC 3986 $ 3.4](http://tools.ietf.org/html/rfc3986#section-3.4)

The characters slash ("/") and question mark ("?") may represent data  within the query component.
query       = *( pchar / "/" / "?" )

Indeed, Apple has perfectly implemented the RFC.

mattt:
> Beware that some older, erroneous implementations may not handle such data correctly when it is used as the base URI for relative references...

> ...it is sometimes better for usability to avoid percent-encoding [slash and question mark]

@slizeray That doesn't exactly inspire confidence. If it comes down to a question of usability concerns versus incorrect server-side handling, I'll err on the side of safety.

slizeray:
Well you put back some older code because of memory consumption. In between you had exactly the behaviour that now you think is wrong?
My problem is that I am using a server that does not understand the URL if I escape the legal "/" character.
Would it be possible to have an option such as we can choose between escaping or not "/"?

mattt:
@slizeray I wasn't aware of this specific difference in behavior previously, but knowing this now, I'm more comfortable keeping consistent default encoding behavior between Alamofire and AFNetworking. I understand how this might be inconvenient, which is why Alamofire offers two different ways to easily override this. First, you could use the `Custom` parameter encoding, supplying your own function to do the parameter escaping. Alternatively, you could bypass all of the URL encoding completely by passing a value of type conforming to `URLRequestConvertible` (`NSURLRequest` or a custom type) directly into `request`, `upload` and `download`.

slizeray:
@mattt OK Thanks I will give it try! As long as there is a solution, I am fine with that! That said, we could have an option both for Alamofire and AFNetworking....

PR #370, opened 2015-03-04
fix escaping query parameters / and ? are valid characters in the query parameters

slizeray:
See [RFC 3986  $ 3.4] (http://tools.ietf.org/html/rfc3986#section-3.4)


mattt:
I cherry-picked the specific change for escaped characters with 9075691ca5e2e53c7fcb6b286a19abee874e97d3. Thanks again, @slizeray!

PR #376, opened 2015-03-09
Update readme - fix cocoapods file example.

skywinder:
With your Podfile example `pod` (0.36.0.rc.1) gets error:
>[!] Pods written in Swift can only be integrated as frameworks; this feature is still in beta. Add `use_frameworks!` to your Podfile or target to opt into using it.

So, you should set additional flag in the beginning if the Podfile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,12 @@ $ gem install cocoapods --pre
 To integrate Alamofire into your Xcode project using CocoaPods, specify it in your `Podfile`:
 
 ```ruby
-use_frameworks!
 source 'https://github.com/CocoaPods/Specs.git'
-platform :ios, '8.0'
 
-pod 'Alamofire', '~> 1.1'
+platform :ios, '8.0'
+use_frameworks!
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,6 +51,7 @@ $ gem install cocoapods --pre
 To integrate Alamofire into your Xcode project using CocoaPods, specify it in your `Podfile`:
 
 ```ruby
+use_frameworks!
 source 'https://github.com/CocoaPods/Specs.git'
 platform :ios, '8.0'
```
cnoon: It would be nice if we followed the same structure as the CocoaPods guides. Could you update this section to the following?

```ruby
source 'https://github.com/CocoaPods/Specs.git'

platform :ios, '8.0'
use_frameworks!

pod 'Alamofire', '~> 1.1.4'
```

skywinder:
Done.

mattt:
Thanks, @skywinder. However, an equivalent change was merged with #380.

PR #379, opened 2015-03-11
Bug fix for deinit race condition on task delegate queue

cnoon:
This is essentially the same issue that I opened #269 for where `Manager` deinitialization was crashing.

#### Problem

Unfortunately, the test I wrote in that case did not reflect all the potential issues. By removing the deinit that contained the `session.invalidateAndCancel()`, that only ended up masking the root cause of the crash. I managed to unwind it today and get to the root issue.

The problem is that the `TaskDelegate` dispatch queue cannot be properly released when in a suspended state. The libdispatch library will fail under certain race conditions. If you do not actually call `resume` on the task, then the queue will always remain in the suspended state until the time when it is deinitialized. The root problem I believe stems from a mismatch between the `dispatch_resume` and `dispatch_suspend` calls. They need to [balance each other out](https://developer.apple.com/library/prerelease/ios/documentation/Performance/Reference/GCD_libdispatch_Ref/index.html#//apple_ref/c/func/dispatch_resume) and we cannot always guarantee that during deinitialization.

### Option 1 - dispatch_resume in `TaskDelegate.deinit`

There are several ways to approach the problem. The first would be to resume the queue during deinitialization of the `TaskDelegate` to attempt to balance out the suspend and resume calls. The problem with this approach is that you will then shut down the queue while it's in the middle of executing the closures inside it. Also, the potential for corrupting state is very high since Alamofire is designed to only `dispatch_resume` when the request has actually completed. I worked through this approach and it is terribly problematic and doesn't fix the race condition.

### Option 2 - Switch to `NSOperationQueue` for `TaskDelegate`

The solution I landed on was to swap out the `dispatch_queue_t` with an `NSOperationQueue`. This gives us the ability to do the following:

* We can still suspend and resume the queue
* We can still define a serial execution behavior
* We can cancel all the operations (biggest limitation of dispatch_queue_t)
* We can resume after canceling all operations to unsure the `underlyingQueue` is not `suspended`

After swapping out the old queue with the operation queue, everything is working exactly as it should. Manager deinitialization works flawlessly.

#### Added `invalidateAndCancel` back into Manager.deinit

IMO, this is the behavior that should be used rather than `finishTasksAndInvalidate`. If the Manager is attempting to deinitialize, then Alamofire should do everything in its power to make sure that deinitialization happens immediately. That cannot happen if there are current running tasks that will keep the session in memory along with the Manager. This seems very problematic when you consider that you're actually in the middle of terminating the app. You want to avoid having any queues working on closures when the termination actually happens to avoid corrupting state.

#### Summary

The change from a `TaskDelegate` `dispatch_queue_t` to an `NSOperationQueue` absolutely has to happen to fix the race condition crashing. Whether to `invalidateAndCancel` or `finishTasksAndInvalidate` or nothing in the `Manager.deinit` I think needs to input from the community.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -211,8 +211,6 @@ extension NSURLRequest: URLRequestConvertible {
 
 /**
     Responsible for creating and managing `Request` objects, as well as their underlying `NSURLSession`.
-
-    When finished with a manager, be sure to call either `session.finishTasksAndInvalidate()` or `session.invalidateAndCancel()` before deinitialization.
 */
 public class Manager {
 
@@ -296,6 +294,10 @@ public class Manager {
     required public init(configuration: NSURLSessionConfiguration? = nil) {
         self.delegate = SessionDelegate()
         self.session = NSURLSession(configuration: configuration, delegate: delegate, delegateQueue: nil)
+    }
+
+    deinit {
+        self.session.invalidateAndCancel()
     }
 
     // MARK: -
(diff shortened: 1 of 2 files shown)


cnoon:
Approved offline by TC.

PR #387, opened 2015-03-15
Water the swift1.2 branch, get example compiling again

rromanchuk:


cnoon:
Thanks for your PR @rromanchuk.

This is no longer necessary as everything has been pulled in as part of the `1.2.0` release.

PR #390, opened 2015-03-21
Update OS X target build settings

nolanw:
`APPLICATION_EXTENSION_API_ONLY` got left out on the OS X target. And since the README specifies OS X 10.9+, I figured the target should agree.

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -395,6 +395,7 @@
 		4DD67C1F1A5C55C900ED2280 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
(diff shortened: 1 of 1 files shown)


cnoon:
Hi @nolanw,

Thank you for your Pull Request...it is much appreciated. I didn't end up merging your PR directly because the `APPLICATION_EXTENSION_API_ONLY` setting is already set properly.  

> The `APPLICATION_EXTENSION_API_ONLY` is set to `YES` in both framework targets. It is set to `NO` in the test and example targets.

I have updated all the deployment targets in ccc40a1a.

> Somehow these were reset when we merged the Swift 1.2 branch into master. Either way, ccc40a1a put us back on track.

nolanw:
Well hey, so long as the right setting gets set I'm happy!

PR #394, opened 2015-03-23
Use system provided SSL validation until proper customization is implemented

mthole:
For `NSURLAuthenticationMethodServerTrust` authentication method, stop indiscriminately trusting all SecTrusts. Instead, rely on default behavior -- which will do basic evaluation of the certificate chain against the System's CA Root Certificates.

Alamofire will presumably want to implement custom SSL certificate pinning and the ability to turn invalid certificates, but IMO this is *much* better default behavior.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -741,17 +741,17 @@ public class Request {
                 if challenge.previousFailureCount > 0 {
                     disposition = .CancelAuthenticationChallenge
                 } else {
-                    // TODO: Incorporate Trust Evaluation & TLS Chain Validation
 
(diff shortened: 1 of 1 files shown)


mattt:
Thanks, @mthole.

kozmic:
Shouldn't this weakness be noted in the Changelog? Users have no way to know they are using a version that is vulnerable to MiTM attacks.

jshier:
SSL verification and customization has been part of Alamofire since the 2.0 release. If you have any concerns with that functionality, please use the Alamofire security email, security@alamofire.org, to contact us with your specific concerns.

PR #400, opened 2015-03-25
Updated the code signing identities for the frameworks and targets

cnoon:
In order to properly support Carthage and CocoaPods at the same time with cross-platform targets, the code signing identities need to be modified a bit to make sure everyone can build Alamofire using Carthage. Thankfully, CocoaPods handles all this independently.

To properly support Carthage, iOS targets and tests need to code sign as `iPhone Developer`. This is due to the fact that Xcode *requires* them to be code signed. Mac OS X targets and tests are different altogether. They don't require code signing at all. Therefore, it should be disabled to allow users without a Mac Developer account to build Alamofire on OS X and run the tests as well as those without an `iPhone Developer` account who are simply trying to build Alamofire through Carthage. All frameworks will get resigned anyways when embedded within a final app target.

This will make creating cross-platform libraries ontop of Alamofire much easier for all those involved. You can find more information [here](https://github.com/Carthage/Carthage/issues/399) and [here](https://github.com/Carthage/Carthage/issues/281).

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -395,6 +395,7 @@
 		4DD67C1F1A5C55C900ED2280 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
 				COMBINE_HIDPI_IMAGES = YES;
 				DEFINES_MODULE = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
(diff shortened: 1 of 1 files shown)


mattt:
Thanks for looking into this, @cnoon. I'm just about to cut Alamofire 1.1.5. When that's released, can you please verify that things are working as expected?

cnoon:
Will do @mattt.

cnoon:
Looks good @mattt. All the tests are passing on both targets on `1.1.5` and the Carthage builds are all good as well.

```
cnoon:~/Desktop/AlamofireCarthage$ carthage update
*** Fetching Alamofire
*** Checking out Alamofire at "1.1.5"
*** xcodebuild output can be found in /var/folders/mc/9x8d6r415qxgyxry1y8vm9h0jj7n1m/T/carthage-xcodebuild.9jtd3m.log
*** Building scheme "Alamofire iOS" in Alamofire.xcworkspace
*** Building scheme "Alamofire OSX" in Alamofire.xcworkspace
```

Thanks as always!

PR #401, opened 2015-03-25
Switched iOS deployment target to iOS 8.0 for iOS target and tests

cnoon:
Setting the iOS deployment target to iOS 7.0 causes warnings all over this project and all other projects that include it. While I understand that Alamofire can be used in iOS 7.0 projects by dragging the `Alamofire.swift` file into the project, the iOS deployment target inside the Xcode project is ignored in those cases.

embedded framework warning

In all cases where the Xcode project is actually being used (Alamofire Xcode project, git submodules for other apps and libraries and Carthage) the warning is actually being thrown. The Xcode project deployment target should be updated to iOS 8.0 to allow all these Xcode project use cases to no longer see the warning which will not affect the actual iOS 7.0 use case.

Hopefully that all makes sense.


mattt:
Makes sense to me. I'll tag a new patch release shortly.

PR #402, opened 2015-03-25
added certificate pinning

isoiphone:
Certificate pinning is a *must have* for my application, so I shoe-horned it in here for my own purposes.

I'm not sure of the best place for it to go in Alamofire, or the best way for it to be configured (setting a shared array of certs is less than ideal)

This PR has a code-smell about it, but it works for me and should be a good start.
To add a certificate I do the following:

    if let path = NSBundle.mainBundle().pathForResource("mywebsite.com", ofType: "cer") {
        if let certData = NSData(contentsOfFile: path) {
            if let cert = SecCertificateCreateWithData(nil, certData) {
                Manager.defaultPinnedCertificates.append(cert.takeRetainedValue())
            }
        }
    }

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -271,6 +271,8 @@ public class Manager {
                 "Accept-Language": acceptLanguage,
                 "User-Agent": userAgent]
     }()
+    
+    public static var defaultPinnedCertificates: [SecCertificate] = []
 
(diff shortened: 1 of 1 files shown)


isoiphone:
Simplicity and "gets the job done" were goals.
A more full featured trust/security implementation is less desirable.

PR #403, opened 2015-03-26
Privatize NSURLSession & NSURLSessionTask objects.

inamiy:
Accessing raw `NSURLSession` & `NSURLSessionTask` objects outside of the framework should be prohibited (e.g. calling `session.invalidateAndCancel()`), so I marked those properties with `private` modifier. These objects's functionality should always be accessed via `Manager` & `Request`'s methods.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -286,6 +286,8 @@ public class Manager {
 
     /// The underlying session.
     private let session: NSURLSession
+    
+    public let configuration: NSURLSessionConfiguration?
 
     /// Whether to start requests immediately after being constructed. `true` by default.
     public var startRequestsImmediately: Bool = true
@@ -295,6 +297,7 @@ public class Manager {
     */
     required public init(configuration: NSURLSessionConfiguration? = nil) {
         self.delegate = SessionDelegate()
+        self.configuration = configuration
         self.session = NSURLSession(configuration: configuration, delegate: delegate, delegateQueue: nil)
     }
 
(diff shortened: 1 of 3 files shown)


cnoon:
While I'm generally all for keeping things as locked down as possible, I'm not a fan of this proposed change. To me, you need to make a much better case as to why these should be made private at all.

There are many use cases where having these publicly available is absolutely necessary. Here are just a few I can think of off the top of my head.

* Using the `NSURLSessionTask` `taskIdentifier` to identify `Request` objects
* Tracking the state of an `NSURLSessionTask` (granted you did consider this case)
* Manually querying the `NSURLSession` to see how many active requests are running and what types of tasks they are (necessary for background sessions)
* Needing to `finishTasksAndInvalidate` or `invalidateAndCancel` for whatever reason

There are WAY to many APIs that would need to be exposed from the `Manager` and `Request` classes to cover all these cases. IMO, the public scope is the proper ACL.

mattt:
Thanks for your PR, @inamiy, but I agree with @cnoon.

I was initially of the same opinion, and originally set their access level to `private`. However, as much as I'd prefer a non-leaky abstraction, where as much of the underlying `NSURLSession` APIs are irrelevant (as CFNetwork is to Foundation), there are just too many potential use cases (known and unknown) where _not_ exposing this would cause undue frustration.

inamiy:
@mattt @cnoon
All right, it seems full encapsulation of `NSURLSession` is not a good choice.
Thanks for reply.

PR #404, opened 2015-03-30
Remove optional type from Request.progress property.

pitiphong-p:
The delegate.progress always has value so I don't see any point that Request.progress property to be Optional type. Please consider on removing Optional type from it.


pitiphong-p:
Note: All of tests passed on my machine. Looks like CI test failed due to network issues.

mattt:
Thanks, @pitiphong-p.

pitiphong-p:
:smile: :+1:

PR #406, opened 2015-04-01
Fixed compilation error for Xcode 6.3 beta 4

nduplessis:


cnoon:
Thanks for the PR @nduplessis. While this will fix the compilation issue, I don't think this is the appropriate way to address the root problem. I'll try to push up a PR today that fixes the root issue so that you can still use trailing closures as designed.

cnoon:
Hi @nduplessis, I just pushed up #408 which solves the root cause of this issue.

PR #412, opened 2015-04-05
Fix minor grammatical error

getaaron:
This sentence in the README is missing a subject and active verb:

> _Since there is no published specification for how to encode collection types, the convention of appending `[]` to the key for array values (`foo[]=1&foo[]=2`), and appending the key surrounded by square brackets for nested dictionary values (`foo[bar]=baz`)._

This PR resolves it by inserting "*AlamoFire follows* the convention of appending…"

Alternatively, this sentence might be made more readable by splitting it into two sentences:

> _There is no published specification for how to encode collection types. AlamoFire follows the convention of appending `[]` to the key for array values (`foo[]=1&foo[]=2`), and appending the key surrounded by square brackets for nested dictionary values (`foo[bar]=baz`)._

PR #415, opened 2015-04-05
[travis] Use Xcode 6.2

kylef:

PR #416, opened 2015-04-06
Fix minor grammatical error

getaaron:
Fixes the issue described in https://github.com/Alamofire/Alamofire/pull/412, but with the appropriate casing of Alamofire.

The lowercase f is not to be confused with the [medial s](http://en.wikipedia.org/wiki/Long_s).


getaaron:
Looks like https://github.com/Alamofire/Alamofire/pull/415 might fix the failing test.

mattt:
Thanks for this, @getaaron.

Issue #419, opened 2015-04-08
Placement of expectation.fulfill() potentially skips XCTAsserts


natecook1000:
The call to `fulfill()` on an `XCTestExpectation` instance needs to happen after all the `XCT`-assertions in an asynchronous block have been evaluated. The call to `waitForExpectationsWithTimeout` suspends the run loop until the all the expectations have been fulfilled, then immediately continues with the testing process. On the final test of the testing run, this can skip any assertions that come after the last call to `fulfill()` as the run loop exits after the final test.

You can test this by adding an `XCTFail` in the last test method of the the last test class: `UploadResponseTestCase.testUploadRequestWithProgress`:

    func testUploadRequestWithProgress() {
        let URL = "http://httpbin.org/post"
        let data = "Lorem ipsum dolor sit amet".dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)

        let expectation = expectationWithDescription(URL)

        let upload = Alamofire.upload(.POST, URL, data!)
        upload.progress { (bytesWritten, totalBytesWritten, totalBytesExpectedToWrite) -> Void in
            expectation.fulfill()

            XCTAssert(bytesWritten > 0, "bytesWritten should be > 0")
            XCTAssert(totalBytesWritten > 0, "totalBytesWritten should be > 0")
            XCTAssert(totalBytesExpectedToWrite > 0, "totalBytesExpectedToWrite should be > 0")

            upload.cancel()
            
            XCTFail("Missed test")
        }

        waitForExpectationsWithTimeout(10) { (error) in
            XCTAssertNil(error, "\(error)")
        }
    }

The "Missed test" call will never be executed, and the test will either complete reporting success or end in a SIGABRT crash.

mattt:
Fixed by #420. Thanks for spotting that, @natecook1000.

PR #420, opened 2015-04-08
Move expectation.fulfill() to end of blocks

natecook1000:
This is a proposed fix for issue #419.


cnoon:
Good catch @natecook1000! These changes all look good.

mattt:
Thanks so much, @natecook1000. Nice catch!

PR #424, opened 2015-04-09
[Tests] Added redirect sessiondelegate tests

jhersh:
These are the tests I wrote for #314 and #350, which were resolved with the merging of #317. These tests cover some of the new HTTP redirect behavior.


jhersh:
Tests pass locally, but Travis is not yet on Swift 1.2 :chart_with_downwards_trend:

mattt:
Thanks for this, @jhersh. I really appreciate it!

PR #451, opened 2015-04-23
Add Carthage/Build folder into .gitignore

tomassliz:
When is Alamofire used with Carthage `--use-submodules` flag, folder `Carthage/Build` is automatically tracked as uncommitted change. But this is unnecessary in this case, so I think that is better to ignore it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ DerivedData
 *.hmap
 *.ipa
 *.xcuserstate
+
+# Carthage
+
+Carthage/Build


cnoon:
I can think of any downsides to adding this ignore flag, and it certainly resolves the issue you pointed out with using Carthage with submodules. IMO this is a good change that would certainly benefit Carthage users. 

Thanks @tomassliz!

tomassliz:
For completeness, below are discussions about this topic:
- [Carthage/issues/420#issuecomment-95772555](https://github.com/Carthage/Carthage/issues/420#issuecomment-95772555)
- [jspahrsummers/xcconfigs/pull/37](https://github.com/jspahrsummers/xcconfigs/pull/37)

cnoon:
Thanks @tomassliz!

tomassliz:
You're welcome, I'm glad to help.

PR #455, opened 2015-04-26
Fixed some inaccurate docs on a few of the public delegate closures

cnoon:
I went through all the public delegate closures docs and fixed any that were a bit off. It appears that a couple suffered from a copy/paste error and another one had the wrong method signature. They all should be fixed at this point.


cnoon:
Declining this PR because I'm going to put up a bigger one that also includes this change.

PR #456, opened 2015-04-26
Updated the delegate closures and docs to match their method signatures

cnoon:
I first went through all the public delegate closure docs and fixed any that were a bit off. It appears that a couple suffered from a copy/paste error and another one had the wrong method signature. They all should be fixed at this point.

Then I realized that the method signatures were written for Swift 1.1 and had not yet been updated for Swift 1.2. The closures have now all been updated to match the delegate method that they can override. I also updated any logic in the tests that were affected by this change. All the tests are now passing as expected.


mattt:
Looks good to me. Thanks, @cnoon!

PR #463, opened 2015-05-01
Add option for session delegate customisation

tomassliz:
I want to use custom session delegate, because sometimes I have to skip validation for invalid certificate. It's a shame that `Manager` class is not providing option how to customize delegate methods handling.


mattt:
You should not need to subclass `SessionDelegate`, as all of the delegate callbacks to be customized are available as closure properties. In your case, you should set `sessionDidReceiveChallenge`.

tomassliz:
Sorry, I missed this one ...  Thank you for this advice!

PR #469, opened 2015-05-04
allow nil for prevent redirection

subicura:
To prevent redirection, completionHandler shuold be set nil.


cnoon:
I'm going to merge this in @subicura, but passing in `nil` at the moment is causing deadlocks. Please see #350 and #424 for more info. There is also additional documentation in the tests.

PR #471, opened 2015-05-05
Split Project into Multiple Swift Files

cnoon:
I broke the project up into multiple Swift files separated into `Core` and `Features` groups. I was very careful to split each major change into a separate commit so we didn't lose any of the critical ACL changes that were required.

I also took a stab at reworking some of the MARK statements to attempt to make it as easy as possible to quickly find the section you are looking for. I think it turned out pretty well, but I really would love some additional feedback if something doesn't seem quite right.

I also found some other minor things that I'd like to switch up, but I filed separate cards in Trello for each of them. That way this PR focuses only on the task at hand 👍🏻

> All the tests are passing on iOS. We have a separate Trello card for fixing up the OS X tests.


mattt:
This looks great, @cnoon. Thanks so much for taking the time to refactor things into something a bit more understandable for anyone less familiar with the library.

meng-zhang:
@cnoon Really nice work! Thanks!

fukemy:
hi, can you tell me the way to get your pr? im new in github

PR #474, opened 2015-05-09
Now working with special characters åäö

troligtvis:


mattt:
From the header documentation for `URLStringConvertible`:

>  A URL that conforms to RFC 2396.
>  Methods accepting a `URLStringConvertible` type parameter parse it according to RFCs 1738 and 1808.
>  See http://tools.ietf.org/html/rfc2396
>  See http://tools.ietf.org/html/rfc1738
>  See http://tools.ietf.org/html/rfc1808

The API consumer is expected to percent-escape accordingly. Otherwise, valid pre-escaped URLs  would not work as expected.

PR #478, opened 2015-05-13
Replace return from NSURLRequest type to NSMutableURLRequest in encode function

skywinder:
Replace return from NSURLRequest type to `NSMutableURLRequest` since it's
always `NSMutableURLRequest` except one case (when parameters == nil).
This change get users more flexibility in usage of this `encode` function and eliminate of redundant casting of `NSMutableURLRequest` to `NSMutableURLRequest` after (because in most of cases `encode` function returns `NSMutableURLRequest` already).

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -63,7 +63,7 @@ public enum ParameterEncoding {
     /**
         Uses the associated closure value to construct a new request given an existing request and parameters.
     */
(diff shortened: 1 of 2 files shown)


cnoon:
Apologies for not commenting on this earlier @skywinder. We really appreciate you submitting the PR! I merged a slightly modified version of your changes in 464e31b5 and credited you with the change.

PR #480, opened 2015-05-13
Fix test in case of respone is nil

skywinder:
Fix test in case of response is `nil` (in my case it is) - we are get crash when try to unwrap optional variable and test not finished at all. So I add additional check, that response is not nil.

screen shot 2015-05-13 at 19 48 13


cnoon:
Thank you for your PR @skywinder. 

We're going to be reworking the entire test system in the `1.2.3` release coming very soon. We will be hardening the tests as well as adding safeguards around each test to ensure crashes cannot occur even if the tests fail.

I'm going to decline this PR because we're going to address this issue in a much bigger way. Thanks again. Cheers :beers:

PR #488, opened 2015-05-17
Alamofire Activity Indicator

mustafaibrahim989:


mattt:
Thanks for the PR, @mustafaibrahim989, but this implementation is incompatible with our project design goals. Specifically, it creates a strong dependency on UIKit, making Alamofire incompatible with OS X. It also has tight coupling to a shared singleton instance and isn't thread-safe. We are looking into ways to incorporate this kind of functionality in the future.

PR #489, opened 2015-05-18
Switch test suite to use Given | When | Then structure

cnoon:
This PR contains several major changes to the test suite including:

1. Refactors all tests to follow the `Given` | `When` | `Then` structure
1. All tests now succeed on iOS and OS X
1. Adds safeguards to all tests to prevent the test suite from crashing when tests fail

My apologies for the horrendous diff. I would suggest checking out the branch and reviewing it that way. None of the tests have been removed. They have only been refactored to follow the new structure and to be safe from potential crashes.

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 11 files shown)


cnoon:
cc @kcharwood, @mattt and @kylef for review.

kcharwood:
What do we need to do to get Travis or some other CI setup here?

cnoon:
We need Travis CI to support Xcode 6.3 because we've moved over to Swift 1.2 which only ships in that version of Xcode.

mattt:
I dig it. Nice work, @cnoon.

kylef:
:+1:

PR #483, opened 2015-05-14
Testing NSURLSessionDataTask download progress

cnoon:
This PR adds a test around data task progress monitoring. It is built with the `Given / When / Then` [structure](http://www.objc.io/issue-15/xctest.html) to make it easy to understand the different portions of the test. By gathering all the data and testing it within the `Then` section, it allows us to test all different combinations of the data once it has all been collected.

PR #494, opened 2015-05-20
Created more robust data, upload and download progress tests

cnoon:
This PR adds a new set of tests are monitoring data progress and also revamps the upload and download progress tests.

All tests are passing on both iOS and OS X.

### DIFF
--- a/Tests/DownloadTests.swift
+++ b/Tests/DownloadTests.swift
@@ -35,18 +35,18 @@ class DownloadResponseTestCase: BaseTestCase {
     func testDownloadRequest() {
         // Given
         let numberOfLines = 100
-        let URL = "http://httpbin.org/stream/\(numberOfLines)"
+        let URLString = "http://httpbin.org/stream/\(numberOfLines)"
 
         let destination = Alamofire.Request.suggestedDownloadDestination(directory: searchPathDirectory, domain: searchPathDomain)
 
-        let expectation = expectationWithDescription(URL)
+        let expectation = expectationWithDescription("Download request should download data to file: \(URLString)")
 
         var request: NSURLRequest?
         var response: NSHTTPURLResponse?
         var error: NSError?
 
         // When
-        Alamofire.download(.GET, URL, destination)
+        Alamofire.download(.GET, URLString, destination)
             .response { responseRequest, responseResponse, _, responseError in
                 request = responseRequest
                 response = responseResponse
@@ -97,7 +97,7 @@ class DownloadResponseTestCase: BaseTestCase {
         }
     }
 
-    func testDownloadRequestWithProgress2() {
+    func testDownloadRequestWithProgress() {
         // Given
         let randomBytes = 4 * 1024 * 1024
         let URLString = "http://httpbin.org/bytes/\(randomBytes)"
--- a/Tests/DownloadTests.swift
+++ b/Tests/DownloadTests.swift
@@ -97,27 +97,41 @@ class DownloadResponseTestCase: BaseTestCase {
         }
     }
 
-    func testDownloadRequestWithProgress() {
+    func testDownloadRequestWithProgress2() {
         // Given
-        let numberOfLines = 100
-        let URL = "http://httpbin.org/stream/\(numberOfLines)"
+        let randomBytes = 4 * 1024 * 1024
+        let URLString = "http://httpbin.org/bytes/\(randomBytes)"
 
-        let destination = Alamofire.Request.suggestedDownloadDestination(directory: searchPathDirectory, domain: searchPathDomain)
+        let fileManager = NSFileManager.defaultManager()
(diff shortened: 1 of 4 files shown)


cnoon:
cc: @kcharwood, @kylef, @mattt

cnoon:
This PR replaces #483 so I'm going close it out and delete the remote branch.

Tests/DownloadTests.swift
```diff
@@ -97,27 +97,41 @@ class DownloadResponseTestCase: BaseTestCase {
         }
     }
 
-    func testDownloadRequestWithProgress() {
+    func testDownloadRequestWithProgress2() {
```
kcharwood: I'd get more descriptive with the name.

kcharwood: I'd actually like to hear everyone's testing philosophy here. I've always tried to follow the "test individual things" approach, and have a test for each specific thing you are testing (don't combine asserts for different things in the same test). I dont always end up with tests built that way, but I'm trying.

I've also been told tests are the only place its ok to repeat code over and over, and because you want to test specific things.

What are your thoughts?

mattt: @kcharwood My testing philosophy is to have tests setup situations that precisely emulate feature usage and assert as much as possible about the conditions that follow. Assertions themselves should be atomic (i.e. not testing compound conditions). Having the same assertions over and over across different tests is great—that's diligence. But if setup code is duplicated, there's a strong case to be made to consolidate those into a single test.

mattt: But as far as test names go, they should be descriptive. If there's a name conflict—again, that's a smell of duplication and an argument for consolidation.

cnoon: #### Name Change
I'm on it...good catch!

#### Repeating Yourself in Tests
I agree. I don't think there's any issue with repeating yourself in tests. I think it's advantageous to use a small amount of inheritance for simple things, but otherwise just repeat the logic so the majority of the code resides directly in the test or at least in the same test file. The more test logic that resides directly in the test method, the easier it is to quickly see all the moving parts of the test.

#### Test Individual Things
Great question @kcharwood. I'm going to try to break down my testing approach here and would love everyone's feedback.

I feel that the philosophy of "Test Individual Things" is a bit outdated when applied to one test, one assert. XCTest can continue running your test even if you trip an assertion. It's perfectly fine to have multiple assertions within the same test, it will report them all to you. As long as you have a policy of always keeping your test suite green, why does it matter that you have one test that actually has multiple failures? If your test suite cannot continue if it fails on an assertion, then I completely agree with the philosophy. 

With that said, I do believe there's merit in attempting to write more focused tests. IMO, tests should focus on a particular operation and break it down into as small of unit as possible, then test the results. Next, you create a new set of input parameters, run the operation again, test the results. Rinse and repeat. Testing individual things to me means "test individual operations", not being limited to only have a single assertion in your test.

I also think this philosophy tends to break down with integration tests. Let's take the download tests for example. I just counted 23 different asserts in those two methods, some of which are called upwards of 20 times in a for loop. If we broke this out into individual tests, we'd be looking at thousands of lines of code to test the same logic. When you end up having to repeat yourself that many times, you make copy/paste errors and I think it makes your tests more fragile. I also believe it makes it difficult to understand what is really being tested since almost every test looks the same. Good descriptive names certainly help, but it gets super repetitive to anyone that didn't actually write the tests in the first place.

I could certainly be convinced in those download tests that we should have individual tests more like the following:

1. Successful download request (test completion handler arguments)
1. Failed download request due to bad URL (test completion handler arguments)
1. Failed download request due to bad file path (test completion handler arguments)
1. Successful download request testing only the progress values
1. Failed download request testing only the progress values

It's difficult for me to see how splitting each of these tests up into 4 or 5 additional methods to test each of these parameters would be a "better" overall approach to appease the testing gods. Especially in the case of the progress tests. Additionally, we wouldn't just be able to move the common logic into a "setup" type method, because the majority of the setup parameters get passed into the actual function being tested. While we could pass back a tuple back with upwards of 5 or 6 parameters inside it, I think our tuples would be getting a bit out of control.

One final point that seems a bit strange to me is how we can actually confirm that our single assertion test really did pass successfully if we don't test all the results that we expected? For example, if I make a data request, I expect that the payload data will exist, and I won't receive an error. If I split these into two different tests, how do I know that it is truly working? What if I'm for some reason I am actually receiving both data and an error, and the way I wrote my tests makes it unlikely that I would be able to catch that condition? That may not be the best example, but hopefully the idea of that issue came across. Dynamic operations, such as networking, could benefit from testing all resulting values in the operation rather than a lone expected one to always cover the "unexpected" behaviors.

In summary, I'm all for writing the best tests possible. I'd love everyone's feedback and would love to find a testing strategy that everyone agrees on and is happy with. My two cents...

cnoon: Method name has been fixed and force pushed.

mattt:
Looks good to me!

PR #500, opened 2015-05-28
Updated typo in API Paramater Abstraction in README (#497)

joshuatbrown:
This fixes #497

Issue #497, opened 2015-05-28
Update API Parameter Abstraction section in README


joshuatbrown:
The README contains the following code snippet:

```Alamofire.request(Router.Search(query: "foo bar", page: 1)) // ?q=foo+bar&offset=50```

I found that the resulting query is not `?q=foo+bar&offset=50` as the comment indicates, but rather `?q=foo%20bar&offset=50`.

Perhaps a bigger issue is that as I'm trying to work with the [GitHub Search API](https://developer.github.com/v3/search/), I'm unable to include `+` signs in my query, and GitHub doesn't seem to like the encoded version of `+` (`%2B`), nor the encoded version of space (`%20`).

tobiasoleary:
@joshuatbrown I believe Alamofire is handling URI encoding correctly, but it is very strange that GitHub Search API wouldn't handle decoding the URI correctly. 

You could work around this issue with the GitHub  Search API by implementing a custom encoded (see `ParameterEncoding.swift`), that wraps the URL encoded and replaces the encoded characters that the API can't handle with decoded characters (`%2B` -> `+`, `%20` -> ` `).

joshuatbrown:
@tobiasoleary Thanks for the tip on implementing a custom encoding - I'll go that route for now. I appreciate it!

joshuatbrown:
It looks like #167 is where this was changed. But since the `+` sign wasn't included in parameter encoded values before, should it be?

joshuatbrown:
And suddenly, GitHub Search is working with the encoded `+` sign (`%2B`).

tobiasoleary:
@joshuatbrown Glad GitHub Search started working for you. I've looked into your question.

After reading (When to Encode - Section 2.4 - RFC3986)[http://tools.ietf.org/html/rfc3986#section-2.4] I believe it should be encoded, since Alamofire does build the URI up from its component parts.

joshuatbrown:
@tobiasoleary :+1: I think that issue can be put to rest, then. I still think the docs need to be updated, so perhaps I'll do that now.

tobiasoleary:
@joshuatbrown I worry that my initial assessment is flawed. After reading [Reserved Characters - Section 2.2 - RFC3989](http://tools.ietf.org/html/rfc3986#section-2.2).

Since encoded reserved characters are not necessarily semantically identical to unencoded characters.
[...]

Issue #166, opened 2014-10-10
Should '+' be encoded with the ParameterEncoding.URL?


nvh:
When parsing query parameters, the plus-sign (+) is interpreted as space. Shouldn't it therefore be part of the escaped characters list?

Issue #155, opened 2014-10-07
Double URL Percent Escaping


b3ll:
When performing a POST request like so:

```
Alamofire.request(
  .POST,
  loginURL,
  parameters: [
    "username" : "someusername",
    "pwd" : "coolpassword//"
  ],
  encoding: .URL)
  .responseString { (URLRequest, URLResponse, string, error) in
    println("\(string)")
}
```

The HTTPBody on the URL request escapes the parameters twice when the Method is POST:

```
116: mutableURLRequest.HTTPBody = (CFURLCreateStringByAddingPercentEscapes(nil, query(parameters!) as NSString, nil, nil, CFStringConvertNSStringEncodingToEncoding(NSUTF8StringEncoding)) as NSString).dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)
```

query(parameters!) already escapes the string, and then CFStringConvertNSStringEncodingToEncoding() percent escapes the already escaped string :(

```
coolpassword// 
```

becomes:

```
coolpassword%252F%252F 
```

when it should be:

```
coolpassword%2F%2F
```

Is there any reason why CFURLCreateStringByAddingPercentEscapes() is needed (removing it fixes this problem :smiley: )?

PR #504, opened 2015-05-30
Added parameter encoding characters along with additional documentation

cnoon:
This PR originates from the discussion in Issue #497 and PR #500. I dug through all the past issues related to percent escaping characters and came across the following: #121, #155, #166, #206, #369 and #370. The percent escaping logic has progressed nicely and is solid and well tested.

To this end, I noticed a couple of things that could be improved.

1. It would be good to clarify in the `escape` function which characters we're escaping and why.
1. In the URL parameter encoding tests, give a description of the test cases and which specs they follow along with the exceptions that have been made.

All this PR does is add the "[" and "]" characters as well as split the legal escaped characters into `generalDelimiters` and `subDelimiters`. I think this makes it much easier to understand when referencing [RFC 3986 - Section 2.2](http://tools.ietf.org/html/rfc3986#section-2.2).

The final change was to update the README mismatch which was kindly pointed out by @joshuatbrown in PR #500. I simply corrected the character to `%2B`.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -151,7 +151,11 @@ public enum ParameterEncoding {
     }
 
     func escape(string: String) -> String {
-        let legalURLCharactersToBeEscaped: CFStringRef = ":&=;+!@#$()',*"
+        let generalDelimiters = ":#[]@" // dropping "?" and "/" due to RFC 3986 - Section 3.4
+        let subDelimiters = "!$&'()*+,;="
+
+        let legalURLCharactersToBeEscaped: CFStringRef = generalDelimiters + subDelimiters
+
         return CFURLCreateStringByAddingPercentEscapes(nil, string, nil, legalURLCharactersToBeEscaped, CFStringBuiltInEncodings.UTF8.rawValue) as String
     }
 }
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -150,8 +150,40 @@ public enum ParameterEncoding {
         return components
     }
 
+    /**
+        Returns a percent escaped string following RFC 3986 for query string formatting.
+
+        RFC 3986 states that the following characters are "reserved" characters.
+
+        - General Delimiters: ":", "#", "[", "]", "@", "?", "/"
+        - Sub-Delimiters: "!", "$", "&", "'", "(", ")", "*", "+", ",", ";", "="
+
+        Core Foundation interprets RFC 3986 in terms of legal and illegal characters.
+
+        - Legal Numbers: "0123456789"
+        - Legal Letters: "abcdefghijklmnopqrstuvwxyz", "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        - Legal Characters: "!", "$", "&", "'", "(", ")", "*", "+", ",", "-",
+                            ".", "/", ":", ";", "=", "?", "@", "_", "~", "\""
+        - Illegal Characters: All characters not listed as Legal
+
+        While the Core Foundation `CFURLCreateStringByAddingPercentEscapes` documentation states
+        that it follows RFC 3986, the headers actually point out that it follows RFC 2396. This
+        explains why it does not consider "[", "]" and "#" to be "legal" characters even though 
+        they are specified as "reserved" characters in RFC 3986. The following rdar has been filed
+        to hopefully get the documentation updated.
+
+        - https://openradar.appspot.com/radar?id=5058257274011648
+
+        In RFC 3986 - Section 3.4, it states that the "?" and "/" characters should not be escaped to allow
+        query strings to include a URL. Therefore, all "reserved" characters with the exception of "?" and "/"
+        should be percent escaped in the query string.
+
+        :param: string The string to be percent escaped.
+
+        :returns: The percent escaped string.
+    */
     func escape(string: String) -> String {
-        let generalDelimiters = ":#[]@" // dropping "?" and "/" due to RFC 3986 - Section 3.4
+        let generalDelimiters = ":#[]@" // does not include "?" or "/" due to RFC 3986 - Section 3.4
         let subDelimiters = "!$&'()*+,;="
 
         let legalURLCharactersToBeEscaped: CFStringRef = generalDelimiters + subDelimiters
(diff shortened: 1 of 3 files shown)


cnoon:
cc @kcharwood, @kylef, @mattt

tobiasoleary:
Glad you caught the missing square brackets. 👏

Source/ParameterEncoding.swift
```diff
@@ -151,7 +151,11 @@ public enum ParameterEncoding {
     }
 
     func escape(string: String) -> String {
-        let legalURLCharactersToBeEscaped: CFStringRef = ":&=;+!@#$()',*"
+        let generalDelimiters = ":#[]@" // dropping "?" and "/" due to RFC 3986 - Section 3.4
```
kcharwood: Is it worth having the comment there about dropping? May just want to add it to the release notes.

kcharwood: Also, is it worth adding any specific tests for the square brackets?

cnoon: We already have a bunch of tests around the square brackets cases, so we should be good there. As for removing the comment, I'm cool with taking it out if you think it's not great to have it in there. I want to make sure we have it documented why we shouldn't have the "?" and "/" listed as general delimiters. I'm afraid I might not remember ole RFC 3986 - Section 3.4 to the letter in 6 months. 😔

kcharwood: Ya makes sense. May be better to document why those aren't there, rather than saying "we're dropping them". "We're dropping them sounds more like a temp comment, rather than documentation expected to live there for a long time. Make sense?

And does that mean the square bracket tests were failing before this PR?

cnoon:
Great point about the comment. I'll update it accordingly. 

As for the tests, great question. I already see where you're heading with that. I did some more investigation to figure out exactly why they were passing before without the change, and they're still passing with the change.

After digging in, I found the root issue. The problem is that `CFURLCreateStringByAddingPercentEscapes` claims to adhere to RFC 3986, but considers `[`, `]` and `#` to be illegal characters. It's really difficult to understand why they consider them illegal characters because the spec itself doesn't refer to characters as legal and illegal. Instead it refers to them as reserved and unreserved characters. Additionally, we're not encoding the "entire" URL, but just the query parameters.

I'm going to investigate some more and probably add some additional tests.

cnoon:
Okay so in 3c95af26, I added a bunch of tests and a public docstring on the `escape` method to make things as clear as possible. Now there shouldn't be any question why or how we percent escape our query string characters. 

Here's a fun little chunk of code to run in a playground that demonstrates the "legal" vs. "illegal" ASCII characters from the Core Foundation PoV.

```swift
import Foundation

func escape(string: String) -> String {
    return CFURLCreateStringByAddingPercentEscapes(nil, string, nil, nil, CFStringBuiltInEncodings.UTF8.rawValue) as String
}

let asciiCharacters = [
    " ", "!", "\"", "#", "$", "%", "&", "'", "(", ")", "*", "+", ",", "-", ".", "/", ":",
    ";", "<", "=", ">", "?", "@", "[", "\\", "]", "^", "_", "`", "{", "|", "}", "~",
    "0123456789", "abcdefghijklmnopqrstuvwxyz", "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
]

asciiCharacters.map { println("\"\($0)\" = \"\(escape($0))\"") }

let nonLegalCharacters = " \"#%<>[]\\^`{}|"
```

Which outputs...

```
" " = "%20"
"!" = "!"
""" = "%22"
"#" = "%23"
"$" = "$"
"%" = "%25"
"&" = "&"
"'" = "'"
"(" = "("
")" = ")"
"*" = "*"
"+" = "+"
"," = ","
"-" = "-"
"." = "."
"/" = "/"
":" = ":"
";" = ";"
"<" = "%3C"
"=" = "="
">" = "%3E"
"?" = "?"
"@" = "@"
"[" = "%5B"
"\" = "%5C"
"]" = "%5D"
"^" = "%5E"
"_" = "_"
"`" = "%60"
"{" = "%7B"
"|" = "%7C"
"}" = "%7D"
"~" = "~"
"0123456789" = "0123456789"
"abcdefghijklmnopqrstuvwxyz" = "abcdefghijklmnopqrstuvwxyz"
"ABCDEFGHIJKLMNOPQRSTUVWXYZ" = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```

Source/ParameterEncoding.swift
```diff
@@ -150,8 +150,41 @@ public enum ParameterEncoding {
...
+
+        Core Foundation interprets RFC 3986 in terms of legal and illegal characters.
+        
+        - Legal Numbers: "0123456789"
+        - Legal Letters: "abcdefghijklmnopqrstuvwxyz", "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        - Legal Characters: "!", "$", "&", "'", "(", ")", "*", "+", ",", "-",
+                            ".", "/", ":", ";", "=", "?", "@", "_", "~", "\""
+        - Illegal Characters: All characters not listed as Legal
+
+        For some reason, Core Foundation does not consider "[", "]" and "#" to be "legal" characters
```
kcharwood: rdar worthy?

kcharwood:
:+1:

cnoon:
So for some reason I decided to look at the Core Foundation header this morning instead of Dash and low and behold, the answer was revealed. The header states that the `CFURLCreateStringByAddingPercentEscapes` method follows [RFC 2396](https://www.ietf.org/rfc/rfc2396.txt), not [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) like the Dash documentation states...(sad trumpet plays).

Not only is the mismatch rdar worthy, but it can cause a TON of confusion. This would be especially problematic for people working with IPv6 addresses. Support for them was added to RFC 2396 in the [RFC 2732](https://tools.ietf.org/html/rfc2732) amendment. Both of these were replaced with the latest version of these specs which is RFC 3986.

I think at this point, I'll file a rdar explaining the documentation mismatch and update the docstring accordingly along with adding the openradar link into it.

Source/ParameterEncoding.swift
cnoon: Yup.

PR #505, opened 2015-06-01
Tests all implemented cache policies against common Cache-Control headers

cnoon:
I added tests around common `NSURLCache` usage with `Cache-Control` headers. This is mostly to demonstrate that caching does work properly with Alamofire as well as the expected behavior with the common types of `Cache-Control` headers. These tests were a bit difficult to get written and I'd love everyone's feedback.

With how Apple has implemented this behavior under the hood, it makes it quite difficult to test whether a particular request hit the network or retrieved the response from the cache. The actual negotiation of checking cached response expirations and going to the network is completely hidden from us. You can start to inspect this behavior with something like Charles Proxy, but we don't have that ability from the public API side of the URL Loading System...more details [here](http://andrewmarinov.com/ioss-corenetwork-lying/). Therefore, the only thing that I was able to use to determine whether a response came from the cache or from the network was to check the response's timestamp. By introducing a 1 second delay, I was able to always confirm whether a response came from the cache or the network.

The initial commit demonstrates my first pass at the tests. This approach was pretty solid, but was too slow. Each test introduced two network calls with a second wait between them. This doesn't scale very well and we'd love to keep our testing times down as much as possible.

The second commit results in the same tests, but optimized for speed and readability. I've added a pretty hefty amount of documentation around the tests to explain how they work. While the code is fairly self documenting, it's still complicated how everything gets set up.

With the second pass, I was able to condense the entire set of cache policy tests into single tests. This makes it a bit more difficult to debug the tests, but makes them run much faster and is much easier to understand what the test is actually testing.

As mentioned before, I'd love everyone's feedback.

### DIFF
--- a/Tests/CacheTests.swift
+++ b/Tests/CacheTests.swift
@@ -1,0 +1,501 @@
+// CacheTests.swift
+//
+// Copyright (c) 2014–2015 Alamofire Software Foundation (http://alamofire.org/)
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Alamofire
+import Foundation
+import XCTest
+
+/**
+    The cache test cases test various NSURLRequestCachePolicy types against different combinations of
+    Cache-Control headers. These tests use the response timestamp to verify whether the cached response
+    data was returned. This requires each test to have a 1 second delay built into the second request
+    which is not ideal.
+*/
+class CacheTestCase: BaseTestCase {
+
+    // MARK: Properties
+
+    let URLString = "http://httpbin.org/response-headers"
+    var manager: Manager!
+    var URLCache: NSURLCache { return self.manager.session.configuration.URLCache! }
+    var requestCachePolicy: NSURLRequestCachePolicy { return self.manager.session.configuration.requestCachePolicy }
+
+    // MARK: Setup and Teardown
+
+    override func setUp() {
+        super.setUp()
+        // No-op
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        self.URLCache.removeAllCachedResponses()
+    }
+
+    // MARK: Test Setup Methods
+
+    func setUpManagerWithRequestCachePolicy(requestCachePolicy: NSURLRequestCachePolicy) {
+        self.manager = {
+            let configuration: NSURLSessionConfiguration = {
+                let configuration = NSURLSessionConfiguration.defaultSessionConfiguration()
+                configuration.HTTPAdditionalHeaders = Alamofire.Manager.defaultHTTPHeaders
+                configuration.requestCachePolicy = requestCachePolicy
+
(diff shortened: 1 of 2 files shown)


cnoon:
cc @kcharwood, @kylef, @mattt

Tests/CacheTests.swift
```diff
@@ -0,0 +1,333 @@
...
+    func testUseLocalCacheDataAndDontLoadFromNetworkPolicy() {
+        let cachePolicy: NSURLRequestCachePolicy = .ReturnCacheDataDontLoad
+
+        executeTest(cachePolicy: cachePolicy, cacheControl: CacheControl.Public, shouldReturnCachedResponse: true)
+        executeTest(cachePolicy: cachePolicy, cacheControl: CacheControl.Private, shouldReturnCachedResponse: true)
+        executeTest(cachePolicy: cachePolicy, cacheControl: CacheControl.MaxAgeNonExpired, shouldReturnCachedResponse: true)
+        executeTest(cachePolicy: cachePolicy, cacheControl: CacheControl.MaxAgeExpired, shouldReturnCachedResponse: true)
+        executeTest(cachePolicy: cachePolicy, cacheControl: CacheControl.NoCache, shouldReturnCachedResponse: true)
+
+        #if os(OSX)
```
kcharwood: Can you explain the difference between OS X and iOS?

Tests/CacheTests.swift
```diff
@@ -0,0 +1,333 @@
...
+        let noStoreResponse = self.URLCache.cachedResponseForRequest(noStoreRequest)
+
+        // Then
+        XCTAssertNotNil(publicResponse, "\(CacheControl.Public) response should not be nil")
+        XCTAssertNotNil(privateResponse, "\(CacheControl.Private) response should not be nil")
+        XCTAssertNotNil(maxAgeNonExpiredResponse, "\(CacheControl.MaxAgeNonExpired) response should not be nil")
+        XCTAssertNotNil(maxAgeExpiredResponse, "\(CacheControl.MaxAgeExpired) response should not be nil")
+        XCTAssertNotNil(noCacheResponse, "\(CacheControl.NoCache) response should not be nil")
+
+        #if os(OSX)
```
kcharwood: Can you explain the difference between OS X and iOS?

cnoon:
Looks like I've still got some work to do here. I'll investigate this tomorrow as to why we're seeing failures on Travis as well as why NSURLCache has different behavior for the `no-store` header value between platforms.

Tests/CacheTests.swift
cnoon: I can now that I've gotten to the bottom of it! 😛

It appears that there was a bug in the iOS 8.2 simulator which I was using when I wrote the tests. Switching over to the iOS 8.3 simulator fixed the issues and we're now seeing the same behavior between iOS and OS X. Thankfully both are producing the "expected" behavior as well at this point since a `Cache-Control` header with a `no-store` value certainly shouldn't cache that request. FWIW, I'm also seeing the wrong behavior in the iOS 8.1 simulator as well.

From what I can tell by scouring the Internets is that people have been reporting all sorts of caching inconsistencies across iOS 8 (including against Alamofire itself). It looks like Apple may have finally gotten iOS 8.3 back in line. I think for now, it's probably best to just move forward. I don't think it makes sense to file a rdar for something they've already fixed. I also don't think it makes a ton of sense to write tests to try to test the older broken behavior. At the moment, we're running Travis against the latest iOS Simulator and OS X x86_64 architecture.

My approach here was to simply get tests around the expected NSURLCache behavior on both platforms, and make sure we stay green here moving forwards. Does that sound reasonable?

cnoon:
Okay, problem solved...Travis is happy!

cnoon:
I can now that I've gotten to the bottom of it! 😛

It appears that there was a bug in the iOS 8.2 simulator which I was using when I wrote the tests. Switching over to the iOS 8.3 simulator fixed the issues and we're now seeing the same behavior between iOS and OS X. Thankfully both are producing the "expected" behavior as well at this point since a `Cache-Control` header with a `no-store` value certainly shouldn't cache that request. FWIW, I'm also seeing the wrong behavior in the iOS 8.1 simulator as well.

From what I can tell by scouring the Internets is that people have been reporting all sorts of caching inconsistencies across iOS 8 (including against Alamofire itself). It looks like Apple may have finally gotten iOS 8.3 back in line. I think for now, it's probably best to just move forward. I don't think it makes sense to file a rdar for something they've already fixed. I also don't think it makes a ton of sense to write tests to try to test the older broken behavior. At the moment, we're running Travis against the latest iOS Simulator and OS X x86_64 architecture.

My approach here was to simply get tests around the expected NSURLCache behavior on both platforms, and make sure we stay green here moving forwards. Does that sound reasonable?

kcharwood:
Awesome work.

Everything looks good here. I'd say the only issue is that now the tests only pass on 8.3 and up. I'd be all for getting travis to test on all supported test targets, as to get more coverage across all of the supported OS's. That would mean having some branching tests here to account for that problem in iOS 8.1 and 8.2, but I think its worth today to provide support for the rest of the library across other iOS versions.

I recently updated the [rakefile](https://github.com/AFNetworking/AFNetworking/blob/master/Rakefile) in AFNetworking, which shows how I am grabbing those simulators out and passing them in as test targets. Not sure if you can do the same thing here, but it would probably help you sleep better :beers:

cnoon:
That sounds good @kcharwood. I'm going to merge this PR and start working on a separate PR to add simulators for all our supported OS's as well as branch the necessary tests.

PR #512, opened 2015-06-05
Added .stream callback to receive streaming data responses

psobot:
One feature that Alamofire currently lacks is the ability to receive data responses from servers incrementally. This can be useful in a number of cases, such as when displaying partial data to a user or streaming data without a known or finite length.

This PR adds a `.stream` function to `Request`, allowing clients of Alamofire to incrementally receive data from servers as it arrives. When using the `.stream` function, data is no longer accumulated into the `responseObject` object, as streams may be of unbounded length. (It's up to the calling code to join together the `NSData` objects returned in the `.stream` callback if the full response's data is needed.)

This PR also includes a test case to verify the behaviour of the `.stream` function, and to ensure that it doesn't conflict with the `.progress` functionality that already exists on `Request`.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -116,6 +116,24 @@ public class Request {
         return self
     }
 
+    /**
+    Sets a closure to be called periodically during the lifecycle of the request as data is read from the server.
+    This closure returns the bytes most recently received from the server, not including data from in previous calls.
+    If this closure is set, data will only be available within this closure, and will not be otherwise saved. 
+    (The .response closure will be called with nil responseData.)
+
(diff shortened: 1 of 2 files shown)


cnoon:
Thank you for this PR @psobot! My apologies for not getting back to you sooner. Our TC has been really busy this week at WWDC. I'm really excited about this change but haven't had enough time to go through it in depth. I should be able to get to it here in the next few days.

Thank you for your patience! 🍻

cnoon:
This is great work @psobot! Thank you very much for your PR. I have pulled it into master in a combination of the following commits: a15dc1d6, 93df9fe2 and 922c1ca3.

PR #516, opened 2015-06-08
Only set cookies in debugPrintable when relevant

alexpls:
If `HTTPShouldSetCookies` is set to `false` in the session's config then any cookies set will not be sent as part of the request.

Previously the cURL command output provided by `debugPrintln()` would include cookies whether or not `HTTPShouldSetCookies` was true or false. This commit will make it so `debugPrintln()` only prints cookies in the cURL output if they'll actually be sent.


cnoon:
Thank you for your PR @alexpls. I accepted your change with slight modifications in c40d7907.

alexpls:
Great! Thanks @cnoon

PR #523, opened 2015-06-11
Setting right availability guard for iOS, fixing build for Xcode 7

coverback:
Without this fix, Alamofire doesn't compile for iOS 7 targets with Xcode 7.

Test project doesn't reveal this because it's iOS 8 and up, but the same guard as for 10.10 is needed for iOS as well.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -217,7 +217,7 @@ public class Request {
                 operationQueue.maxConcurrentOperationCount = 1
                 operationQueue.suspended = true
 
-                if #available(OSX 10.10, *) {
+                if #available(OSX 10.10, iOS 8.0, *) {
                     operationQueue.qualityOfService = NSQualityOfService.Utility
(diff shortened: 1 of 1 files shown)


cnoon:
Thanks for the PR @coverback. It is much appreciated!

The `swift-2.0` does not include this check on purpose. We are going to be dropping iOS 7 support here very soon. The reason is that we cannot properly test against iOS 7 in Travis-CI without rewriting our entire test suite. The separate test suite would need to import all the Swift files (since you cannot import frameworks in iOS 7) into the target and all the tests would need to be re-written and duplicated. Since this would be a huge effort and very difficult to maintain, we've decided that it is in the best interest of the project to keep moving forward and not bring iOS 7 support along with us.

You will see more formal updates about these changes take place in the README on the `swift-2.0` branch as well as `master` here in a future release.

Thanks again.

PR #525, opened 2015-06-12
fixing a missing #available flag (needed for iOS 7.0 targets..)

mishagray:
Your iOS target level is set to iOS 8 in the project, so it doesn't see this compiler error.  (Since it makes a dynamic framework).

For those of us still "dropping" the swift into the project without pods or carthage (I REALLY want to drop iOS 7.0 support asap) than you won't see this.

You may want to think about adjusting the IOS target level in the project also, so you can spot any more of these in future.  I THINK it's ok, even though the dynamic framework can't run on iOS 7.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -217,7 +217,7 @@ public class Request {
                 operationQueue.maxConcurrentOperationCount = 1
                 operationQueue.suspended = true
 
-                if #available(OSX 10.10, *) {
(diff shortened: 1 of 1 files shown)


cnoon:
Thanks for your PR @mishagray. We will not be supporting iOS 7 in the `swift-2.0` branch. It too difficult to maintain multiple test targets to ensure the quality of that OS. Please see #523 for more details.

Cheers.

PR #529, opened 2015-06-15
I want to resolve #528, suggestions are welcome

### LINKED ISSUE #528: Validation errors do not provide any information
For any validation, the error reported to the handler is:

    NSError(domain: AlamofireErrorDomain, code: -1, userInfo: nil)

When using validation together with responseJSON (for example) the failure reason is unknown to the response handling block. This causes the validation to be duplicated when handling the error:

    let request = Alamofire.request(method, url, parameters: parameters, encoding: encoding)
        .responseJSON { request, response, json, error in
            if contains(200..<300, response.statusCode) { ... }

Also, the response data is not available to the handler block in case of a validation error.

berikv:


cnoon:
Thank you for your PR @berikv. As I mentioned in #528, this is something we're going to take a much larger approach to solving properly. Thanks again.

PR #530, opened 2015-06-15
Updated to Swift 2.0 Changes

terhechte:
Just stumbled upon this when I was using Alamofire with one of my Swift 2.0 projects, and thought I'd do a quick fork.


cnoon:
Looks like I missed the playground in the conversion. Thanks @terhechte!

Issue #447, opened 2015-04-22
Better errors from .validate method.


tobiasoleary:
Currently if validation fails the error object seems to always be `NSError(domain: AlamofireErrorDomain, code: -1, userInfo: nil)`.

It would be nice if the `validator` would return something besides a `Bool` (maybe `NSError?`) so that if the validation failed because of a 404 the error object given to the response handler might have a description of "Invalid Status Code - 404 File Not Found".

allaire:
@tobiasoleary @cnoon Is there also a way to capture the body content of a failing request?

```
Alamofire.request(Router.ReadOrganizations).validate().responseSwiftyJSON({(request, response, json, error) in
    if error == nil {
        self.performSegueWithIdentifier("SigninToMySchedule", sender: self)
    } else {
        println(response?.statusCode)
        println(error)
    }
})
```

With the code above, when an error happens, `error` (`com.alamofire.error Code=-1`) is meaningless and json is nil. At least we have access to the statusCode. Is this by design @mattt ?

tobiasoleary:
After thinking about this. I think by design to the keep the `Validator` type easy to understand and create. The client should write their own error messages. On the other hand it would be nice to know which Validator failed, just to narrow down what the issue could be.

So I have been working with an API that sends HTML error pages when it fails. This is useful debugging information for me, so I've been using both `responseString` and `responseJSON` completionHandler. 

See example below.
```
final class Box<T> {
    private let value: T
    
    var unbox: T {
        return value
    }
    
    init(_ initValue: T) {
        value = initValue
    }
}

enum Result<SuccessType> {
    case Success(Box<SuccessType>)
    case Failure(NSError)
}

func liveUpdate(#teamIdentifier: String, filter: String, sinceIdentifier: String, completionHandler: (Result<LiveUpdateData>) -> Void) {
        
        Alamofire.request(Router.LiveUpdate(teamIdentifier, filter, sinceIdentifier))
            .validate()
            .responseString {
                (request, response, responseString, error) in
                if let error = error {
                    Debug.printRequest(request, response: response, responseString: responseString, error: error)
                }
            }.responseJSON {
                (request, response, JSON, error) in
                
                if let error = error {
                    completionHandler(.Failure(error))
                } else if let JSON = JSON as? [String : AnyObject], let liveUpdateData = Mapper<LiveUpdateData>().map(JSON) {
                    completionHandler(Result.Success(Box(liveUpdateData)))
                } else {
[...]

PR #534, opened 2015-06-17
Validation returns an NSError? too

marcelofabri:
While #447 involves a bigger change (even supporting Swift 2.0), this just change `Validation` to return a tuple `(Bool, NSError?)`, so validators can also inform an error.

### DIFF
--- a/Source/Validation.swift
+++ b/Source/Validation.swift
@@ -22,12 +22,15 @@
 
 import Foundation
 
+public let AlamofireErrorInvalidStatusCode = 1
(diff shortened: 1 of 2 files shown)


cnoon:
Thank you for your PR @marcelofabri. 

We are going to revisit all error handling logic for the Alamofire 2.0.0 release. We'll keep these changes in mind when we are considering alternatives for Validation handling, but for now I'm going to close out this PR since this would not be backwards compatible.

Cheers 🍻

PR #539, opened 2015-06-22
Feature - MultipartFormData

cnoon:
This pull request adds support for uploading `MultipartFormData` to Alamofire. I would love feedback from the community to make sure everything here is in tip-top shape before rolling it in. Any feedback would be much appreciated.

## MultipartFormData

The general idea is that you append body parts to the larger MultipartFormData object, then encode it when you want the cumulative result. Encoding can be performed in two different ways:

1. Encoding In-Memory
1. Encoding Data by Streaming to Disk

### Encoding In-Memory

Encoding in-memory should only be used with small datasets. If the `contentLength` of the `MultipartFormData` is too large, you run the risk of your app being terminated due to memory usage.

### Encoding Data by Streaming to Disk

Sometimes you need to upload a very large dataset to a server such as a bundle of large images or videos. In these cases, your dataset is too large to perform the encoding in-memory. For these cases, the `MultipartFormData` class supports encoding the data using a combination of input and output streams keeping the memory footprint very small.

## Uploading from Manager

The `upload` methods for uploading `MultipartFormData` support the idea of a memory threshold. To make it as easy as possible, the memory threshold is used to determine whether the `MultipartFormData` can be encoded in-memory or whether it needs to be written to disk first. The default value for the memory threshold is currently set to 10MB.

In order to make the encoding process asynchronous, the upload methods could not return a `Request` object directly. To handle this case, the `MultipartFormDataEncodingResult` enumeration was added and exposed in an `encodingCompletion` closure. The resulting `Request` object is contained as an associated value in the `.Success` case which can then be used to chain additional methods against. While this is a bit different than the other public APIs, it still is quite simple to use.

```swift
Alamofire.upload(
    .POST,
    URLString: "http://httpbin.org/post",
    multipartFormData: { multipartFormData in
        multipartFormData.appendBodyPart(fileURL: unicornImageURL, name: "unicorn")
        multipartFormData.appendBodyPart(fileURL: rainbowImageURL, name: "rainbow")
    },
    encodingCompletion: { encodingResult in
    	switch encodingResult {
    	case .Success(let upload, _, _):
    		upload.responseJSON { request, response, JSON, error in
    			println(JSON)
    		}
    	case .Failure(let encodingError):
    		println(encodingError)
    	}
    }
)
```

## Summary

This was a fairly complicated feature to put together, so I'd really love to gather everyone's feedback. If you have a solid grasp of RFC 2388 and 2045, then please take the time to review these changes before we merge it in.


cnoon:
cc @kylef and @kcharwood

tobiasoleary:
Great job!

omarps:
Nice!

dyguests:
:+1: Great!

cnbuff410:
:+1:

nickvelloff:
:+1:  :100:

tangamampilia:
Thanks. It's possible to append regular POST vars?

Source/MultipartFormData.swift
```diff
@@ -0,0 +1,624 @@
...
+        while inputStream.hasBytesAvailable {
+            var buffer = [UInt8](count: self.streamBufferSize, repeatedValue: 0)
+            let bytesRead = inputStream.read(&buffer, maxLength: self.streamBufferSize)
+
+            if inputStream.streamError != nil {
+                error = inputStream.streamError
+                break
+            }
+
+            if bytesRead < 0 {
```
kylef: @cnoon I think we should handle when the bytesRead is returned to 0. Since there might not actually be data to read even though `hasBytesAvailable` is `true`.

`hasBytesAvailable` can be true when a read must be made to determine if their are bytes available, in which case read will return 0 when there are no actual bytes available.

See these statements from the documentation:

> `hasBytesAvailable`: If you want to be semantically compatible with `NSInputStream`, return `YES` if a read must be attempted to determine if bytes are available.

> `read:maxLength:`: `0` indicates that the end of the buffer was reached;

So something like:

```swift
if bytesRead > 0 {
    encoded.appendBytes(buffer, length: bytesRead)
} else if bytesRead < 0 {
    let failureReason = "Failed to read from input stream: \(inputStream)"
    error = errorWithCode(AlamofireInputStreamReadFailed, failureReason: failureReason)
    break
} else {
    break
}
```

cnoon: That's a great catch @kylef. I agree completely. I'll get this method fixed up and also apply this approach to the other streaming methods as well.

kamoljan:
+1 
Can you merge it with `swift-2.0` branch?

cnoon:
We will definitely merge this into the swift-2.0 branch once this gets merged into master @kamoljan.

kamoljan:
Hi @cnoon!
I am looking forward to seeing it soon.
Thank you!

Source/MultipartFormData.swift
cnoon: Changes made in 14f74c3a @kylef.

morizotter:
:+1:

kaelite:
awesome!

susieyy:
Great ! ! !

orschaef:
Just tried it out. Maybe I missed something but:
The "**Content-Type**" in my multipart POST header is **not** set to

"**multipart/form-data; boundary=alamofire.boundary...**"

which is what I need.
Is there any elegant way to achieve that? The only workaround I found was:

```swift
Alamofire.upload(
      request,
      multipartFormData: { multipartFormData in
        request.setValue("multipart/form-data; boundary=\(multipartFormData.boundary)", forHTTPHeaderField: "Content-Type")
        multipartFormData.appendBodyPart(fileData: fileData, name: "file", fileName: "asset.png", mimeType: "image/png")
      }, ...
```

kaelite:
Have the same problem with orschaef, how to get "multipart/form-data; boundary=alamofire.boundary..." in post request?

Source/Upload.swift
```diff
@@ -156,6 +156,150 @@ extension Manager {
...
+            multipartFormData(formData)
+
+            if formData.contentLength < encodingMemoryThreshold {
+                let encodingResult = formData.encode()
+
+                dispatch_async(dispatch_get_main_queue()) {
+                    switch encodingResult {
+                    case .Success(let data):
+                        let encodingResult = MultipartFormDataEncodingResult.Success(
+                            request: self.upload(URLRequest, data: data),
```
kylef: @cnoon I think this is missing a step to take the `contentType` property from `MultipartFormData` and set it on the actual request going out (and also a test for this).

minhoryang:
+1

cnoon:
Thanks @orschaef and @kaelite. That's totally my bad. I simply forgot to set the `Content-Type` on the `URLRequest`. I'll have it all fixed up shortly and will also adds tests around it @kylef.

Source/Upload.swift
cnoon: `Content-Type` header is now being set in f1557414.

cnoon:
@orschaef and @kaelite...the `Content-Type` issue should be resolved with f1557414. If you could give it a go I would really appreciate it.

orschaef:
@cnoon Seems to work now. Thank you! Glad we could help!

cnoon:
Thank you for confirming @orschaef. I'm going to go ahead and merge this. Thank you everyone for all your help getting this feature tightened up! 🎉 🎉 🎉

damienrambout:
Awesome!

I'm just a little bummed about the multipart `upload` functions that do not return an (optional) `Request` synchronously, instead of through the `encodingCompletion` block. (AFNetworking would return a `NSURLSession` synchronously when using `POST:parameters:constructingBodyWithBlock:success:failure`).
I guess this is due to internal implementation reasons and there's nothing you can do about it.

cnoon:
Hi @damienrambout, thanks for your feedback. 

Just for anyone else reading through this, I'd like to clarify a bit. The `upload` function doesn't return a `Request` directly because it needs to encode the `MultipartFormData` before creating the `Request`. Since the encoding operation can be very small or extremely large, you really don't want to block the caller thread while executing this. Therefore, the encoding is asynchronous and the `encodingCompletion` closure returns the `Request`. This is all intentional and by design.

A second approach in AFNetworking is to subclass `NSInputStream` and provide a dynamic upload stream rather than encoding it prior to the upload. While this is more convenient, it is extremely error prone. Additionally, subclassing `NSInputStream` is a very difficult and quite finicky. It's backed by an opaque type that needs to have certain methods overridden when toll-free bridging. It's just not a very robust or maintainable solution.

Our goal with Alamofire is to provide the absolute highest quality networking layer on top of the URL Loading System that we possibly can. Therefore, sometimes we will choose to implement certain features differently than AFNetworking to continue to improve the overall quality of its successor, Alamofire.

ivangodfather:
how can i append other vars to the POST?

cnoon:
@ivangodfather you can either append them as query parameters to an `NSURLRequest` or you can add them as multipart form data values using the following method:

```swift
func appendBodyPart(#data: NSData, name: String)
```

ivangodfather:
Would be like this? 

let myInt = 3
let myString = "test"               
 multiPart.appendBodyPart(data: String(my_int).dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: true)!, name: "my_int")

                multiPart.appendBodyPart(data: myString.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: true)!, name: "my_string")

ivangodfather:
and for an image:                        
    data.appendBodyPart(fileData: UIImagePNGRepresentation(image), name: "image", fileName: "random_name.png", mimeType: "image/png") ?

cnoon:
You got it!

robindoble:
Hi @cnoon - just tried implementing your code snippet and am getting a .POST error, "Could not find member .POST", not having this issue with other Alamofire methods, including the single file upload function. Any thoughts?

cnoon:
Hi @robindoble, please check out the test [logic](https://github.com/Alamofire/Alamofire/blob/master/Tests/UploadTests.swift#L134-L552). My guess is that you just have a syntax issue. If you do end up finding a mistake in an example code snippet in the README or something like that, please submit a PR or file a new issue. Thanks!

ivangodfather:
@cnoon and why i've to specify in my Podifle the branch master since its the default branch?
        pod 'Alamofire', :git => 'https://github.com/Alamofire/Alamofire.git', :branch => 'master'

Otherwise, if i only write pod 'Alamofire' i dont get the methods with multipart.

robindoble:
@cnoon, now working, many thanks @ivangodfather

ivangodfather:
@robindoble  yeah, happened to me,  forgot to mention you, sorry, btw, isnt supposed to be the default branch?

robindoble:
@ivangodfather @cnoon - whats the best way to set custom headers on multipart requests, I saw this in a thread above, wondering if there is a best practice?

request.setValue("multipart/form-data; boundary=\(multipartFormData.boundary)", forHTTPHeaderField: "Content-Type")

@ivangodfather - can you clarify your last post, are you suggesting podfile should be :branch => 'default' instead?

ivangodfather:
@robindoble i'm asking why if the patch is in the "master" branch i have i
specify branch => 'master' since when i do pod 'Alamofire' my guess is that
the default branch is "master".

On 13 July 2015 at 19:48, robindoble <notifications@github.com> wrote:

> @ivangodfather <https://github.com/ivangodfather> @cnoon
> <https://github.com/cnoon> - whats the best way to set custom headers on
> multipart requests, I saw this in a thread above, wondering if there is a
> best practice?
>
> request.setValue("multipart/form-data;
> boundary=(multipartFormData.boundary)", forHTTPHeaderField: "Content-Type")
>
> @ivangodfather <https://github.com/ivangodfather> - can you clarify your
> last post, are you suggesting podfile should be :branch => 'default'
> instead?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Alamofire/Alamofire/pull/539#issuecomment-121003963>.
>



-- 
Este mensaje se dirige a su destinatario y puede contener información
confidencial sometida a secreto profesional o cuya divulgación esté
prohibida por la Legislación Vigente. Cualquier persona distinta a su
destinataria tiene prohibida su reproducción, uso, divulgación o impresión
total o parcial. Si ha recibido este mensaje por error, notifíquelo de
inmediato al remitente borrando el mensaje original juntamente con sus
ficheros anexos. Gracias.

robindoble:
@ivangodfather - thats one for @cnoon, the author ;) @cnoon - would appreciate any advice on setting custom headers please.

mcudich:
Now that this is in master, is there any chance you'd be willing to merge it into the swift-2.0 branch?

cnoon:
Hi @robindoble, the `Content-Type` header is set automatically for you. If you need to set additional headers, you can either set them in the `NSURLSessionConfiguration` when you create the `Manager` instance, or you can set them on the `NSURLRequest` before you pass it into the `upload:multipartFormData` method as a `URLRequestConvertible`.

allaire:
Just want to add my voice that this is working well in our app and we could ditch the custom method we used :+1:

SteveNewhouse:
I'm migrating from AFNetworking and have been struggling to add a dictionary to a multipart request so that it ends up being encoded in the same way it would be with an AFNetworking request like this:

```swift
     AF.POST(url, parameters: user, constructingBodyWithBlock: { (formData: AFMultipartFormData!) in
        if imageData != nil {
            formData.appendPartWithFileData(imageData!, name: "avatarImage", fileName: "blah", mimeType: "image/jpeg")
        }
    },
    ...
```
I've tried it like so, but the backend isn't happy with the encoding (i.e. something is different):

```swift
    Alamofire.upload(Router.SignUp(user, imageData),
        multipartFormData : { multipartFormData in
            if imageData != nil {
                multipartFormData.appendBodyPart(data: imageData!, name: "avatarImage", fileName: "blah", mimeType: "image/jpeg")
            }

        var serializedUser = ""
        for (key, value) in user {
            serializedUser += "user[\(key)]=\(value)"
	    }
	    multipartFormData.appendBodyPart(data:    serializedUser.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: true)!, name: "user")
    }
````

Anything jump out as being wrong here?

**Comment by ivangodfather** [comment]
data.appendBodyPart(data: dateString.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: true)!, name: "scheduled_closing_date")
                    for image in images {
                        data.appendBodyPart(fileData: UIImageJPEGRepresentation(image, 0.6),name: randomStringWithLength(8), fileName: randomStringWithLength(8) + ".jpg", mimeType: "image/jpeg")
                    }

where data is multipartFormData in your case

**Comment by SteveNewhouse** [comment]
But how do I deal with a dictionary? That's what I can't recreate in Alamofire upload

**Comment by ivangodfather** [comment]
You need to transform it as a JSON ( https://github.com/Hearst-DD/ObjectMapper ) and send it with 
let params = ["user":YOUR_JSON_USER]
request(.POST, YOUR_URL, parameters: params, encoding: .JSON)

**Comment by cnoon** [comment]
That's not how to go about this @ivangodfather.

**Comment by cnoon** [comment]
@SteveNewhouse you want to use the `ParameterEncoding.queryComponents` API to help you out here. Here's some sample code to get you on track:

```swift
let imageData: NSData? = NSData()
let URLRequest = NSURLRequest(URL: NSURL(string: "https://httpbin.org/post")!)

Alamofire.upload(URLRequest,
    multipartFormData : { multipartFormData in
        if imageData != nil {
            multipartFormData.appendBodyPart(data: imageData!, name: "avatarImage", fileName: "blah", mimeType: "image/jpeg")
        }

        let user: [String: String] = [:]

        for (key, value) in ParameterEncoding.URL.queryComponents("user", user) {
            multipartFormData.appendBodyPart(data: value.dataUsingEncoding(NSUTF8StringEncoding)!, name: key)
        }
    },
    encodingCompletion: { result in
        // todo
    }
)
```

SteveNewhouse:
Excellent- thank you Christian

PR #541, opened 2015-06-22
Update README.md

rojotek:
Added link to the documentation.

Given that complete documentation is one of the touted features, it seems to make sense to link to it.

It also makes it much easier to get to for the occasional swift dev (like myself :)).


cnoon:
While that was not the original intent of the feature list, I can't think of any reason not to include it. Thanks for the PR @rojotek!

PR #542, opened 2015-06-22
Lint full podspec on travis CI

kylef:
We originally linted the as quick because Travis CI didn't have a version of Xcode / CocoaPods that was compatible with this before.

Changing to full will mean that CocoaPods will build the project on the lowest supported platforms and ensures all the sources work together.


cnoon:
Looks good dude!

PR #557, opened 2015-06-30
[Refactoring] Replace many of force unwrapping usages with optional binding and where clause

ikesyo:


Source/Download.swift
```diff
@@ -150,8 +150,8 @@ extension Request {
         // MARK: Delegate Methods
 
         func URLSession(session: NSURLSession, downloadTask: NSURLSessionDownloadTask, didFinishDownloadingToURL location: NSURL) {
-            if downloadTaskDidFinishDownloadingToURL != nil {
-                let destination = downloadTaskDidFinishDownloadingToURL!(session, downloadTask, location)
+            if let handler = downloadTaskDidFinishDownloadingToURL {
+                let destination = handler(session, downloadTask, location)
```
cnoon: In all the places where you are replacing force unwrapping with optional binding, could you use the following format instead?

```swift
if let downloadTaskDidFinishDownloadingToURL = self. downloadTaskDidFinishDownloadingToURL {
    let destination = downloadTaskDidFinishDownloadingToURL(session, downloadTask, location)
}
```

We have a open task for replacing all property accesses with the `self.` prefix for clarity. This would be a good place to start that work on all these changes.

Source/Validation.swift
```diff
@@ -40,10 +40,8 @@ extension Request {
     */
     public func validate(validation: Validation) -> Self {
         delegate.queue.addOperationWithBlock {
-            if self.response != nil && self.delegate.error == nil {
-                if !validation(self.request, self.response!) {
-                    self.delegate.error = NSError(domain: AlamofireErrorDomain, code: -1, userInfo: nil)
-                }
+            if let response = self.response where self.delegate.error == nil && !validation(self.request, response) {
+                self.delegate.error = NSError(domain: AlamofireErrorDomain, code: -1, userInfo: nil)
```
cnoon: Nice!

cnoon:
Thanks for this PR @ikesyo! I have just a single comment that will require a few changes throughout the PR. If you could get those changes made, I'll gladly merge it.

Thank you for taking the time to clean up the logic.

ikesyo:
@cnoon Thank you for the review! I've fixed the format what you pointed and rebased to master. Could you check it?

cnoon:
Great work @ikesyo...thanks for the changes! 🍻

PR #558, opened 2015-07-01
Execute the taskDidComplete closure in addition to the default implementation

oanhof:
We ran into the problem that when the taskDidComplete closure was set the response handlers were never executed because the closure replaced the default implementation.

This change fixed the problem for us and we think it would be useful for everyone.

### DIFF
--- a/Source/Manager.swift
+++ b/Source/Manager.swift
@@ -277,9 +277,8 @@ public class Manager {
         }
 
         public func URLSession(session: NSURLSession, task: NSURLSessionTask, didCompleteWithError error: NSError?) {
-            if taskDidComplete != nil {
-                taskDidComplete!(session, task, error)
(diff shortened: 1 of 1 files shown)


cnoon:
Thank you for your PR @oanhof.

The `taskDidComplete` closure is exposed to allow you to completely override the default behavior if you so choose. It was not intended to allow you to hook into the `SessionDelegate` callbacks for notifications and still execute the default behavior.

As an alternative, you can use the `response` serializers to know once a task was completed. Additionally, we plan to open up more of the `Request` class in the 1.3.0 release to allow you to extend a `Request` to have additional hooks along with the already provided ones such as `validate`, `authenticate`, `response`, `responseJSON`, etc.

Hopefully this helps clarify things a bit. Cheers.

PR #569, opened 2015-07-08
Multipart upload: support to upload non-file data like JSON string from the form input

top2tech:


Source/MultipartFormData.swift
```diff
@@ -266,6 +266,29 @@ public class MultipartFormData {
...
+    
+    - `Content-Disposition: form-data; name=#{name};mimeType=#{mimeType}` (HTTP Header)
+    - Encoded data
+    - Multipart form boundary
+    
+    :param: data The data to encode into the multipart form data.
+    :param: name The name to associate with the data in the `Content-Disposition` HTTP header.
+    :param: mimeType The MIME type to associate with the data in the `Content-Type` HTTP header.
+    */
+    public func appendBodyPart(#data: NSData, name: String, mimeType: String) {
```
kylef: Perhaps it would make sense to slighly refactor `appendBodyPart(fileData data: NSData, name: String, fileName: String, mimeType: String) ` to make the fileName optional. When omitted it has this behaviour.

/c @cnoon

cnoon: I was actually thinking the same thing @kylef. The more I think about it though, wouldn't you always want to provide a filename as a hint to the server? We could certainly make the `fileName` parameter optional, but is there really a realistic case where that is what the client should do?

cnoon:
Hi @top2tech, @kylef and I spoke offline and both agree that this body part structure is something that we haven't seen before in any API / service. Could you provide us with some example APIs or usages to help us better understand your proposed changes? 

Otherwise I'm afraid we'll have to reject this change.

cnoon:
@top2tech could you get back to us on this? I'm going to close this PR out tomorrow if I don't hear back from you. Cheers.

cnoon:
Hi @top2tech, I'm going to close this PR out since I haven't heard back from you. If you really want to re-open this, please provide answers to the questions that have been posed in this thread. Cheers.

PR #581, opened 2015-07-13
Feature - TLS Evaluation through Server Trust Policy

cnoon:
This PR adds the ability to evaluate the validity of the TLS handshake in many different ways. It is not 100% complete, but is getting very close. I wanted to push it up to gather feedback from everyone before I finalize all the public docstrings and add documentation to the README (which will be extensive).

# Features

There are a whole bunch of new features added in this PR. I'll try to break down each of them one-by-one.

### Evaluation Policies per Host

Since different servers and web services have leaf certificates, intermediate and even root certificates, it is important to have the flexibility to specify evaluation policies on a per host basis. Otherwise, wouldn't be able to do things like use default evaluation for `host1`, certificate pinning for `host2` and public key pinning for `host3`.

The new `ServerTrustPolicy` enumeration and `ServerTrustPolicyManager` class help make this happen. You specify a policy for each host and Alamofire will handle the rest.

### Default Evaluation

The session delegate and task delegate did receive challenge callbacks both still use the `.PerformDefaultHandling` auth challenge disposition by default. This guarantees that anyone using Alamofire without a custom `ServerTrustPolicy` will get the Apple's default evaluation behavior. Therefore, you aren't getting any kind of custom behavior unless you actually specify you want it.

Here's a list of the custom types of things you can do with the new `ServerTrustPolicy` enumeration:

* The `.PerformDefaultEvaluation` case allows you to control whether you want to use host validation for the challenge. This could be helpful if you have a certificate with a DNS Name that doesn't match the host. It is certainly recommended though to ALWAYS validate the host.

  > We recommend that you always validate the host.

* You can now disable evaluation for a given host using the `.DisableEvaluation` case. This can be very helpful for self-signed certificates in development environments.

  > We would strongly advise against ever using this in a production environment.

* The `.CustomEvaluation` case allows you to provide an evaluation closure for your host to determine whether the challenge should be authenticated or not. This is syntactically much easier and safer to do than using the session and/or task delegate override closures. This also allows you to get all the default behavior for any other auth challenge type such as basic auth.

## Certificate Pinning

Certificate pinning is a useful way to guarantee that a certificate chain is not only valid, but also includes a certificate that you have provided. This is very helpful in mitigating a MITM attack. Here are some additional notes:

* You can provide multiple certificates for a given host. Each certificate is used as an anchor certificate in the evaluation of the server certificate chain.
* You can control the host validation as well when pinning certs.

## Public Key Pinning

Public key pinning has some additional advantages over certificate pinning. Public keys can be reused between different certificates after they expire. This makes it much easier to deal with revocation dates. The `ServerTrustPolicy` supports public key pinning and has some pretty powerful customizable options.

* Pin one or multiple public keys to a given host. This allows you to specify backup keys that can be used just in case your current keyset gets compromised.
* You can control whether host validation is ON or OFF.
* You can also specify whether to allow invalid certificates or not. This can be very useful in certain situations such as you have an expired leaf certificate or an intermediate certificate is missing.
> These issues should be resolved on Production environments. We recommend using public key pinning in combination with host validation and full certificate chain evaluation.

#### Where's my SPKI (Subject Public Key Info) Pinning?

We are working very hard to make this happen, but it will not be part of this original feature set. Apple does not expose the SPKI through the Security framework which means we can't extract that chunk of data directly from the certificate data. In order to implement this, we're considering rolling our own DER decoder to do this on-the-fly. AFNetworking has already made some good progress on this front. Regardless, we won't add this feature unless we're 100% confident in our implementation. Please stay tuned. The TC is all over this.

---

## Tests Tests and More Tests

So I wrote more tests than I can possibly imagine. To give you a rough idea of how many tests I wrote, all the logic to support the `ServerTrustPolicy` is about 250 lines and the tests are ~1600 lines long. I'm sure there are still many more tests that I could write as well. Check out the `ServerTrustPolicyTests` and `TLSEvaluationTests` for more info. 

> Seriously...check it out. We need to make sure we have every possible case accounted for.

## Sidenotes

* The `NSURLSession` extension ended up being a MUCH cleaner approach than I originally had. It's fairly tricky to make the server trust policies available to the session and task delegate callbacks without any of them knowing about the `Manager`. It's great design, but presented some issues when trying to gracefully share around the server trust policies with everyone that needed them. The extension ended up being a nice way to pass it around since all the callbacks pass the `NSURLSession` in as an argument. Win win.

## TODO Before Merging

* [X] Gather feedback from the community
* [X] Finalize the public docstrings
* [X] Add section to the README covering `ServerTrustPolicy` usage

---

# Questions For You

So with this large of a change, I'm looking for the community to help give me some feedback. Security is VERY difficult to get right, and a ton of effort has gone into this original PR. However, that doesn't mean that it is all the way there. Please help me dig through this PR with a fine tooth comb so we can get all the issues ironed out before it gets merged. Here are a few questions I have for everyone.

* Are the `ServerTrustPolicy` cases clear? I have written over 10 variations of them and these were the ones that I felt were most clear.
* How about `ServerTrustPolicy`? I wanted to make sure it was specific to dealing with server trust auth challenges. I also wanted to remove SSL or TLS from the name to avoid confusion. I don't want to have to close out issues asking for us to add SSL pinning just b/c we named our APIs TLS something or other.
* What tests am I missing? I feel like I did a pretty solid job of covering all the possible cases for `ServerTrustPolicy`.
  * How about for TLS evaluation?
* Should certificate pinning have the option of `allowInvalidCertificates`? At the moment this is something that I didn't support, but I'm looking for feedback.

Thank you everyone as always for helping out!


cnoon:
cc @kylef, @kcharwood

vkedwardli:
allowInvalidCertificates but protected with pinning would be great for development purposes

cnoon:
I agree @vkedwardli. Thinking through it more it did seem to be a validate case. Support added in a1ea7318.

cnoon:
I just pushed up another set of commits that completed the following:

* Refactored `allowInvalidateCertificates` to `validateCertificateChain` to be much more inline with the `validateHost` naming convention.
* Added ability to set whether to `validateCertificateChain` for certificate pinning.
* Added a bunch of tests around the new behavior.
* Added public docstrings for all public APIs.
* Added a Security section to the README.

At this point I feel really good about this set of changes. This should be good to go pending a full review from the TC.

zgosalvez:
Well done @cnoon! Thanks. I've successfully used this on my own shared manager instance.

Instead of implementing our own shared instance of manager, is there a way to set this default for `Alamofire.Manager.sharedInstance`? Would be easier to use `Alamofire.get` etc... iOS9 is requiring HTTPS anyway.

cnoon:
Thank you for the feedback @zgosalvez. 

I definitely thought about the use case of using a custom `ServerTrustPolicy` with the `Alamofire.Manager.sharedInstance`. The reason I chose not to support it is that you are always fighting a potential race condition. By opening up the ability to modify the `serverTrustPolicyManager` instance on the `session`, people could change this out while `Request`s are running which causing various threading issues.

I also think that if someone is going to go to the effort of setting up a custom `ServerTrustPolicy`, they are probably not using the `Alamofire.Manager.sharedInstance` object. They are most likely configuring a custom `Manager` instance anyways to change the default behavior such as using a custom `NSURLSessionConfiguration`.

I'll continue to think on this some more, but I'm definitely in the camp of not supporting this for now.

crewshin:
Anyone have a code snippet that demonstrates how to use this?

cnoon:
Hi @crewshin,

Directly from the README updates in this PR:

### Security

Using a secure HTTPS connection when communicating with servers and web services is an important step in securing sensitive data. By default, Alamofire will evaluate the certificate chain provided by the server using Apple's built in validation provided by the Security framework. While this guarantees the certificate chain is valid, it does not prevent man-in-the-middle (MITM) attacks or other potential vulnerabilities. In order to mitigate MITM attacks, applications dealing with sensitive customer data or financial information should use certificate or public key pinning provided by the `ServerTrustPolicy`.

#### ServerTrustPolicy

The `ServerTrustPolicy` enumeration evaluates the server trust generally provided by an `NSURLAuthenticationChallenge` when connecting to a server over a secure HTTPS connection.

```swift
let serverTrustPolicy = ServerTrustPolicy.PinCertificates(
    certificates: ServerTrustPolicy.certificatesInBundle(),
    validateCertificateChain: true,
    validateHost: true
)
```

There are many different cases of server trust evaluation giving you complete control over the validation process:

* `PerformDefaultEvaluation`: Uses the default server trust evaluation while allowing you to control whether to validate the host provided by the challenge. 
* `PinCertificates`: Uses the pinned certificates to validate the server trust. The server trust is considered valid if one of the pinned certificates match one of the server certificates.
* `PinPublicKeys`: Uses the pinned public keys to validate the server trust. The server trust is considered valid if one of the pinned public keys match one of the server certificate public keys.
* `DisableEvaluation`: Disables all evaluation which in turn will always consider any server trust as valid.
* `CustomEvaluation`: Uses the associated closure to evaluate the validity of the server trust thus giving you complete control over the validation process. Use with caution.

#### Server Trust Policy Manager

The `ServerTrustPolicyManager` is responsible for storing an internal mapping of server trust policies to a particular host. This allows Alamofire to evaluate each host against a different server trust policy. 

```swift
let serverTrustPolicies: [String: ServerTrustPolicy] = [
    "test.example.com": .PinCertificates(
        certificates: ServerTrustPolicy.certificatesInBundle(),
        validateCertificateChain: true,
        validateHost: true
    ),
    "insecure.expired-apis.com": .DisableEvaluation
]

let manager = Manager(
    configuration: NSURLSessionConfiguration.defaultSessionConfiguration(),
    serverTrustPolicyManager: ServerTrustPolicyManager(policies: serverTrustPolicies)
)
```

These server trust policies will result in the following behavior:

* `test.example.com` will always use certificate pinning with certificate chain and host validation enabled thus requiring the following criteria to be met to allow the TLS handshake to succeed:
  * Certificate chain MUST be valid.
  * Certificate chain MUST include one of the pinned certificates.
  * Challenge host MUST match the host in the certificate chain's leaf certificate.
* `insecure.expired-apis.com` will never evaluate the certificate chain and will always allow the TLS handshake to succeed.
* All other hosts will use the default evaluation provided by Apple.

vkedwardli:
@zgosalvez or anyone who loves racing, just `public` the `NSURLSession` extension inside ServerTrustPolicy.swift so you can access `sharedInstance.session.ServerTrustPolicyManager`

cnoon:
Thanks for the feedback @vkedwardli. While you are certainly correct, I would strongly advise against making such a change. Anyone that needs to customize the behavior of Alamofire should be creating their own instance of a `Manager`. The `sharedInstance` is just the common case scenario. This gives you the freedom to modify the session configuration headers as well as create a ephemeral or more commonly, a background session. It also avoids the race condition of setting the `ServerTrustPolicy` after you have already started requests.

briv:
I only took a look at this now (better late than never I hope), and I've got to say, great job @cnoon ! I really like your `NSURLSession` extension integration and how you added the per-host settings with the umbrella `ServerTrustPolicyManager`.
The one thing I would like to lobby for is a small addition to the Alamofire README and/or in-code documentation: that setting `validateCertificateChain` to `false` still leads to a byte equality check on certificates/keys. Or perhaps some example use-cases (is using an invalid certificate while in dev as @vkedwardli mentioned the primary one ?) to set the minds of the jittery like me at ease =)
This is actually what led to starting #356. When looking into AFNetworking a while back, the `validatesCertificateChain` property sounded a bit dangerous, and I didn't feel comfortable setting it to `false` without looking at the source code. Alamofire didn't have certificate pinning at that point and so it seemed like a good way to try and help out. It makes sense knowing the domain vocabulary of TLS and X.509 certificates but some extra hand-holding might be nice for a security feature ?

cnoon:
Thanks for the feedback @briv! I couldn't agree with your comments about the documentation. I just pushed 0d8803e6 to hopefully address your concerns. Thanks again!

FaridHaddad:
Hello, 

I am using Alamofire this way:

Alamofire.request(.POST, "https://www.test.com/test.php", parameters: ["cardNumber": cardNumber, "cardCode": cardCode]).responseJSON { response in
                    print(response.request)  // original URL request
                    print(response.response) // URL response
                    print(response.data)     // server data
                    print(response.result)   // result of response serialization
                    
                }

Any idea on how to implement SSL Pinning in it ? 
I just can't understand how .. 

Thank you.

dkavraal:
@FaridHaddad https://github.com/Alamofire/Alamofire/pull/581#issuecomment-215528326 and for later investigators, do this:


```swift
let serverTrustPolicies: [String: ServerTrustPolicy] = [
    "test.example.com": .pinCertificates(
        certificates: ServerTrustPolicy.certificates(),
        validateCertificateChain: true,
        validateHost: true
    ),
    "insecure.expired-apis.com": .disableEvaluation
]

let sessionManager = SessionManager(
    serverTrustPolicyManager: ServerTrustPolicyManager(policies: serverTrustPolicies)
)

sessionManager.request(.POST, "https://www.test.com/test.php",     //...rest same
```

Don't forget to include the certificates in your project.

mataspetrikas:
@dkavraal sorry for a dumb question, but how do I include the certificate correctly so AlamoFire can find it? Is it enough to just drag into the Resources group in XCode?

And is there a way to debug the validation errors like the
`Error Domain=NSURLErrorDomain Code=-999 "cancelled" `?

gunhansancar:
@mataspetrikas I just had to deal with the same question myself.

So the solution that I found by trial and error is that I needed to add DER encoded .cer file in my main bundle of the app. Just drag and drop to the project and copy files.

To get the DER encoded .cer file, I used Firefox and opened the url and by using the address bar You can go more details of the certificate then Security(Tab) -> View Certificate -> Details(Tab)->Export...
Then select X.509 Certificate (DER)

Hope it helps.

dkavraal:
@mataspetrikas on the contrary it's a good question. As @gunhansancar mentions, make sure you have der certificate. Then, apart from where you put the file, (move the file with your mouse into) copy the file into "**copy bundle resources**" inside Project settings -> Build phases. Same as in here: http://stackoverflow.com/questions/3095612/warning-the-copy-bundle-resources-build-phase-contains-this-targets-info-plist

dvertovsek:
Certificate pinning stopped working with Swift 3.0
I am using the following code:

```
let myServer = "...". //string in format without https://

let pathToCert = Bundle.main.path(forResource: "certificate", ofType: "der")
let localCertificate = NSData(contentsOfFile: pathToCert!)!

 let serverTrustPolicy = ServerTrustPolicy.pinCertificates(
            certificates: [SecCertificateCreateWithData(nil, localCertificate)!],
            validateCertificateChain: true,
            validateHost: true
        )

let serverTrustPolicies = [
            myServer: serverTrustPolicy
        ]

afManager = SessionManager(
            serverTrustPolicyManager: ServerTrustPolicyManager(policies: serverTrustPolicies)
        )

afManager.request("... some other domain ...", method: .get).response { response in
            log.info(response)
        }
```
I still get response (status code 200) from the other domain
What could be wrong here?

dvertovsek:
Could something be wrong with my certificate and/or the way I input my domain?
This is what my certificate looks like
screen shot 2017-06-29 at 11 48 48

```
let serverTrustPolicies = [
            "*.mydomain.com": serverTrustPolicy
        ]
```

PR #590, opened 2015-07-16
Feature - Opening Up Request

cnoon:
There are a couple of changes here. 

## Removal of respondsToSelector

The first was to remove the `respondsToSelector` override in the `Manager` class. The only reason I could come up with as to why they were there in the first place was that the original `SessionDelegate` override closures were internal. Having this override would allow clients to "peak" into whether or not those internal closures were set. Since the override closures are now public themselves, this logic no longer seems necessary.

> If anyone can think of a good reason as to why these shouldn't be removed...please comment on this PR!

## Opening the TaskDelegate Queue

I originally set out to open up the `TaskDelegate` APIs completely. After a deep investigation of the way the task delegate override closures are implemented and abstracted, I think it's best to leave them as they are. They are extremely well designed and provide an amazing amount of abstraction that can be overridden by the `SessionDelegate` closures. If a user really needs to customize the default behaviors, the `SessionDelegate` override closures should be used instead. Opening the `TaskDelegate` override closures would not only be redundant, but I'm afraid it would open up an abundance of confusion to the general user.

With that said, there is still an opportunity to allow users to implement custom behaviors on the `TaskDelegate` queue. At the moment, you cannot customize any `Request` operation logic until one of the `response` serializers are called. This is problematic if you need to run any other operations prior to this. For example, if you wanted to run some pre and post operations around validation, you couldn't do so.

Because of this, I opened up the `TaskDelegate` and `queue` to allow for customization. I also added some unit tests around this behavior to ensure it works as expected. This should give users ultimate flexibility without exposing hardly any of the internal API. Win win win. I'm excited to see what users will end up doing with this new found freedom! 😁


cnoon:
cc @kylef and @kcharwood

avf:
The removal of respondsToSelector seems to have broken NTLM (and maybe other authentification methods?) support. I've opened a separate issue #721 for this.

Issue #296, opened 2015-01-11
Serializer should be able to create non-object from response


tkrajacic:
It would make sense to be able to create a serializer that can for example return a `struct` from a response.
Since _Alamofire_ uses `AnyObject?` for Serializers this seems not possible.
Is there any particular reason, why this has to be an object type?

~~Switching out `AnyObject?` with `Any?` in the right places crashes the compiler with a segmentation fault though.~~ This is produced by the JSON and PLIST Serializers when using `Any`

Not sure why the cast from AnyObject? to Any? fails here:
`let JSON: Any? = NSJSONSerialization.JSONObjectWithData(data!, options: options, error: &serializationError)`

tkrajacic:
Sorry, didn't mean to ask for support. I was wondering if there is a specific reason why Alamofire can return only reference types from response serializers and can't serialize the response to an enum for example. (Would be great for e.g. a Swift XML solution)
This would make it more of a feature request I guess. I hit a snag, when the compiler crashed while trying to add this feature.

bilby91:
I'm interested in this too.

bilby91:
Something like this could work @mattt ?

https://github.com/bilby91/Alamofire/commit/2c88aeb83aa3846b2cf4d9da6ce626fb42e7aaf5

cnoon:
We're looking to add better support for this in the 1.3.0 release. Please stay posted 👍🏻

NicholasTD07:
+1 I hit this wall when I was trying to add a `Serializer` for a struct which conforms to a protocol `Decodable` from another framework `Argo`. So there's really nothing I can do to change the protocol.

@cnoon If I want to help with this, could you give me some guidance on how can it be done and what should be considered?

jqsilver:
I wrote a workaround using a Box class to wrap the structs. I took some shortcuts with the json part but you can get the general idea.  https://gist.github.com/jqsilver/82a515bfaca9a8552ad0

cnoon:
Hey everyone, thank you so much for your patience! I just pushed up PR #593 for Response Serializer Generics. If you could check it out and help with feedback I'd really appreciate it.

If you could please redirect all comments to #593 from here on out, that would be helpful. I'm going to go ahead and close this issue out.

PR #593, opened 2015-07-17
Feature - Response Serializer Generics

cnoon:
This PR addresses issue #296 by refactoring the response serialization logic to leverage the power of generics. With the addition of the `ResponseSerializer` protocol and the `GenericResponseSerializer` generic struct, it is now much easier to work directly with value types as well as get the actual type returned directly rather than having to cast the `AnyObject`.

## Improvements

* The four default response serialization methods now all return the most applicable type directly instead of an `AnyObject`. You no longer have to cast the result from an `AnyObject?` to `NSData`...Alamofire now handles this internally using generics.
> JSON and PropertyLists still return an `AnyObject` since that is the most applicable type in those cases.
* Custom response methods can now support generic types directly without having to cast the result of the underlying `response` method implementation.

## API Changes

The coolest part about these changes is that they have almost the exact same public API. As you can see from all the test logic, the high level functions that everyone is consuming remain unchanged. The only difference is if you are using the underlying `response` method. In this case, the `serializer` parameter has been renamed to `responseSerializer` for clarity purposes.

Since the underlying `response` implementation previously returned an `AnyObject?`, those implementations will still "just work". What it will result in is that you will no longer need to cast the serialized object being returned.

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -22,9 +22,100 @@
 
 import Foundation
 
-// MARK: String
+// MARK: - ResponseSerializer
+
+/**
+    The type in which all response serializers must conform to in order to serialize a response.
+*/
+public protocol ResponseSerializer {
+    /// The type of serialized object to be created by this `ResponseSerializer`.
+    typealias SerializedObject
+
+    /// A closure used by response handlers that takes a request, response, and data and returns a serialized object and any error that occured in the process.
+    var serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> (SerializedObject?, NSError?) { get }
+}
+
+/**
+    A generic `ResponseSerializer` used to serialize a request, response, and data into a serialized object.
+*/
+public struct GenericResponseSerializer<T>: ResponseSerializer {
+    /// The type of serialized object to be created by this `ResponseSerializer`.
+    public typealias SerializedObject = T
+
+    /// A closure used by response handlers that takes a request, response, and data and returns a serialized object and any error that occured in the process.
+    public var serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> (SerializedObject?, NSError?)
+}
+
+// MARK: - Default
 
 extension Request {
+
+    /**
+        Adds a handler to be called once the request has finished.
+
+        :param: queue The queue on which the completion handler is dispatched.
+        :param: responseSerializer The response serializer responsible for serializing the request, response, and data.
+        :param: completionHandler The code to be executed once the request has finished.
+
+        :returns: The request.
+    */
+    public func response<T: ResponseSerializer, V where T.SerializedObject == V>(
+        queue: dispatch_queue_t? = nil,
+        responseSerializer: T,
+        completionHandler: (NSURLRequest, NSHTTPURLResponse?, V?, NSError?) -> Void)
+        -> Self
+    {
+        self.delegate.queue.addOperationWithBlock {
+            let result: V?
+            let error: NSError?
+
+            (result, error) = responseSerializer.serializeResponse(self.request, self.response, self.delegate.data)
+
+            dispatch_async(queue ?? dispatch_get_main_queue()) {
+                completionHandler(self.request, self.response, result, self.delegate.error ?? error)
+            }
+        }
+
+        return self
+    }
+}
+
+// MARK: - Data
+
+extension Request {
+
+    /**
+        Creates a response serializer that returns the associated data as-is.
+
+        :returns: A data response serializer.
+    */
+    public static func dataResponseSerializer() -> GenericResponseSerializer<NSData> {
+        return GenericResponseSerializer { request, response, data in
+            return (data, nil)
+        }
+    }
+
+    /**
+        Adds a handler to be called once the request has finished.
+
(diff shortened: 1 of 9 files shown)


cnoon:
cc @kylef and @kcharwood

Source/ResponseSerialization.swift
```diff
@@ -22,9 +22,100 @@
...
+        Adds a handler to be called once the request has finished.
+
+        :param: completionHandler The code to be executed once the request has finished.
+
+        :returns: The request.
+    */
+    public func response(completionHandler: (NSURLRequest, NSHTTPURLResponse?, NSData?, NSError?) -> Void) -> Self {
+        return response(
+            responseSerializer: Request.dataResponseSerializer(),
+            completionHandler: { (request, response, data: NSData?, error) in
```
jqsilver: I'm curious, why can't you just pass `completionHandler` directly here?

NicholasTD07: +1

Source/ResponseSerialization.swift
```diff
@@ -32,8 +123,8 @@ extension Request {
 
         :returns: A string response serializer.
     */
-    public class func stringResponseSerializer(var encoding: NSStringEncoding? = nil) -> Serializer {
-        return { _, response, data in
+    public static func stringResponseSerializer(var encoding: NSStringEncoding? = nil) -> GenericResponseSerializer<String> {
```
NicholasTD07: Why cant it be a class method?

Source/ResponseSerialization.swift
```diff
@@ -22,9 +22,100 @@
...
+
+        :param: queue The queue on which the completion handler is dispatched.
+        :param: responseSerializer The response serializer responsible for serializing the request, response, and data.
+        :param: completionHandler The code to be executed once the request has finished.
+
+        :returns: The request.
+    */
+    public func response<T: ResponseSerializer, V where T.SerializedObject == V>(
+        queue: dispatch_queue_t? = nil,
+        responseSerializer: T,
```
NicholasTD07: Changing the parameter's name from `serializer` to `responseSerializer` will break the API, right? If so, the MAJOR version number needs to be bumped, right? But if this is the only reason to bump the Major version number, then it seems to be a very small change in code causing a very big change in the framework version number. Would it be an  overkill?

Does it make sense to keep the parameter's name as `serializer` right now? So we can keep the MAJOR version at 1 and we can change the name to `responseSerializer` later along with other API breaking changes.

cnoon: These are all great points @NicholasTD07 that I certainly considered. 

I did everything possible here to keep the public APIs as consistent as possible, while still fixing the root problem of always using an `AnyObject?` return type in the root `response` method. There are technically three changes that break semver.

* The response serializer implementations have been changed from `class` to `static` methods.
* The `serializer` parameter in the `response` method has been renamed to `responseSerializer` to provide more clarity.
* The type on `responseSerializer` is no longer a `Serializer` closure, but is instead a generic that conforms to the `ResponseSerializer` protocol.

Generally, I'd fall into the camp that this is an Alamofire 2.0 type of change. However, these changes above will affect only a small group of consumers. The majority of users are not ever going to implement the `response` method directly or ever subclass any of the `ResponseSerializer` methods. Additionally, the Alamofire 1.3.0 release is the last release that we're planning before Alamofire 2.0 work begins. The major downside to deferring this change to 2.0 is that this change wouldn't drop for months. The Alamofire 2.0 release is currently slated to go out roughly with the Swift 2.0 GM. This could still change if we decide to actually push a release on a beta version of Xcode, but I doubt that we'll go that route.

Therefore, to me, I feel that the slight deviation from semver here to serve the greater overall good is the right way to go. Renaming a variable for a very low API that gives me way more power and safety is all good in my book. But that's me, and I want to make sure that the community and the TC agrees. If the everyone would rather follow semver to the book here, I'm happy to defer this change until the 2.0 release.

Source/ResponseSerialization.swift
cnoon: Great catch guys! It was previously necessary and I didn't catch that it could be removed. Addressed in ed752d39.

Source/ResponseSerialization.swift
cnoon: So it certainly could be, but my thought was that it shouldn't be. Here are a couple questions that I think will help us find the right balance:

* Does it really make sense to subclass the `data`, `string`, `JSON` or `propertyList` response serializers? 
* Does it even make sense to subclass `Request` at all?

> IMO the answer to both is NO.

The recommended usage here is certainly to extend `Request` to provide your own custom response serializations. You would implement your own `responseSomeCustomName` method that could either call one of the four default response serializers, or you would also create a `CustomResponseSerializer<T>` to handle your use case. 

Even in the event that you did end up subclassing `Request`, would it make sense to really override any of the four default response serializers with custom behavior? Wouldn't you instead implement your own generic response serializer and call your own custom response method? This was my reasoning for going with `static` rather than `class`.

Thoughts?

NicholasTD07:
@cnoon I agree with everything you said. I also think it is ok to break the rules(semver) once a while. Just need to make sure to document it well. So that when people update their dependencies(which include Alamofire) no matter with or without the knowledge that there's a breaking change, they can find their answer quick.

Great job on the change itself and the comment about it. Thank you for being patient explaining everything. :)

Have a good day.

PR #595, opened 2015-07-18
Feature - Download Resume Data

cnoon:
This PR adds much better support for resume data with regards to download requests to address #141.

## Problem

Previously, there were only two ways to get resume data out of a request.

1. Cancel the request using the `Request.cancel()` method and retrieve the resume data from the `AnyObject?` in the `response` method.
1. Extract the resume data from the `error` returned in any of the `response` methods from the `userInfo` dictionary using the `NSURLSessionDownloadTaskResumeData` key.

The downsides to this is that you could only get direct access to the resume data by canceling the request. It would never be directly returned in the `response` method if it was interrupted by say a loss of network connection. Additionally, you could only ever access the resume data in the `response` parameters through either the `data` parameter in the event of a cancelled request, or through the `error` by digging the data out from the `userInfo` dictionary. This made it impossible to fetch the resume data without the use of the `response` serializers.

## Solution

To improve the resume data support, I made a couple of key changes that should make it MUCH easier to work with resume data.

1. For any failed download request, the resume data is automatically extracted out of the `error` object if it is available.
1. Added a `resumeData` computed property on the `Request` to allow you to access that data outside the `response` serialization methods. For example, you can now access the `resumeData` in any of the chained `Request` methods if you so choose.

  > This change coupled with the changes in #590 should give you complete customization control of the behavior of download requests and how the serial queue executes after completion.

I also added several tests around this behavior to demonstrate the usage and verify the behavior.

### DIFF
--- a/Source/Download.swift
+++ b/Source/Download.swift
@@ -134,6 +134,17 @@ extension Request {
         }
     }
 
+    /// The resume data of the underlying download task if available after a failure.
+    public var resumeData: NSData? {
+        var data: NSData?
+
+        if let delegate = self.delegate as? DownloadTaskDelegate {
+            data = delegate.resumeData
+        }
+
+        return data
+    }
+
     // MARK: - DownloadTaskDelegate
 
     class DownloadTaskDelegate: TaskDelegate, NSURLSessionDownloadDelegate {
--- a/Source/Download.swift
+++ b/Source/Download.swift
@@ -141,7 +141,7 @@ extension Request {
         var downloadProgress: ((Int64, Int64, Int64) -> Void)?
 
         var resumeData: NSData?
-        override var data: NSData? { return resumeData }
+        override var data: NSData? { return self.resumeData }
 
         // MARK: - NSURLSessionDownloadDelegate
 
--- a/Tests/DownloadTests.swift
+++ b/Tests/DownloadTests.swift
@@ -253,6 +253,8 @@ class DownloadResumeDataTestCase: BaseTestCase {
         XCTAssertNil(response, "response should be nil")
         XCTAssertNil(data, "data should be nil")
         XCTAssertNotNil(error, "error should not be nil")
+
+        XCTAssertNil(download.resumeData, "resume data should be nil")
     }
 
     func testThatCancelledDownloadResponseDataMatchesResumeData() {
--- a/Tests/DownloadTests.swift
+++ b/Tests/DownloadTests.swift
@@ -66,12 +66,8 @@ class DownloadInitializationTestCase: BaseTestCase {
 // MARK: -
 
 class DownloadResponseTestCase: BaseTestCase {
-    // MARK: - Properties
-
     let searchPathDirectory: NSSearchPathDirectory = .CachesDirectory
     let searchPathDomain: NSSearchPathDomainMask = .UserDomainMask
-
-    // MARK: - Tests
 
     func testDownloadRequest() {
         // Given
@@ -181,7 +177,7 @@ class DownloadResponseTestCase: BaseTestCase {
 
         // Then
         XCTAssertNotNil(responseRequest, "response request should not be nil")
-        XCTAssertNotNil(responseResponse, "response response should not be nil")
+        XCTAssertNotNil(responseResponse, "response should not be nil")
         XCTAssertNil(responseData, "response data should be nil")
         XCTAssertNil(responseError, "response error should be nil")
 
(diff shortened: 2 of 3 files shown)


cnoon:
cc @kylef and @kcharwood

Source/Download.swift
```diff
@@ -141,7 +152,7 @@ extension Request {
         var downloadProgress: ((Int64, Int64, Int64) -> Void)?
 
         var resumeData: NSData?
-        override var data: NSData? { return resumeData }
+        override var data: NSData? { return self.resumeData }
```
kylef: I don't think this change is necessary.

Tests/DownloadTests.swift
```diff
@@ -216,3 +212,123 @@ class DownloadResponseTestCase: BaseTestCase {
...
+            }
+
+        download.cancel()
+
+        waitForExpectationsWithTimeout(self.defaultTimeout, handler: nil)
+
+        // Then
+        XCTAssertNotNil(request, "request should not be nil")
+        XCTAssertNil(response, "response should be nil")
+        XCTAssertNil(data, "data should be nil")
```
kylef: Should this not also assert that resumeData is nil?

cnoon: Yes it ABSOLUTELY should. Will fix.

Source/Download.swift
cnoon: It's certainly not necessary, but brings it inline with all other properties used throughout Alamofire. 

Previously, there were some instances where we were using `self` on properties, and others where we were not. IMO, it is MUCH easier to not make mistakes if you specify `self` on properties. Otherwise naming collisions can actually cause some un-anticipated problems. Especially when your function parameter names overlap with your property names which is a pretty common use case.

Therefore, a few commits ago, I audited the codebase for properties without `self` and added it to have the same convention throughout Alamofire. When working through this PR, I realized I missed one and made the change. While I agree with you that it's not necessary, I would certainly vote to keep the convention.

Thoughts?

Tests/DownloadTests.swift
cnoon: Addressed in c129015c.

mahadshahib:
how should i use this? this is my main function how can i resume after connection lost ? `func sStartDownload () {
        
        let fileUrl = self.getSaveFileUrl(fileName: Data[0] as String)
        let destination: DownloadRequest.DownloadFileDestination = { _, _ in
            return (fileUrl, [.removePreviousFile, .createIntermediateDirectories])
        }
        
        downloadRequest = Alamofire.download(Data[0] as String , to:destination)
        
            .downloadProgress { (progress) in
                
                self.progressCircle.progress = progress.fractionCompleted
               
            .validate()
            
            .responseData { defaultDownloadResponse in
                print(defaultDownloadResponse.resumeData as Any)
                self.Timeer?.invalidate()
                
                if  (defaultDownloadResponse.error != nil) || defaultDownloadResponse.resumeData != nil  {
                    self.startButton.isHidden = false
                    self.pauseButton.isHidden = true
                    print("someThing went wrong \(String(describing: defaultDownloadResponse.error))")
                    
                    print(defaultDownloadResponse.resumeData as Any)
                    
                    if let resumeDataa = defaultDownloadResponse.resumeData {
                        let Resume = CachableData(urll: self.Data[0].lastPathComponent as String, Resume: resumeDataa)
                        self.cacher.persist(item: Resume) { url, error in
                            if let error = error {
                                print("Text failed to persist: \(error)")
                            } else {
                                print("Text persisted in \(url?.absoluteString ?? "fuck")")
                                return
                            }
                        }
                        
                    }
                }
                
                else {
                    if self.Data.count >= 2 {
                        self.Data.removeFirst()
                        self.startButton.isHidden = false
                        self.pauseButton.isHidden = true
                        
                    }
                    else if self.Data.isEmpty == true {
                        self.startButton.isHidden = true
                        self.pauseButton.isHidden = true
                        return
                    }
                    else {
                        self.sendNotif(state: true)
                        self.Data.removeLast()
                        self.startButton.isHidden = true
                        self.pauseButton.isHidden = true
                        
                    }
                    if !self.Data.isEmpty && self.nextState == true  {
                        self.BeingDownload()
                    }
                    
                    self.progressCircle.progress = 0
                    self.progressLabel.text = "0%"
                    self.stateLabel.text = "Downloaded"
                    self.downloadTableView.reloadSections([0], with: UITableViewRowAnimation.fade)
                    self.SaveData()
                    
                }
        }
    }`

Issue #591, opened 2015-07-16
Support for plain text and json parts in a multipart-form upload


chrisco314:
### Support for plain text and json parts in a multipart-form upload

**TL;DR:**

I needed support for text / json parts in a multipart-form upload, added a single function to do so, and forked Alamofire here: https://github.com/chrisco314/Alamofire/tree/multipart_data_form_support_for_additional_data_types to do so.

**Context:**

I posted a question to SO here: http://stackoverflow.com/questions/31442411/alamofire-support-for-uploading-multipart-forms-with-text-and-file-parts in my efforts to figure out how to upload textual data in tandem with binary data, but got no response.

**Implementation:**

In the end, to achieve the compact form of:

    form.appendBodyPart(Multipart.plain("patient", text:"\(last.id!)")!)

I needed to add a single override to MultipartFormData.swift:

    /**
    Creates a body part from the data and appends it to the multipart form data object.

    The body part data will be encoded using the following format:

    - `Content-Disposition: form-data; name=#{name}` (HTTP Header)
    - Encoded file data
    - Multipart form boundary

    :param: headers The headers to encode into the `multipart-form` part.
    :param: data The data to encode into the `multipart-form` part.
    */
    public func appendBodyPart(#headers:[String: String], data: NSData) {
        
        let bodyStream = NSInputStream(data: data)
        let bodyContentLength = UInt64(data.length)
        let bodyPart = BodyPart(headers: headers, bodyStream: bodyStream, bodyContentLength: bodyContentLength)
        self.bodyParts.append(bodyPart)
    }

This allowed me to add support for text and json data payloads through the code pasted below. The approach I took was to build up abstracted multipart-data parts through the use of typesafe enums and method chaining:

        class MultipartPlain : MultipartProtocol {
            
            init?(name:String, text:String) {
                
                // initialize header
                self.headers = MultipartHeader()
                    .contentType(.Plain)
                    .contentDisposition(.Name(name))
                
                // serialize body
                self.body = text.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)
                if self.body == nil {
                    return nil
                }
            }
            
            /// The headers used to generate a `text/plain` part in a `multipart/form-data` part
            let headers:MultipartHeader
            
            /// The body used to generate a `text/plain` part in a `multipart/form-data` part
            var body:NSData!
        }

The extension code in its entirety is:

    import Alamofire

    /**
    Extension to the Alamofire MultipartFormData type to support direct
[...]

Issue #555, opened 2015-06-29
Multipart upload: file content not in the payload


top2tech:
In branch feature/multipart_form_data, try to upload some string and a file. The string is in the payload, but the file is not. 

### File url: 
assets-library://asset/asset.JPG?id=E2741A73-D185-44B6-A2E6-2D55F69CD088&ext=JPG

### Captured Traffic:

Accept-Encoding: gzip, deflate
Content-Type: application/x-www-form-urlencoded
Content-Length: 340
Accept-Language: en-us
Accept: /
Connection: keep-alive
Connection: keep-alive
User-Agent: app01/1 CFNetwork/711.3.18 Darwin/14.3.0

--alamofire.boundary.2373462a015381c7
Content-Disposition: form-data; name="postdata"

{"Content":"Lorem ipsum dolor sit er elit lamet, consectetaur cillium adipisicing pecu, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation"}
--alamofire.boundary.2373462a015381c7--

### Code:

class func addPost(post:Post, fileUrls: [NSURL], callback:(Post) -> Void)
{
    let url = resRootUrl

    let jStr = Mapper().toJSONString(post, prettyPrint: false)
    let data = jStr?.dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)

    println(fileUrls[0])

    sManager.upload(.POST, url,
        multipartFormData:
        {
            multipartFormData in
            multipartFormData.appendBodyPart(data:data!, name:"postdata")
            multipartFormData.appendBodyPart(fileURL: fileUrls[0], name: "unicorn")

        },
        encodingCompletion:
           {
              encodingResult in
                switch encodingResult {
                  case .Success(let upload, _, _):
                     upload.responseJSON
                    {
                       request, response, JSON, error in
                       println(JSON)
                    }
                  case .Failure(let encodingError):
                       println(encodingError)
            }
        }
    )

}

cnoon:
I would guess that when you are attempting to append the `fileURL` that it is actually failing. What happens if you do the following?

```swift
    multipartFormData:
    {
        multipartFormData in
        multipartFormData.appendBodyPart(data:data!, name:"postdata")
        let error = multipartFormData.appendBodyPart(fileURL: fileUrls[0], name: "unicorn")
        println(error)
    },
```

top2tech:
[...]

PR #596, opened 2015-07-19
Feature - MultipartFormData Modifications

cnoon:
This PR addresses three issues:

1. When uploading `MultipartFormData` where appending body parts would fail, the encoding would succeed and the upload would take place without the failed body parts.
1. Added a method for appending body parts with data, name and mime type.

  > See #591 and #555 for more details.

1. Added a completely open `appendBodyPart(stream:length:headers:)` method that allows you to append any kind of data with any type of headers you wish.

## Problem

The upload methods for `MultipartFormData` are very convenient and make it very easy to construct the data and upload it. The problem though is that the file URL methods will return an error if something goes wrong when creating the `BodyPart`. This is a very easy case to forget to handle. Even worse, the encoding does not realize there was an error and will continue onwards.

For example, say you had the following logic:

```swift
Alamofire.upload(
    .POST,
    URLString: "http://httpbin.org/post",
    multipartFormData: { multipartFormData in
        multipartFormData.appendBodyPart(fileURL: unicornImageURL, name: "unicorn")
        multipartFormData.appendBodyPart(fileURL: rainbowImageURL, name: "rainbow")
    },
    encodingCompletion: { encodingResult in
        switch encodingResult {
        case .Success(let upload, _, _):
            upload.responseJSON { request, response, JSON, error in
                println(JSON)
            }
        case .Failure(let encodingError):
            println(encodingError)
        }
    }
)
```

If `rainbowImageURL` is not a valid URL, appending the body part will actually return an error, but the encoding succeeds and only the unicorn image is uploaded to the server. This is very problematic and needs a better solution.

### Solution

Since it is clunky to have to handle the error each time you append a body part, I chose to handle the error internally and bubble it up when encoding takes place. What this means is that in the above example, you don't need to check if appending each body part succeeds or not, you just need to handle whether encoding succeeds which the `upload` methods already do for you automatically.

Another advantage is that encoding will fail immediately if there was an issue appending any of the body parts. This approach seems like the best way to solve the problem. Even when moving to Swift 2.0, we won't want to have to `try` each time we try to append a body part. It's much easier from the user's perspective to bubble up the error at the time of encoding.

We could also use both approaches and still return an error optional for the file URL methods, but that seemed redundant and not optimal.


cnoon:
cc @kylef and @kcharwood

Samback:
Is there any way to use it at swift 2.0 now?

cnoon:
@Samback there is not. We'll port all the 1.3.0 changes over to the `swift-2.0` branch once we cut the 1.3.0 release.

Samback:
@cnoon Oh, I see. Is there any way how I can to solve this problem for now ?

cnoon:
You'll have to roll your own @Samback for the next week or so if you absolutely have to have it now, or you'll need to use AFNetworking. There are some posts on Stack Overflow explaining how to put together a quick and dirty approach to generating the boundaries manually. Otherwise, I'd recommend you just hang on a bit longer.

Samback:
It's ok for me wait to the next week. 
Thanks !

PR #597, opened 2015-07-19
Brings 2.0 Branch Up To Speed

forgot:
Pulled in all recent changes from Master, resolved any conflicts I encountered, and it now builds without any issues. All unit tests currently included pass as well.

I've only been actively using (not just reading about) Swift for the last two weeks, so please forgive any newbish mistakes, and I would greatly appreciate any comments, suggestions, etc.


cnoon:
Thank you very much for all your hard work here @forgot!

I'm very sorry but I need to decline this PR. I don't want anyone else picking up all these changes without them having been made by our TC. I want to go through each of the new features that have been added for the 1.3.0 release one-by-one to make sure they get ported properly.

With that said, I ensure you that this work is forthcoming. We're very close to wrapping the 1.3.0 release and this will be the very next thing we pick up.

Thank you for your patience...and apologies for declining. I don't like to flat out decline PRs...but I think it's the best course of action for Alamofire and the community for time being.

Cheers 🍻

cnoon:
In the future, before you take on such a large chunk of work, please reach out to us through our [@AlamofireSF](https://twitter.com/AlamofireSF) twitter handle or open an issue asking for guidance from the TC as to whether to proceed. That would have given us the opportunity to let you know our thoughts before you went to all the work.

forgot:
No worries. At the very least, I learned quite a bit. I look forward to the updates from the team, and hopefully I can contribute in some way in the future.

Cheers 🍻

Issue #510, opened 2015-06-03
head :no_content (204 status code) returns error


allaire:
Looks like Alamofire isn't supporting 204 status code or is not able to parse the response. I tried `.response` and .`responseString` without any success.

Thanks

cnoon:
Hi @allaire, any chance you could provide some additional information (code sample or Github project) or even better, a set of tests that reproduce the issue and submit a PR?

allaire:
Hi @cnoon, thanks for the quick reply!

I'm currently using a Rails API, returning `head :no_content` on a DELETE endpoint, which returns the following error:

```
Optional(<NSHTTPURLResponse: 0x7fbd33666d60> { URL: http://192.168.0.196:5060/api/organizations/6295710d-03c9-427a-b1d2-8e12951c4238/requests/878166fb-1bdc-4962-9218-71b306c82da0 } { status code: 204, headers {
    "Cache-Control" = "no-cache";
    "Set-Cookie" = "request_method=DELETE; path=/, locale=ImZyLUNBIg%3D%3D--f4c51516b9f7125f95cb9ca87028fc996c54dc42; path=/; expires=Tue, 05 Jun 2035 13:34:38 -0000, session_updated_at=8597778371; path=/, XSRF-TOKEN=Ya6GOlesHxBTSTsPrntco8BazpeZhe0%2Fce0xjcU8M8B7K1ytZn%2BIzmWMHbkUy1lVQ%3D%3D; path=/, _my_session=RzMyN0s5Z2Z3S2xadVMrYUZTalcyMVVGZ084bFFMR2hvc1FibTBLU1U4bVd5YUN6Y0VkOWhqV1FSSGdNU01WRG40RGptM1pWWFdvb3RZRldJdXArQUpCSUtXMEkrcFNIZzVYak8weTlLTXpM%3D--98831ed1e111854eabd6b3189ab79a2e5f807bc1; path=/; HttpOnly";
    "X-Content-Type-Options" = nosniff;
    "X-Frame-Options" = SAMEORIGIN;
    "X-Meta-Request-Version" = "0.3.4";
    "X-Request-Id" = "c30e4b82-889e-4d78-b7bd-5546bed20ad3";
    "X-Runtime" = "0.259388";
    "X-XSS-Protection" = "1; mode=block";
} })
null
Optional(Error Domain=com.alamofire.error Code=-1 "The operation couldn’t be completed. (com.alamofire.error error -1.)")
```

Here my request code:

[...]

PR #598, opened 2015-07-19
Potential Fix for Wildcard Accept Type Failing Validation

cnoon:
This PR hopefully addresses Issue #510.

## Quick Problem Summary

The issue is described in detail [here](https://github.com/Alamofire/Alamofire/issues/510#issuecomment-112581617). The validation is failing if no `Content-Type` comes back in the response.

## Reproduction Attempts

I tried every http://httpbin.org trick in the book to recreate the issue they are seeing in #510, but was unsuccessful. I cannot trick the `NSHTTPURLResponse` into be so confused that it cannot guess what the `MIMEType` is. It always comes back valid. I would really like to have a test around this change, but just can't seem to get httpbin to do my bidding. If someone could provide a test, that would be amazing!

## Solution

The solution here seems pretty simple. A wildcard accept type `*/*` should always pass validation whether a `MIMEType` exists on the response or not. The current implementation will only pass if the MIMEType on the response is valid. Therefore, my change just adds a small condition that allows `Content-Type` validation to succeed if one of the acceptable content types was `*/*`. Additionally, I can't see any downsides of this change.

Since I can't get a test around this change, I'd love it if @allaire and/or @MartinRogalla could confirm that this change does in fact fix the issue they reported in #510.

### DIFF
--- a/Source/Validation.swift
+++ b/Source/Validation.swift
@@ -117,6 +117,12 @@ extension Request {
                         return true
                     }
                 }
+            } else {
+                for contentType in acceptableContentTypes {
+                    if let MIMEType = MIMEType(contentType) where MIMEType.type == "*" && MIMEType.subtype == "*" {
+                        return true
+                    }
+                }
             }
 
             return false


cnoon:
cc @kylef and @kcharwood

allaire:
@cnoon Will give it a try today or tomorrow, thanks for the PR!

cnoon:
Thank you @allaire...much appreciated!

allaire:
Hi @cnoon!

 First, great work on the detailed PR you always create, really good work :+1: 

Second, it works well with a Rails app that returns `head :no_content` . :100: 

As for the test, can't you try to stub the response, or at least try to modify it (removing the accept type) just before passing it to the validation?

Thanks!

cnoon:
Thank you for confirming @allaire! That makes me feel much better about this change. I'm going to go ahead and merge it.

As for your question about stubbing it, that's a pretty difficult thing to do that will require a custom `NSURLProtocol` which has all sorts of weird behaviors with `NSURLSession`s. Additionally, the validation gets registered as a closure into the task delegate queue which runs after the request completes, thus just making it difficult to simulate.

Once we have spent some time working through stubbing requests with Alamofire, it will make this a much easier thing to simulate. Until then, I'm just going to throw a ticket into our Trello project and we'll get to it later.

Thanks again for all the diligence...apologies for it taking a month for us to get it fixed.

Jirka1111:
@cnoon It is still not working. I don't know what I have supposed to use for this kind of request. The request is correct (checked by Postman), but Alamofire throws HTTP 404 and I am still screwed... May I help you with some kind of testing?

cnoon:
Hi @Jirka1111, if you are having a problem, please open a new issue and link back to this one if it is in fact related. I'm very confident in this change as well as being having it verified by @allaire who helped debug this issue for weeks.

allaire:
@Jirka1111 Make sure you use the latest release :)

Jirka1111:
@allaire Dear allaire, I certainly use Alamofire 1.3.0. The request is:

Alamofire.request(.PUT, "https://\(domainText).projektove.cz/issues/\(self.ids[indexPath.row]).done.json")
                .validate(contentType: ["*/*"])
                .responseJSON { (request, response, data, error) in ...


This is the output of response and error:

Optional(<NSHTTPURLResponse: 0x7ffd8e111250> { URL: https://mnd.projektove.cz/issues/151.done.json } { status code: 404, headers {
    Connection = "keep-alive";
    "Content-Length" = 485;
    "Content-Type" = "text/html; charset=utf-8";
    Date = "Wed, 29 Jul 2015 20:58:18 GMT";
    Server = "nginx/1.8.0";
    "X-Rack-Cache" = "invalidate, pass";
    "X-Request-Id" = 704421789695500c680a8131d085780f;
    "X-Runtime" = "0.005469";
} })

Optional(Error Domain=NSCocoaErrorDomain Code=3840 "The operation couldn’t be completed. (Cocoa error 3840.)" (Invalid value around character 0.) UserInfo=0x7ffd8bc053d0 {NSDebugDescription=Invalid value around character 0.})

I am little confused. If I use Postman, everything is fine and I get HTTP 204. But Alamofire gives me 404, which is the same as when I copy and paste the URL to the Chrome (The page don't exist).

cnoon:
@Jirka1111, please take your question to Stack Overflow. I'm almost certain you just have an issue with the way you're formatting your request. This [thread](http://stackoverflow.com/questions/25986129/cocoa-error-3840-invalid-value-around-character-0-afnetworking) may help shed some light.

Jirka1111:
@cnoon Thank you, I will try that solution. This can´t be done by Alamofire?

Jirka1111:
@cnoon Yeah, so now I am not any smarter. If Alamofire can processed this kind of action, I presumed there should be answer. Not just "proof of concept". I was blocked at Stackoverflow due to questions without answer so I can't asked another question. As I understand, my request should work. I don't want to use another library (or same library in different language). I am very satisfied with Alamofire, so I don't have a problem with some kind of a code "hack". HTTP 204 is common so I don't understand why there is not a solution (maybe in instructions?).

cnoon:
@Jirka1111 please be a bit more mindful of our time. We point you towards Stack Overflow when we know there's not an issue with the library but instead an issue with your code. It's purely to allow the community to help us with support questions. If we just sat and answered support questions, we'd never have time to work on moving the library forwards.

With that said, you are not receiving a 204 but instead of 404. You are most likely not hitting the server properly. You need to use cURL debugging or some other form to figure out where your issue lies. Please  do not continue to seek advice through this closed issue.

Best of luck.

Jirka1111:
@cnoon I am sorry, cnoon, I understand you. But you don't understand me. I am receiving 204 in Postman but 404 in Alamofire so I think it is not a problem with code. And that is why I still write to this thread. But ok, thank you a have a great time :-)

PR #609, opened 2015-07-27
Fix for non-converted Unmanaged<...> usages for Swift 2.0 branch

robskillington:


cnoon:
Hi @robskillington, as far as I can tell, these changes have already been made and are available in the `swift-2.0` branch. I thought I had already cleaned this all up. Looks like it all was changed in 78f155d8. Could you please confirm?

I'm not exactly sure why it's showing the diff b/c those changes have already been made in the branch your submitting the PR against. Seems strange to me.

robskillington:
Hey @cnoon it definitely wouldn't compile for me until I made these changes, using Version 7.0 beta 3 (7A152u)

cnoon:
Okay, your changes are completely valid and definitely need to occur. However, it appears to me that everything has already been changed. Check out the code [here](https://github.com/Alamofire/Alamofire/blob/swift-2.0/Source/ServerTrustPolicy.swift#L161-L219) that is the current form of the evaluate method in the `swift-2.0` branch. You'll see that the changes are already in there.

cnoon:
I feel like you're probably running into some out-of-date upstreaming issue.

ikesyo:
@robskillington That is because you are using beta 3, as opposed to the current `swift-2.0` branch is compatible with beta 4.

robskillington:
I see.

PR #616, opened 2015-07-27
Target for watchOS 2 added

tomco:


cnoon:
Nice...thanks for this @tomco. I have a few other things that I'm working on completing in Alamofire at the moment but I'll get to this as soon as possible!

jurvis:
@tomco think you should also check the "Shared" checkbox in the watchOS scheme, else Carthage will not be able to discover the scheme when running `carthage update`

tomco:
@jurvis thx missed that. Fixed now.

cnoon:
Thank you for all your hard work here @tomco! I squashed your PR, tweaked it slightly and pushed it into the `swift-2.0` branch in b4cb7cc3 giving you attribution. Great work!

Cheers 🍻

PR #619, opened 2015-07-27
Add note for changes to encoding in 1.3.0 to Change Log

AnthonyMDev:
Added a note to the backwards compatibility explanation for 1.3.0 for the changes to the return type for encoding. This is a change that affects backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,7 @@ make slight modifications to your global function calls.
 `ResponseSerializer` protocol. This has many advantages, the most obvious being that
 the `response` convenience method now returns an `NSData?` optional instead of an
 `AnyObject?` optional. Nice!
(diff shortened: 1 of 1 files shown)


cnoon:
This seems like an unnecessary addition to me because backwards compatibility is still maintained. For example, there is not any issue that I'm aware of with the following code block:

```swift
import Foundation

func URLRequest() -> NSMutableURLRequest {
    return NSMutableURLRequest(URL: NSURL(string: "")!)
}

let request1: NSURLRequest = URLRequest()
let request2: NSMutableURLRequest = URLRequest()
let request3 = URLRequest()
```

Am I missing something obvious?

AnthonyMDev:
That makes sense. It actually shouldn't change backwards compatibility, considering `NSMutableURLRequest` is a subclass of `NSURLRequest`.

When I was updating to use `Alamofire 1.3.0` yesterday, I was getting some odd build errors that looked related to this, but now I'm unable to replicate them. It may have been a derived data problem. Xcode does some odd things sometimes.

Issue #610, opened 2015-07-27
ServerTrustPolicy.swift crashes Swift compiler on cmd line


pcantrell:
On swift-2.0 branch, using XC7b4:

    $ xcodebuild -project Alamofire.xcodeproj/
    ...
    CompileSwift normal arm64 /Users/paul/work/opensource/Alamofire/Source/ServerTrustPolicy.swift
    ...
    Assertion failed: (isa<X>(Val) && "cast<Ty>() argument of incompatible type!"), function cast, file /Library/Caches/com.apple.xbs/Sources/swiftlang/swiftlang-700.0.47.4/src/llvm/include/llvm/Support/Casting.h, line 237.
    0  swift                    0x000000010cee3b8b llvm::sys::PrintStackTrace(__sFILE*) + 43
    1  swift                    0x000000010cee42cb SignalHandler(int) + 379
    2  libsystem_platform.dylib 0x00007fff8e11cf1a _sigtramp + 26
    3  libsystem_platform.dylib 0x00007fbb5aaebe30 _sigtramp + 3432836912
    4  libsystem_c.dylib        0x00007fff8f9fdb53 abort + 129
    5  libsystem_c.dylib        0x00007fff8f9c5c39 basename + 0
    6  swift                    0x000000010b3fca10 hoistOrCopySelf(swift::ApplyInst*, swift::SILInstruction*, swift::DominanceInfo*, bool) + 624
    7  swift                    0x000000010b3fc3ef swift::ArraySemanticsCall::hoistOrCopy(swift::SILInstruction*, swift::DominanceInfo*, bool) + 319
    8  swift                    0x000000010b24d4bb (anonymous namespace)::SwiftArrayOptPass::run() + 4283
    9  swift                    0x000000010b1c00ae swift::SILPassManager::runFunctionPasses(llvm::ArrayRef<swift::SILFunctionTransform*>) + 1598
    10 swift                    0x000000010b1c0d28 swift::SILPassManager::runOneIteration() + 984
    11 swift                    0x000000010b1bdcd4 swift::runSILOptimizationPasses(swift::SILModule&) + 500
    12 swift                    0x000000010aec9e08 performCompile(swift::CompilerInstance&, swift::CompilerInvocation&, llvm::ArrayRef<char const*>, int&) + 9992
    13 swift                    0x000000010aec74e3 frontend_main(llvm::ArrayRef<char const*>, char const*, void*) + 2627
    14 swift                    0x000000010aec35af main + 2015
[...]

PR #620, opened 2015-07-27
Fix for #610

pcantrell:
This is @robrasmussen’s fix turned into a PR.

Verified locally using:

    xcodebuild -project Alamofire.xcodeproj/ clean build


pcantrell:
Gyarr. Based against master instead of swift-2.0. Abort retry fail.

PR #632, opened 2015-07-31
Swift 2.0

adelmojunnior:
I was getting errors in the ServerTrustPolicy class with the unmanaged object then i fixed it.


cnoon:
We're not going to merge the `swift-2.0` branch into `master` until we're ready to release Alamofire 2.0 which will roughly coincide with the Xcode 7 GM release date.

PR #633, opened 2015-07-31
Correct compile swift 2

adelmojunnior:
I was getting errors in the ServerTrustPolicy class with the unmanaged object then i fixed it.
P.S. Sorry, I'm newbie


cnoon:
You just need to update to Xcode 7 beta 4 and all your issues will be resolved.

Issue #554, opened 2015-06-29
'qualityOfService' is only available on iOS 8.0 or newer


sugarwyc:
https://github.com/Alamofire/Alamofire/blob/swift-2.0/Source/Request.swift

```
if #available(OSX 10.10, *) {
    operationQueue.qualityOfService = NSQualityOfService.Utility
}
```
"'qualityOfService' is only available on iOS 8.0 or newer"

cnoon:
Thank you for the suggestion @sugarwyc. 

We will not be incorporating this change though since we are no longer supporting iOS 7 with Swift 2.0. Please see #525 for more information.

Cheers.

PR #635, opened 2015-08-02
allow embedding of source files to iOS 7 targeted apps

jankase:
Solve small issue when source code is embedded in Swift 2 app targeting iOS 7


cnoon:
Thank you for your PR @jankase. Unfortunately, we will not be merging this change. Please see #525 and #554 for more information. 

Cheers.

PR #637, opened 2015-08-02
Result bool type

jankase:
I think using BooleanType protocol is more Swift-like approach


jankase:
Sorry, accidentally created on wrong branch.

PR #638, opened 2015-08-02
Result enum conform to BooleanType protocol

jankase:
I think using BooleanType is more Swift-like approach.

### DIFF
--- a/Source/Result.swift
+++ b/Source/Result.swift
@@ -30,23 +30,18 @@ import Foundation
     - Failure: The request encountered an error resulting in a failure. The associated values are the original data 
                provided by the server as well as the error that caused the failure.
 */
-public enum Result<Value> {
+public enum Result<Value> : BooleanType {
     case Success(Value)
     case Failure(NSData?, NSError)
-
+    
     /// Returns `true` if the result is a success, `false` otherwise.
-    public var isSuccess: Bool {
+    public var boolValue : Bool {
         switch self {
         case .Success:
             return true
         case .Failure:
             return false
         }
-    }
-
-    /// Returns `true` if the result is a failure, `false` otherwise.
-    public var isFailure: Bool {
-        return !isSuccess
     }
 
     /// Returns the associated value if the result is a success, `nil` otherwise.


cnoon:
Thank you for your PR @jankase. I really hesitate to make this change because I feel it could be misleading to users to make a `Result<T>` conform to `BooleanType`. Here's a quick snippet that I put together to demonstrate why.

```swift
func testResultConditionals() {
    let result1 = Result.Success("Testing")
    let result2 = Result<String>.Failure(
        NSData(),
        NSError(domain: Error.Domain, code: -1, userInfo: nil)
    )

    // Switching on cases
    switch result1 {
    case .Success:
        print("Result 1 is a success")
    case .Failure:
        print("Result 1 is a failure")
    }

    // Pattern matching in conditionals
    if case .Success = result1 {
        print("Success comparison works")
    }

    if case .Failure = result2 {
        print("Failure comparison works")
    }

    // Convenience methods
    if result1.isSuccess {
        print("Success check works...clear and concise")
    }

    if result2.isFailure {
        print("Failure check works...clear and concise")
    }

    // Leveraging `BooleanType`
    if result1 {
        print("Result conforms to BooleanType so it can be used like this...could be confusing")
    }
}
```

Could you make a more substantial case as to why you think it should be added? At the moment I'm leaning towards not making this change.

cc @kcharwood and @kylef...do you guys have an opinion here?

kylef:
I think I agree with @cnoon on this.

Apple discourage conforming to BooleanType when you're representing more than a simple boolean value, take a look at the [BooleanType protocol documentation](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Reference/Swift_BooleanType_Protocol/).

> Only two types provided by Swift, Bool and ObjCBool, conform to BooleanType. Expanding this set to include types that represent more than simple boolean values is discouraged.

cnoon:
Ah yeah I didn't check the docs. That totally makes sense @kylef. Thank you for the PR @jankase, but I'm going to decline this PR. Thanks again!

jankase:
Thanks for @cnoon and @kylef comments. I'm Swift newbie and I'm glad for all comments helping me to implement better apps.

PR #639, opened 2015-08-02
fixed example project and using responseString with old syntax

jankase:
Only small fix to allow compilation of Example project


cnoon:
Thanks @jankase...much appreciated!

PR #650, opened 2015-08-06
Support Xcode 7 beta 5

mcudich:
Please let me know if this set of fixes looks right to you. Thanks in advance!

### DIFF
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -261,7 +261,7 @@ public class MultipartFormData {
             isReachable = fileURL.checkPromisedItemIsReachableAndReturnError(&error)
         } else {
             // Assume reachability on platforms where it's not possible to detect failure.
-            isReachable = true;
+            isReachable = true
         }
 
         guard isReachable else {
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -257,7 +257,7 @@ public class MultipartFormData {
         var isReachable = false
 
         if #available(OSX 10.10, *) {
-            var error: NSError?;
+            var error: NSError?
             isReachable = fileURL.checkPromisedItemIsReachableAndReturnError(&error)
         } else {
             // Assume reachability on platforms where it's not possible to detect failure.
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -256,16 +256,17 @@ public class MultipartFormData {
 
         var isReachable = false
 
-        do {
-            try fileURL.checkResourceIsReachable()
-            isReachable = true
-        } catch {
-            // No-op
+        if #available(OSX 10.10, *) {
+            var error: NSError?;
+            isReachable = fileURL.checkPromisedItemIsReachableAndReturnError(&error)
+        } else {
+            // Assume reachability on platforms where it's not possible to detect failure.
+            isReachable = true;
         }
 
         guard isReachable else {
-            let error = Error.errorWithCode(NSURLErrorBadURL, failureReason: "The file URL is not reachable: \(fileURL)")
-            setBodyPartError(error)
+            let reachabilityError = Error.errorWithCode(NSURLErrorBadURL, failureReason: "The file URL is not reachable: \(fileURL)")
+            setBodyPartError(reachabilityError)
             return
         }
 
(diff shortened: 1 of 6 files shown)


Source/MultipartFormData.swift
```diff
@@ -256,16 +256,17 @@ public class MultipartFormData {
 
         var isReachable = false
 
-        do {
-            try fileURL.checkResourceIsReachable()
-            isReachable = true
-        } catch {
-            // No-op
+        if #available(OSX 10.10, *) {
```
jshier: Need iOS in this check too?

Source/MultipartFormData.swift
```diff
@@ -256,16 +256,17 @@ public class MultipartFormData {
...
+            var error: NSError?;
+            isReachable = fileURL.checkPromisedItemIsReachableAndReturnError(&error)
+        } else {
+            // Assume reachability on platforms where it's not possible to detect failure.
+            isReachable = true;
```
jshier: Unnecessary semicolon. :laughing:

Source/MultipartFormData.swift
mcudich: I don't think so. The deployment target is set to 8.0, which is the OS version where this API was introduced.

Source/MultipartFormData.swift
mcudich: Removed - thanks :)

Source/MultipartFormData.swift
```diff
@@ -256,16 +256,17 @@ public class MultipartFormData {
 
         var isReachable = false
 
-        do {
-            try fileURL.checkResourceIsReachable()
+        if #available(OSX 10.10, *) {
```
jshier: Alamofire supports iOS 7, so this does need an iOS check.

Source/MultipartFormData.swift
```diff
@@ -256,16 +256,17 @@ public class MultipartFormData {
...
+            var error: NSError?;
```
jshier: Another unnecessary semicolon.

mcudich: Sorry - old habit!

Source/MultipartFormData.swift
mcudich: Are you sure? Looking at both the Xcode project and the podspec, iOS 8.0 is listed as the deployment target. Am I looking in the wrong place?

jshier: Hmm, the README says 10.9 and iOS 7, so it seems like it should. @cnoon ?

mcudich: The 10.9 statement is still valid. It's just that the iOS target seems to have moved up: https://github.com/Alamofire/Alamofire/blob/master/Alamofire.podspec

jshier: That's for the CocoaPod, since CocoaPods only supports iOS 8 for Swift pods.

mcudich: Ah - got it. I'll wait for @cnoon to weight in here.

cnoon: Just to clear this up, Alamofire 2.0 is not going to support iOS 7. The README in the `swift-2.0` branch reflects that decision.

cnoon:
Thank you for putting this together @mcudich...great work!

I merged a slightly modified version of your changes into a squashed commit in 62947d6d that still gives you attribution.

Thanks again! 🍻

Source/MultipartFormData.swift
jshier: @cnoon Is Alamofire 2.0 when official Swift 2 support will be landed? If official support for Swift 2 will land before then, it seems like supporting iOS 7 until the 2.0 release is necessary. So far, the changes to support iOS 7 in the `swift-2.0` branch have only been a single (other) `#available` check. While I'm not usually one to want (or need) much legacy support, I am in the middle of a project that has to target iOS 7 while using Swift 2 and it would be unfortunate to be stuck on an old version of Alamofire.

cnoon: The main reason we're dropping iOS 7 support in Alamofire 2.0 is that we can't test it. We don't want to, nor do we have the bandwidth to re-write our entire test suite to support iOS 7 since it doesn't support frameworks. Additionally,  Travis-CI does not support iOS 7 leaving us in a position where it just doesn't make sense to support it. This is just a decision that we've made for the greater good overall.

With that said, there's no reason you can't run a fork for yourself to make the changes. It's just not something we're officially going to support moving forwards.

Cheers.

PR #657, opened 2015-08-07
Fixes compile issues and some test problems. 

bwynne:
I made these changes on Xcode 7.0.4 and ran the iOS test suite.  Not knowledgeable enough to fix the cache-control func issue; no idea why response was nil:  added a TODO for someone with more expertise.


ikesyo:
You should update to Xcode 7 beta 5.

bwynne:
Ah. Ok.

PR #658, opened 2015-08-08
Swift 2-Style try/catch unwrapping for Result

maxgoedjen:


cnoon:
Thank you for your PR @maxgoedjen. 

In this case I think this is not a good addition for Alamofire. You are duplicating the `value: T?` computed property functionality and also introducing a `throws` that isn't necessary which just adds unnecessary complexity. A `Result` type should never have to `throw`. If you really want to keep this type of functionality, you can always use it as your own extension on the `Result` type.

Thanks again.

PR #659, opened 2015-08-08
Feature - Stream Task Support

cnoon:
This PR adds the ability to create `NSURLSessionStreamTask` objects directly with Alamofire as well as catch all the delegate callbacks.

## API Design

The strangest thing about this PR is the task override computed properties and then the duplicated private stored properties inside the base `SessionDelegate` implementation. Given the current limitations in Swift, I couldn't find any other way to do it. Here's a list of things I was fighting against:

* Stored properties don't support availability checks
* Cannot create stored properties in extensions
* Needed the `SessionDelegate` to not conform to the `NSURLSessionStreamDelegate` protocol on `watchOS`

Because of these limitations, I ended up implementing private stored properties as `Any` values and then implementing public computed properties in the extension. I also added an OS check to only compile the `Stream.swift` file on `iOS` and `OS X` as well as the availability checks for iOS 9.0+ and OSX 10.11+.

> If anyone has any ideas on how to clean this up a bit more, I'm all ears!


cnoon:
cc @kylef and @kcharwood

PR #660, opened 2015-08-08
Feature - Opening up ParameterEncoding

cnoon:
This PR changes the `queryComponents` and `escape` method ACLs from `internal` to `public`. This will make it easier for those implementing a `.Custom` parameter encoding. Right now if you need to use either function in a `.Custom` case, you have to implement both of these methods on your own.


cnoon:
cc @kylef and @kcharwood

PR #661, opened 2015-08-08
Feature - Custom Result Initializers

cnoon:
This PR adds custom initializers for the `Result` success and failure cases. It also adds tests around the usage.

### DIFF
--- a/Source/Result.swift
+++ b/Source/Result.swift
@@ -34,6 +34,29 @@ public enum Result<Value> {
     case Success(Value)
     case Failure(NSData?, NSError)
 
+    /**
+        Creates a `.Success` result with the given value.
+
+        - parameter value: The value to store in the `.Success`.
+
+        - returns: A new `.Success` result.
+    */
+    public init(value: Value) {
+        self = .Success(value)
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


cnoon:
cc @kylef and @kcharwood

Source/Result.swift
```diff
@@ -34,6 +34,29 @@ public enum Result<Value> {
...
     case Failure(NSData?, NSError)
 
+    /**
+        Creates a `.Success` result with the given value.
+
+        - parameter value: The value to store in the `.Success`.
+
+        - returns: A new `.Success` result.
+    */
+    public init(value: Value) {
```
kylef: I'm wondering, do they need to be public? Should anyone outside of Alamofire be creating a Result?

cnoon: That's really a good point. I was a bit torn as well. After adding the tests, I realized it's not any easier to use the initializers than the default case initialization. Maybe this whole PR should just be declined. What you think @kylef?

kylef: It's up to you, but I'd only merge this if we want to use it in Alamofire.

cnoon: Okay, I'm going to close it out. I don't think we'll end up using it.

PR #665, opened 2015-08-10
updated example

johndpope:


Example/DetailViewController.swift
```diff
@@ -72,9 +72,12 @@ class DetailViewController: UITableViewController {
             let end = CACurrentMediaTime()
             self.elapsedTime = end - start
 
-            for (field, value) in response!.allHeaderFields {
-                self.headers["\(field)"] = "\(value)"
+            if let _ = response{
```
kylef: This should be `if let response = response`, and that way you don't have to explicitly unwrap the optional below `response!`.

johndpope:
ok  - the commit was more about fixing the critical crash and getting demo working than reducing the exclamation marks. I've seen this pattern - if let response = response but I think there's a push to abbreviate it. thanks

cnoon:
Thank you for your PR @johndpope. I merged a modified form of your PR in 2bffaf1b giving you attribution. Thanks again!

johndpope:
fyi - this pull request added additional NSAppTransportSecurity to plist fixing the bug with example not working. The example is still not working. 

<key>NSAppTransportSecurity</key>
	<dict>
		<key>NSAllowsArbitraryLoads</key>
		<true/>
	</dict>

cnoon:
It works fine for me. I've migrated the example over to use all HTTPS calls which should play nice with ATS. Could you possibly provide more info and open a new PR if necessary? Could you check against the Alamofire 2.0.0-beta.1 release?

PR #667, opened 2015-08-10
updated iOS availability conditions

mackuba:
This adds missing minimum iOS version specifiers in some `#availability` clauses.

### DIFF
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -256,7 +256,7 @@ public class MultipartFormData {
 
         var isReachable = true
 
-        if #available(OSX 10.10, *) {
+        if #available(OSX 10.10, iOS 8.0, *) {
             isReachable = fileURL.checkPromisedItemIsReachableAndReturnError(nil)
         }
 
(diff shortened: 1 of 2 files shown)


kylef:
:+1: from me.

cnoon:
Thank you for your PR @mackuba. 

These changes are only to support iOS 7 and introduce compiler warnings into our main project. The Alamofire 2.0.0 release is going to drop support for iOS 7 altogether since we can't test it and give it the proper support it needs. Since we don't want to introduce warnings into the project, we're not going to be able to make this change.

Thanks again for the PR.

mackuba:
Wait, what warnings does it introduce? I assumed this wouldn't change anything at all if you only support iOS 8+...

cnoon:
It's not a necessary check if your deployment target is 8.0.

mackuba:
It's not, but it fixes the build for those who still want to build it on iOS 7, and I was assuming that it's completely harmless for those who build for iOS 8 only. What are the warnings that you get if you target iOS 8?

cnoon:
We're dropping support for iOS 7 with the Alamofire 2.0 release. Therefore, we don't want to introduce fixes for iOS 7 that generate compiler warnings for iOS 8 and iOS 9 users.

mackuba:
Maybe I'll just answer myself:

> Unnecessary check for 'iOSApplicationExtension'; minimum deployment target ensures guard will always be true

Makes sense in this case to not add it, I guess.

PR #687, opened 2015-08-14
Use custom parameter encoding even if `parameters` == nil

yanke-guo:


cnoon:
Thank you for your PR @yanke-guo. Could you provide me a good description of why you feel this change should be made? At the moment I don't feel this is a good change to make for the project.

cnoon:
Okay @yanke-guo, I'm going to close this out.

PR #696, opened 2015-08-19
@jriehn more flexible server trust policy via public method

jriehn:
Hello,

as given in the documentation "...it is important to have the flexibility to specify evaluation policies on a per host basis." In addition it should also be possible to enabled a mapping on domain level or even for wildcards. 

By enriching the visibility of the server trust policy we enable clients to implement specific requirements.

Cheers,

Jan

### DIFF
--- a/Source/ServerTrustPolicy.swift
+++ b/Source/ServerTrustPolicy.swift
@@ -42,7 +42,7 @@ public class ServerTrustPolicyManager {
         self.policies = policies
     }
 
-    func serverTrustPolicyForHost(host: String) -> ServerTrustPolicy? {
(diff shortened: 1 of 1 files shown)


cnoon:
Is the intention to be able to subclass the `ServerTrustPolicyManager` and override this method to be able to support things like wildcarded domain names? If that's the case, do we also need to open up the `policies` dictionary so you can access it?

```swift
public private(set) let policies: [String: ServerTrustPolicy]
```

cnoon:
Thanks for this @jriehn. I pushed your changes into the `swift-2.0` branch in 6c38ad01 and into the `master` branch in 8d659cc0. I also opened up the `policies` dictionary and added docstrings in subsequent commits.

Thanks again! 🍻

PR #706, opened 2015-08-24
beta 6 fix: replaced join with joinWithSeparator

mackuba:


cnoon:
Thank you for your PR @mackuba. Unfortunately I ended up merging the changes in #707 that fixed the iOS Example as well. Thanks again.

PR #686, opened 2015-08-13
Fixed error when project iOS Deployment Target is 7.0

Jamagas:
Screenshots indicating that issue exist on project with iOS Deployment Target 7.0:

screen shot 2015-08-13 at 14 58 41
screen shot 2015-08-13 at 14 59 14

PR #714, opened 2015-08-27
Augmented #availability checks to include iOS

scottdelly:
Xcode7 refuses to compile an iOS project without these additions


ikesyo:
See #686 and so on.

cnoon:
Thanks @ikesyo.

PR #716, opened 2015-08-27
Update ResponseSerialization.swift

brocoo:
Prevent unnecessary call to the response serialiser in case of an error picked up by the TaskDelegate.

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -106,11 +106,14 @@ extension Request {
         -> Self
     {
         delegate.queue.addOperationWithBlock {
-            var result = responseSerializer.serializeResponse(self.request, self.response, self.delegate.data)
-
-            if let error = self.delegate.error {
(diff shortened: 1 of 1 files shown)


Source/ResponseSerialization.swift
```diff
@@ -106,11 +106,14 @@ extension Request {
         -> Self
     {
         delegate.queue.addOperationWithBlock {
-            var result = responseSerializer.serializeResponse(self.request, self.response, self.delegate.data)
-
-            if let error = self.delegate.error {
-                result = .Failure(self.delegate.data, error)
-            }
+
```
cnoon: Could you kill this extra newline?

cnoon:
Great catch! Just one comment and then we can get it merged.

Source/ResponseSerialization.swift
brocoo: Done!

PR #726, opened 2015-09-01
Updated playground

jschmid:
To match with the new Result type


cnoon:
Thanks for your PR @jschmid! I pushed a slightly modified version of your changes in d4301808 giving you attribution.

Thanks again! 🍻

PR #731, opened 2015-09-02
CJK characters are broken.

cheolhee:
CJK letters were broken when I sent strings with POST method.
I think this charset should be a default set because the NSData is made with NSUTF8StringEncoding.

mutableURLRequest.HTTPBody = query(parameters!).dataUsingEncoding(NSUTF8StringEncoding, allowLossyConversion: false)

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -110,7 +110,7 @@ public enum ParameterEncoding {
                 }
             } else {
                 if mutableURLRequest.valueForHTTPHeaderField("Content-Type") == nil {
-                    mutableURLRequest.setValue("application/x-www-form-urlencoded", forHTTPHeaderField: "Content-Type")
+                    mutableURLRequest.setValue("application/x-www-form-urlencoded; charset=utf-8", forHTTPHeaderField: "Content-Type")
                 }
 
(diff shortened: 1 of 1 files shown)


cnoon:
Is there any way you could provide a failing test for this change? That would help us debug to make a more informed decision.

cnoon:
Thanks for your PR @cheolhee! 

After much investigation, I've decided to merge this change in 21040abd. I've also added a test around the POST behavior in ba6d3dd5.

With that said, I only made this change due to the fact you're seeing issues without the change. I cannot see any downsides to making the change. This [thread](http://stackoverflow.com/questions/16819502/application-x-www-form-urlencoded-and-charset-utf-8) makes me believe though that the `charset` should not really be required. Additionally, no actual unicode characters will ever make it into the HTTP body since all parameter key/value pairs are percent escaped prior to be encoded into UTF-8.

I'm assuming your server is requiring the `charset` when it actually shouldn't. Either way, as I mentioned, I cannot see any downsides to adding it.

For future reference, there's a good amount of information about the encoding algorithm [here](http://www.w3.org/TR/html5/forms.html#application/x-www-form-urlencoded-encoding-algorithm).

PR #741, opened 2015-09-05
Factored out cURLRepresentation() as NSURLRequest extension

dduan:
This will make `cURLRepresentation()` much more reusable. A user could
use it with an NSURLRequest object outside of Alamofile

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -23,7 +23,7 @@
 import Foundation
 
 /**
-    Responsible for sending a request and receiving the response and associated data from the server, as well as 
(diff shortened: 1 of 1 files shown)


cnoon:
Thank you for your PR @dduan...much appreciated! 

Unfortunately, I've decided not to merge this change into Alamofire. I definitely see where you were trying to go with this, but I think the fact that you need to hand it the `session` and the `credential` are simply proof that it is trying to do too much. I also think that by `nil`ing out the parameters, users could get confused why their `cURL` commands aren't working properly when they're not getting an accurate representation of the actual request.

Thanks again.

Issue #646, opened 2015-08-06
POST dont work very well


ErickB14:
I have a problem with one request POST in ios using swift ...

In curl i send one request like this:

curl -H "Authorization: token ce2800d53d520b4a73a005a611d53d299e0c1d5e"....
and me server response with 200

However en ios i make me request like this:

let parametersH = [
    "Authorization": "token 9640e65f66429415fb9359739ed8bc3f57cb0566"
]

Alamofire.request(Alamofire.Method.POST, "httpURL", parameters: parameters,headers:parametersH).progress { bytesWritten, totalBytesWritten, totalBytesExpectedToWrite in
    }.responseJSON{ request, response, JSON, error in
        if (error == nil) {
            println(response)
        }
        else{
            println(error)
        }
    }

	
But.....When i send the request for first time.....After install the application...The request....is correct and send one code 200 But... i close the application, and when in run again... i send the same request....(In this moments i hardcode the tokens and the url , for reason of test ) the server responds with un 403 .... I dont know what happened...why the first time work very good and the second not work... The token ...when i test this case is valid..

cnoon:
Hi @ErickB14,

Instead of creating a support issue in Github, please open a question on [Stack Overflow](https://stackoverflow.com) and tag `Alamofire`. The Alamofire community is very helpful and supportive. Stack Overflow is the best place to get your question answered quickly and correctly.

Best of luck!

---

From our [Contribution Guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md)

## Asking Questions

We don't use GitHub as a support forum. For any usage questions that are not specific to the project itself, please ask on [Stack Overflow](https://stackoverflow.com) instead. By doing so, you'll be more likely to quickly solve your problem, and you'll allow anyone else with the same question to find the answer. This also allows maintainers to focus on improving the project for others.

gburlet:
I also am running into this issue and have been debugging for a day. I have not found any answers on StackOverflow for this specific issue, including the StackOverflow post that the OP made. I have tried `encoding: .JSON`, `encoding:  .URLEncodedInURL`, as many of the project Github issues suggest as well as several other combinations of parameters and attempts.

Curl works great. Retrofit (Java REST request library for Android) works great. Alamofire successfully makes a POST request on the first attempt and then results in a 403 thereafter.

Code:
```
        let url = apiRoot + "auth/login/"

        let parameters = [
            "username": username,
            "password": password
        ]
        
[...]

Issue #623, opened 2015-07-28
[Swift 2] POST requests don't get parameters added.


BasThomas:
Doing a simple request with GET adds my parameters nicely, but doing a POST doesn't.

```swift
let parameters = ["foo": "bar"]

Alamofire.request(.POST, "url", parameters: parameters)
.responseJSON { request, response, json, error in
  print("request: \(request)")
}
```

returns

`request: Optional(<NSMutableURLRequest: 0x7f9864109cb0> { URL: https://api.github.com/repos/BasThomas/junk/issues })`

while

```swift
let parameters = ["foo": "bar"]

Alamofire.request(.GET, "url", parameters: parameters)
.responseJSON { request, response, json, error in
  print("request: \(request)")
}
```

returns

`request: Optional(<NSMutableURLRequest: 0x7f9ef07ef0f0> { URL: https://api.github.com/repos/BasThomas/junk/issues?foo=bar })`

Am I missing something here?

cnoon:
Hi @BasThomas,

Instead of creating a support issue in Github, please open a question on [Stack Overflow](https://stackoverflow.com) and tag `Alamofire`. The Alamofire community is very helpful and supportive. Stack Overflow is the best place to get your question answered quickly and correctly.

Best of luck! 🍻

---

From our [Contribution Guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md)

## Asking Questions

We don't use GitHub as a support forum. For any usage questions that are not specific to the project itself, please ask on [Stack Overflow](https://stackoverflow.com) instead. By doing so, you'll be more likely to quickly solve your problem, and you'll allow anyone else with the same question to find the answer. This also allows maintainers to focus on improving the project for others.

jonauz:
I'm getting the same issue, and I starting to believe it is an issue within Alamofire core code.
I did found solutions on Stack Overflow, but they are work arounds. I think it's more important to solve it properly.

BasThomas:
You should probably set the encoding to `.JSON`.

jonauz:
Tried, didn't helped.
Tried to test with all mentioned methods that made sense, but no luck. Backend wasn't receiving any parameters using POST method.
Unless I add them straight to the URL part as postfix "...?key=value"
I also used NSURLSession and NSURLRequest in the past, so I know backend part worked. And it works when I test with Postman (i.e. plugin on Chrome).

jonauz:
[...]

Issue #374, opened 2015-03-06
Multiple encoding methods in single request


AnthonyMDev:
I need to make a `POST` request with an HTTP Body with a `JSON` object, but I also need to use url query parameters in the same request.

```
POST: http://www.mysite.com/api/create?param1=value&param2=value
HTTP Body: { foo : [ bar, foo], bar: foo}
```

Is this supported? How would I go about doing this?

Thanks

AnthonyMDev:
I posted this on StackOverflow
http://stackoverflow.com/questions/28908680/multiple-encoding-types-for-alamofire-request

If it's already supported and I just need a usage example, then it should probably be answered there. If this is not a supported feature yet, it seems like something work discussing implementing here.

AnthonyMDev:
At this point, I've decided to solve this by manually encoding an `NSURLRequest` with the URL parameters, retrieving the URL from that request, and using that to create the final request. I've created a function to return the query parameter encoded request:
```
    private func queryParameterEncodedRequestURL(urlString: String,
    values: [String]) -> NSURL {
      
      let URL = NSURL(string: urlString)
      var request = NSURLRequest(URL: URL)
      
      let parameters = [
      "param1": values[0]!,
      "param2": values[1]!
      ]
      
      let encoding = Alamofire.ParameterEncoding.URL
      (request, _) = encoding.encode(request, parameters: parameters)
      
      return (request.URL, nil)
    }
```
This works fine, but I would definitely like to see Alamofire support multiple encoding types more easily. This feels like a workaround to me.

@mattt What is your opinion on implementing a native way to handle multiple encodings?

mattt:
Multiple parameter encodings isn't really a common convention for web applications, so it's not something I'd be interested to build any more specific support for. That said, `ParameterEncoding`, with it's `Custom` case and associated closure value can be used to compose a request without reimplementing either JSON body or URL query string encoding. The solution you described should work just fine.

mxl:
@mattt I think it's not correct to name it `ParameterEncoding.URL` if it encodes parameters depending on request method.

mattt:
@mxl Why not? It's encoded in the URL for HTTP methods that do not support am message body, or in the body as `application/x-www-form-**urlencoded**` (emphasis added) for HTTP methods that designate a message body. That's just HTTP.

mxl:
[...]

Issue #283, opened 2014-12-31
POST not sending parameters


brunomacf:
My post request with alamofire is not sending the parameters!! This is my request:

```
func login(username: String!, password: String!, callback: (NSError?, AnyObject?) -> Void) {
        var params = [
            "login": username,
            "password": password
        ]
        
        Alamofire.request(.POST, baseURL+"login", parameters: params)
            .validate()
            .responseJSON {(request, response, data, error) in
                if(error != nil) {
                    var errMsg = data?["error"] as String
                    
                    println(response?.statusCode)
                    println(errMsg)
                }
        }
```
And i'm getting from the server
```
Invalid login: undefined
```
The request data in responseJSON function is 
```
<NSMutableURLRequest: 0x7f8d4b751670> { URL: https://smarttbot.com/api/v1/login }
```
I already tried pass encoding: .JSON in the request function, but the problem persists! What i'm doing wrong?

mattt:
POST requests don't send parameters in the URL.

This is not a support forum. Please take any specific usage questions to Stack Overflow in the future.

PR #742, opened 2015-09-06
Added `URLEncodedInURL` case to Parameter Encoding

cnoon:
This has been a long withstanding problem that has been reported and questions many times on GitHub or Stack Overflow (#283, #374, #623, #646). Issue #374 has a wealth of information about the different options and a healthy debate between @mattt and @mxl. For this particular issue, I tend to side with @mxl.

Having personally had to deal with services constantly that require appending parameters to the query string for .POST and .PUT methods, I felt it was time to add support into Alamofire to make this situation more palatable for everyone. The main challenge is avoid complicating the `ParameterEncoding` enumeration to keep it simple even though we need to add another case or condition. I think this change strikes a nice balance.

I tried to update the documentation to make it as clear as possible to understand the differences between the two cases.

Other alternatives I consider were `.URLEncodedInQueryString` (too verbose), `.URLQuery` (too vague), and also adding a flag to the `.URL` case such as `.URL(encodedInURL: Bool)`. The problem with this approach is that you cannot specify a default parameter in an associated value.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -36,23 +36,32 @@ public enum Method: String {
 /**
     Used to specify the way in which a set of parameters are applied to a URL request.
 
-    - URL:          A query string to be set as or appended to any existing URL query for `GET`, `HEAD`, and `DELETE` 
-                    requests, or set as the body for requests with any other HTTP method. The `Content-Type` HTTP header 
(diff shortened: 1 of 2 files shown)


cnoon:
cc @kcharwood and @kylef

kcharwood:
I think this looks like a pretty sensible solution. I'm cool with it.

PR #748, opened 2015-09-08
Add compatibility checks for pre-iOS 8.

bojan:
When the library is included as source for projects targetting iOS 7.0, the compiler fails because of this compatibility check.


cnoon:
Thank you for your PR @bojan. We are not going to support iOS 7 in Alamofire 2.0. Please see #554, #525, #523, etc. for more info.

PR #759, opened 2015-09-11
Update installation via Carthage instructions on README

barbosa:
Same change made here https://github.com/Alamofire/Alamofire/commit/343e24073990172de3c55871762f68d49838a3ce, but for Carthage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ $ brew install carthage
 To integrate Alamofire into your Xcode project using Carthage, specify it in your `Cartfile`:
 
 ```ogdl
-github "Alamofire/Alamofire" "swift-2.0"
+github "Alamofire/Alamofire" ~> 2.0
 ```
 
 ### Manually


README.md
```diff
@@ -85,7 +85,7 @@ $ brew install carthage
 To integrate Alamofire into your Xcode project using Carthage, specify it in your `Cartfile`:
 
 ```ogdl
-github "Alamofire/Alamofire" "swift-2.0"
+github "Alamofire/Alamofire" ~> 2.0
```
ikesyo: This should be `2.0.0`.

barbosa: Even if it is `2.0` for CocoaPods? (see https://github.com/Alamofire/Alamofire/commit/343e24073990172de3c55871762f68d49838a3ce)

Does Carthage handle this in a different way?

ikesyo: `2.0` and `2.0.0` are both OK for Carthage, but `2.0.0` is more appropriate for semver.

kylef: `~> 2.0.0` and `~> 2.0` mean different things. `~> 2.0` means that any minor version of 2.0 is allowed whereas `~> 2.0.0` means only patch level changes (minor and major must be 2.0).

kylef: It should be `~> 2.0` because semver guarantees API compatibility between minor releases.

barbosa:
Thanks @kylef

ikesyo:
@kylef Thank you for the clarification.

PR #767, opened 2015-09-15
tvOS support

bscarano:
"Alamofire tvOS" framework + "Alamofire tvOS tests" targets.  Tests all pass and tested in a separate sample app.


cnoon:
Thanks for your PR @bscarano...much appreciated! 

I merged a slightly modified version of your PR in 8873379b. I wanted to make sure all the settings were correct, so I created the framework and test targets myself going through each step carefully. I still gave you attribution for the change. I also want to keep this separated from the `master` branch until it's officially released.

Thanks again! 🍻

PR #780, opened 2015-09-17
Fix iOS target by moving dependency to test bundle

nickygerritsen:
It seems that in 8873379b67185e220eb09b05e8227ed19a0d296f `Alamofire iOS` got a dependency on its own. This should actually be a dependency for the test bundle so I took the liberty to move this dependency.


cnoon:
Thanks for the PR @nickygerritsen! I'm not exactly sure how that managed to get messed up in 8873379, but I'm glad it's fixed!

Thanks again! 🍻

Issue #779, opened 2015-09-17
result.data is nil when .responseJSON is used


rafaelc0sta:
When .responseJSON is used, and if result.value isn't nil, result.data is nil. Although this does not make the library unusable, it is still unexpected behaviour: You should be able to access BOTH the JSON value and the NSData raw response.

rafaelc0sta:
Update: if it is needed, I believe I have fixed this, I can submit a pull request.

jshier:
I'd find this useful as well. I ran into this issue when was trying to put some logging into my generic response handler.

cnoon:
Could you both give me a bit better description of what you are trying to do and why the current design of the `.Success` case doesn't give you the flexibility that you need?

jshier:
@cnoon My usage case is rather simple. Take this code, which uses my implementation of `GenericResponseSerializer` and breaks the `Result` apart for my completion blocks.
```swift
func responseObject<T: Decodable where T == T.DecodedType>(completionHandler: (responseObject: T?, error: CloudFlareError?) -> Void) -> Self {
    let fullCompletionHandler = { (request: NSURLRequest?, response: NSURLResponse?, result: Result<T>) in
        switch result {
        case let .Success(value):
            completionHandler(responseObject: value, error: .None)
        case let .Failure(_, error):
            if let error = error as? CloudFlareError {
                completionHandler(responseObject: .None, error: error)
            }
            else if let error = error as? NSError {
                // Must have received a network error.
                completionHandler(responseObject: .None, error: CloudFlareError.NetworkError(error: error))
            }
            else {
                fatalError("ResponseObject received an error that was neither a CloudFlareError nor an NSError.")
            }
        }
    }
    
    return response(responseSerializer: Request.CloudFlareResponseSerializer(), completionHandler: fullCompletionHandler)
}
```
What I'd like to be able to do is call a logging function from within `responsObject`. My old implementation of that method essentially wrapped `responseJSON` and so had access to all of the intermediate information and errors. This new version makes that much more difficult, and for the `.Success` case, essentially impossible. Now, I'd be content to only log the model object created and not any of the intermediate state (string, JSON, Argo JSON), if only I could figure out how to create a log function that can be called with a generic parameter. But in the end, having `data` even in the success case would make this logging trivial, even if I had to redo some of the parsing (this logging isn't enabled for release and so isn't really performance critical).

rafaelc0sta:
[...]

Issue #757, opened 2015-09-10
Result ErrorType Generic


AnthonyMDev:
Why doesn't `Result` have two generics, one for the value and one for the error type?

This would allow us to know what type of error we should expect and access properties on the error based on its type. Custom serializers should still be able to return their own `ErrorType`, while the current `Alamofire` serializers can use a `Result` with `NSError`.

klundberg:
there's a dedicated result framework here: https://github.com/antitypical/Result
Alamofire should just have this as a dependency and extend it as needed.

AnthonyMDev:
@klundberg This framework seems to do what I'm describing. Either way, it doesn't really hurt for Alamofire to use it's own `Result`. It is easy to implement, and allows for more possible customization later. 

But having the error being part of the generic seems like the right way to go.

cnoon:
@AnthonyMDev thanks for your feedback. You are absolutely right and it should be implemented as a double generic `Result`.

For a bit of background, the `Result` type originally took an `NSError`. When working with @jshier in #732, I changed the `Result` to take an `ErrorType` 64a00447. Instead, I should have made the `Result` a double generic and forced the error to conform to `ErrorType`. 

I'm currently working with the ASF to figure out how to proceed here. I'll update the issue once we decide what we're going to do. Thanks again for bringing this to our attention.

Rupert-RR:
I would vote to use the dedicated `Result` framework too — there's no point in re-inventing the wheel, and this would allow compatibility with other frameworks that use it.

cnoon:
We will not be adding a dependency for a Result type @Rupert-RR. Please see #764 for more info.

cnoon:
Okay @AnthonyMDev, we've finally landed on how to attack this issue and I've just pushed up #791 which addresses the core design flaw. If you could check it out and provide feedback, I'd certainly appreciate it. I'm going to close this issue out. Please redirect all future comments to #791.

Issue #732, opened 2015-09-02
Use ErrorType instead of NSError


jschmid:
Sorry to be late to the party, but I was wondering why you used NSError instead of the `ErrorType` protocol. This is related to PR #627.

My REST client uses Alamofire and in case of error returns an enum conforming to `ErrorType`. I also want to add custom validation. It feels a bit wrong to have my validation return an `NSError` then having to translate it later into my `ErrorType` when I return my result. 

From what I saw, in the code, all errors are bound to `Error.Code` so there will never be an "extra" NSError somewhere that prevents to use `ErrorType`.

Any thoughts?

jschmid:
Also the `Result` type does indeed use `ErrorType` and not `NSError`.

cnoon:
We have now transformed the `Result` type and all uses of it to use the `ErrorType` for exactly that reason. We continue to generate `NSError` objects inside Alamofire to guarantee that that is the only error type that will ever come out of Alamofire. Using the `ErrorType` though in the `Result` type gives all developers the freedom to wrap all Alamofire error in their own `ErrorType` enumeration.

jschmid:
So you confirm that if I want to add a custom validation step, I will need to return an `NSError` wrapped in a `ValidationResult.Failure(NSError)`, then have to unwrap in in my `responseString` call to convert it into my custom ErrorType ?

That odd. I really don't get why a `ValidationResult.Failure` would always require an `NSError`. 

For example with a custom validation:


    enum MyErrors: ErrorType {
        case LeetError
    }

    let myDomain = "myDomain"

    func callback(val: String?, _ error: ErrorType?) {
        print(val)
        print(error)
    }

    Alamofire.request(.GET, "http://httpbin.org/get")
        .validate({ (req, res) -> Request.ValidationResult in
            let nserror = NSError(domain: myDomain, code: -1337, userInfo: nil)
            return .Failure(nserror)
        })
        .responseString { _, _, result in
            switch result {
            case .Success(let val): callback(val, nil)
            case .Failure(_, let error as NSError) where error.domain == myDomain && error.code == -1337:
                callback(nil, MyErrors.LeetError)
            case .Failure(_, let error): callback(nil, error)
            }
    }

I feel that my custom validation should be allowed to return directly a `MyError`, and it would not break anything from Alamofire. Since I do my own validation, I have to be aware that I will receive any kind or error.

What do you think ?

jshier:
[...]

PR #791, opened 2015-09-19
Feature - Double Generic Result Type

cnoon:
This PR is a non-backwards compatible set of changes solving several shortcomings of the current `Result` and `ResponseSerializer` design. Here is a quick list of callouts:

* `Result` type now takes two generic parameters (Value and Error) where the Error conforms to `ErrorType`.
    > No longer stores the `data` in the `.Failure` case.

* All response serializers now return the original server data as an `NSData` optional.
    > This solves all the issues reported in #779.

* The `TaskDelegate` now stores the error as an `NSError` instead of an `ErrorType`.
    > This was changed in #732 and this should not have been. With these changes, consumers no longer need to cast the `error` to an `NSError`. 

* The `ValidationResult` failure case now requires an `NSError` and not an `ErrorType`.
    > This means that a custom `Validation` closure MUST return an `NSError`. The change made to address #732 should NOT have been made. All errors generated by Alamofire (including custom Validation) need to generate `NSError` types. Otherwise we force all consumers to have to cast all error objects returned from Alamofire even though they are not using their own custom validation error types.

For more details, please refer to #732, #757 and #779.

### DIFF
--- a/Example/DetailViewController.swift
+++ b/Example/DetailViewController.swift
@@ -75,7 +75,7 @@ class DetailViewController: UITableViewController {
         refreshControl?.beginRefreshing()
 
         let start = CACurrentMediaTime()
-        request.responseString { request, response, result in
+        request.responseString { request, response, _, result in
             let end = CACurrentMediaTime()
             self.elapsedTime = end - start
 
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -193,7 +193,7 @@ public class Request {
         let progress: NSProgress
 
         var data: NSData? { return nil }
-        var error: ErrorType?
+        var error: NSError?
 
         var credential: NSURLCredential?
 
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -30,11 +30,12 @@ import Foundation
 public protocol ResponseSerializer {
     /// The type of serialized object to be created by this `ResponseSerializer`.
     typealias SerializedObject
+    typealias Error: ErrorType
 
     /**
         A closure used by response handlers that takes a request, response, and data and returns a result.
     */
-    var serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> Result<SerializedObject> { get }
+    var serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> Result<SerializedObject, Error> { get }
 }
 
 // MARK: -
@@ -42,14 +43,17 @@ public protocol ResponseSerializer {
 /**
     A generic `ResponseSerializer` used to serialize a request, response, and data into a serialized object.
 */
-public struct GenericResponseSerializer<T>: ResponseSerializer {
+public struct GenericResponseSerializer<T, E: ErrorType>: ResponseSerializer {
     /// The type of serialized object to be created by this `ResponseSerializer`.
     public typealias SerializedObject = T
+
+    /// The type of error to be created by this `ResponseSerializer` if serialization fails.
+    public typealias Error = E
 
     /**
         A closure used by response handlers that takes a request, response, and data and returns a result.
     */
-    public var serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> Result<SerializedObject>
+    public var serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> Result<T, E>
 
     /**
         Initializes the `GenericResponseSerializer` instance with the given serialize response closure.
@@ -58,7 +62,7 @@ public struct GenericResponseSerializer<T>: ResponseSerializer {
 
         - returns: The new generic response serializer instance.
     */
-    public init(serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> Result<SerializedObject>) {
+    public init(serializeResponse: (NSURLRequest?, NSHTTPURLResponse?, NSData?) -> Result<T, E>) {
         self.serializeResponse = serializeResponse
     }
 }
@@ -77,7 +81,7 @@ extension Request {
     */
     public func response(
         queue queue: dispatch_queue_t? = nil,
-        completionHandler: (NSURLRequest?, NSHTTPURLResponse?, NSData?, ErrorType?) -> Void)
+        completionHandler: (NSURLRequest?, NSHTTPURLResponse?, NSData?, NSError?) -> Void)
         -> Self
     {
         delegate.queue.addOperationWithBlock {
@@ -102,17 +106,11 @@ extension Request {
     public func response<T: ResponseSerializer, V where T.SerializedObject == V>(
         queue queue: dispatch_queue_t? = nil,
         responseSerializer: T,
-        completionHandler: (NSURLRequest?, NSHTTPURLResponse?, NSData?, Result<V>) -> Void)
+        completionHandler: (NSURLRequest?, NSHTTPURLResponse?, NSData?, Result<V, T.Error>) -> Void)
         -> Self
     {
         delegate.queue.addOperationWithBlock {
-            let result: Result<T.SerializedObject> = {
-                if let error = self.delegate.error {
-                    return .Failure(self.delegate.data, error)
-                } else {
-                    return responseSerializer.serializeResponse(self.request, self.response, self.delegate.data)
-                }
-            }()
+            let result = responseSerializer.serializeResponse(self.request, self.response, self.delegate.data)
 
             dispatch_async(queue ?? dispatch_get_main_queue()) {
                 completionHandler(self.request, self.response, self.delegate.data, result)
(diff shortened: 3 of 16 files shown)


cnoon:
cc @kcharwood and @kylef

cnoon:
I would also love feedback here from @jshier, @AnthonyMDev and @jschmid if you have time to check it out.

jshier:
@cnoon In this new setup, how am I supposed to handle underlying errors when I'm using my own `GenericResponseSerializer`? As far as I can tell, there's no way to access the errors from the network or even Alamofire's own validation. Even the built in serializers don't have access. What am I missing?

cnoon:
Hi @jshier, you aren't missing anything. What you are proposing has never actually been possible. In Alamofire 2.0 your response serializer just isn't called if there was error that occurred prior to response serialization. 

Let me dig through this issue a bit more and see if there's something we can do to make this a bit easier. Off the top of my head I'm thinking that the delegate error (if one exists) should be passed through to the response serializer.

cnoon:
Okay @jshier, after digging in a bit more, I realized that the issue was MUCH worse than I originally thought. Thank you for pointing that out! I owe you a beer for that one. 🍻

I've addressed the issue using the approach I mentioned previously in 025f33bc. All response serializers now receive the delegate error as a parameter in the `serializeResponse` closure.

IMO, this has three major benefits. 1) It fixes the major bug that I introduced 😉 2) Ensures all response serializers are always actually executed 3) Allows each response serializer to inspect the error and figure out how to proceed. Certain types of errors will actually cause the server to return different payloads. This abstraction would actually allow you to use different parsing schemes based on the error type.

jshier:
Looks great. This accomplishes all of goals for my custom response serializer, lets me do logging, and even allows some simplification of code. Plus, I can now have custom serializer behavior based on incoming errors. :+1: 

Are the changes you made here compatible with #792? I haven't looked at it closely but it looks like a good change.

cnoon:
Awesome! Yes #792 is PR'd on top of this one. The combination of the two is what I'm thinking response serialization for Alamofire 3.0 looks like.

ziogaschr:
:+1:  I really like this new PR, thanks @cnoon. Do you have an ETA for when it will merged to master?

cnoon:
Thanks @ziogaschr! I'm looking to get it merged into master today or tomorrow. Need to get the ASF to make a review pass here. Once #791, #792 and #795 are merged, I'll be cutting Alamofire 3.0.0-beta.1 to get testing going.

AnthonyMDev:
@cnoon Sorry I wasn't able to look at this over the weekend. I'm very happy with all the changes up to this point. I'll be waiting for `3.0.0-beta.1` as I'm planning on implementing these changes in my own projects ASAP. Thank you!

kcharwood:
This looks good to me @cnoon. Do we need to make another migration change, and explain why we are bumping soon, and the clear benefits of upgrading?

AnthonyMDev:
@cnoon Is there a branch that we can install as a pod while we wait for the ASF? I'd like to start integrating this into a current project ASAP.

jshier:
@AnthonyMDev Check the branch this PR is pulling from. I used it in a CocoaPods project to test my own code and it worked fine.

AnthonyMDev:
@jshier thanks. I knew there were multiple new features coming from PR #791, 792, and #795. Wasn't sure if there was one branch that incorporated all of them.

cnoon:
@kcharwood awesome...thanks for digging through this! 👊🏼 

As for the migration guide, I'll be putting that together here in the next couple of days. I'll be sure to cover all your suggestions in detail.

@AnthonyMDev I'm in the process of merging all these PRs right now and cutting the Alamofire 3.0.0-beta.1 release. I won't have a migration guide put together for a couple of days, but I'm sure you can figure out the API differences from the PRs and the test code.

Everyone else, thanks for helping get this through! 🍻

jschmid:
Agin I'm late to the party. Shouldn't have taken a long weekend off :open_mouth:
Apparently everyone agrees that this PR improves the library. I'll add my 2 cents anyway.

Prior to this, I was using the `validate()` method to validate things prior to any deserialization. For example if for one request I needed a HTTP 201 response, I would check it and return a `ErrorType` error, not a `NSError`. 
My client returns `ErrorType` errors. I'm not going to wrap my error into a `NSError` to comply with the `validate()` method, then unwrap it in the response serializer. The `validate()` method suddenly became useless for me and I'll to do the validation in the serializer.
For me the job of the *serializer* is to *serialize*, not check HTTP statuses or handle any kind of validation. I feel that I'll have to write a thousand-lines serializer like the one @jshier showed in [this comment](https://github.com/Alamofire/Alamofire/issues/732#issuecomment-137142628) whereas before I could chain simple validations before doing any serialization (which was not even needed since there was an error).

I have other problems related to the way I handle the response after it has been serialized, but I guess it's only because I tried to do a convoluted generic responseSerializer.

ziogaschr:
Can it be that merge of this PR and #792 are in conflict. And the one replaced the code needed by the other?

jshier:
@jschmid I think you've misunderstood what this PR (and the other AF3 PRs) require on the part of the developer. I don't know what your response serializer works or how you're generating your errors, but this new pattern doesn't require the use of 'NSError'. I refactored the serializer I showed in that comment you linked to down to [this] (https://gist.github.com/jshier/f08f08e05d994b5bdc7d). To handle the new error coming into the serializer (underlying errors and Alamofire validate errors both), all I had to do was `guard` against it and wrap it in my custom error type if it existed. Also, please remember that my serializer was so complicated because 1) I was doing custom JSON deserialization using Argo and 2) I had to check the partially parsed response for success, not just whether it parsed. Perhaps if you posted what you were trying to do

jschmid:
Thanks @jshier for the reply. 

As I stated in #732, it thought i would be great to be able to return `ErrorType` errors in the **validation process**. (Not returning an error from the `ResponseSerializer`)

As a simple example I had:

    extension Request {
        func validateIsStatus(code: Int, _ error: ErrorType) -> Self {
            return validate { _, res in
                if res.statusCode != code {
                    return Request.ValidationResult.Failure(error)
                } else {
                    return Request.ValidationResult.Success
                }
            }
        }
    }

That I could use like:

    enum ClientError: ErrorType {
      case MyError(Int)
    }

    let request: Request =  ...

    request
    .validateNotStatus(401, ClientError.MyError(401))
    .validate()
    .responseObject(completionHandler: { ... })

Now you see that I use the validation to return custom errors. But this PR prevents returning `ErrorType` errors now.

I don't want the serializer to do any serialization if there was a validation error. Now I just have to do it and do the custom validation in the callback of my custom serializer.

    request
    .validate()
    .responseObject { response, obj, error in
      if response?.statusCode == 401 {
        callback(. MyError(401))
        return
      }

      // Handle obj here
    }

Do you see what I mean here?

jschmid:
Note that maybe I'm using the validation for the wrong purpose, and I should do this in the response callback anyway.

jshier:
Are you using your own `ResponseSerializer`? If so, you can easily replicate this functionality while not parsing the response by using a `guard` at the top of the serializer, as I did in the refactored serializer I linked to above. 

If you're not using your own `ResponseSerializer`, you're right, it does make your case somewhat less elegant. Perhaps @cnoon would have some ideas. 

Additionally, now that the PR has landed on master, you may want to open an issue regarding not being able to return custom error types from `.validate()`.

cnoon:
Okay @jschmid, while I appreciate your feedback, you are using validation and response serialization incorrectly. You should not be creating a custom validation extension to validate the status code, that EXACT functionality already exists in Alamofire and all the error handling is taken care of by the response serializers. If you need to handle a validation failure in the response serializer completion, just switch on the error code like you do for all other types of errors.

Additionally, why would you want to start throwing your own custom errors when you aren't creating your own custom response serializers? That doesn't make sense to me. IMO you're just making your life more difficult...on purpose! Instead, keep all the errors coming out of Alamofire as NSErrors and you can just switch on the error code. Why would you want to have to switch on NSError codes AND your custom error type? That's actually MORE complicated.

By locking down the error type stored in Alamofire to ALWAYS have to be an NSError, it allows all response serializers to be strongly typed which means you don't have to cast the error to an NSError every time you need to access it. This is the 99% case for the Alamofire community. Power users are like @jshier are certainly going to create custom response serializers tailored to their own needs, but the common usage pattern should be as simple as possible. This fits with our overall design philosophy which is to cater Alamofire to the masses by making it as simple and easy to use as possible, while flexible enough to serve the power users.

Locking down the NSError internal type allows us to support some powerful features and more consistent APIs:

* Strongly typed NSError returned by the no-op response tuple method
* Strongly typed errors returned by all generic response serializers
    > No more casting to NSError

* Custom response serializers can specify their own strongly typed error encapsulating the NSError being produced by Alamofire if necessary (see @jshier's sample code)
* Consistent generic usage patterns between ResponseSerializer, Response and Result types

Hopefully that helps clear up any possible confusion as well as shed some light on the big picture.

jschmid:
Thank you for the thorough answer. 

I understand the end goal of having only NSErrors. I do have a custom serializer. I guess I'll have to rewrite it a bit to handle my errors there.

Thanks again!

PR #792, opened 2015-09-19
Feature - Response Generic Type

cnoon:
This PR creates a generic `Response` type to help simplify response serialization. Instead of returning a tuple with four different values, response serializers now return a single generic `Response` object. This makes it MUCH easier to use response serializers, and will help users avoid making mistakes such as incorrectly naming the tuple parameters.

I've seen MANY cases of people swapping the `request`, `response`, `data` and `result` parameters in the tuple. They constantly get misused with tends to lead to a large amount of confusion. Additionally, it can cause strange compiler errors that are quite difficult to debug. This PR attempts to address these issues by replacing the tuple with a generic `Result` type with a generic `Response` struct.

### Old Usage

```swift
Alamofire.request(.GET, "http://httpbin.org/get", parameters: ["foo": "bar"])
    .response { request, response, data, result in
        print(request)
        print(response)
        print(data)
        debugPrint(result)
    }
```

### New Usage

```swift
Alamofire.request(.GET, "http://httpbin.org/get", parameters: ["foo": "bar"])
    .response { response in
        debugPrint(response)
 
        // or
        
        print(response.request)
        print(response.response)
        print(response.data)
        debugPrint(response.result)
    }
```

This should really help alleviate misuse and should simplify response serialization usage with the concrete struct.

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -28,6 +28,9 @@
 		4C341BBA1B1A865A00C1B34D /* CacheTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C341BB91B1A865A00C1B34D /* CacheTests.swift */; };
 		4C341BBB1B1A865A00C1B34D /* CacheTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C341BB91B1A865A00C1B34D /* CacheTests.swift */; };
 		4C3C53B21BAC5D40004F7F68 /* Alamofire.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F8111E3319A95C8B0040E7D1 /* Alamofire.framework */; settings = {ASSET_TAGS = (); }; };
+		4C3C53DF1BADEF79004F7F68 /* Response.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4C3C53DE1BADEF79004F7F68 /* Response.swift */; settings = {ASSET_TAGS = (); }; };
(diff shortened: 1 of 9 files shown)


cnoon:
cc @kcharwood and @kylef

cristeahub:
I would suggest adding an updated README along with this to help avoid confusion.

cnoon:
Absolutely @cristeahub. I'd rework all the README code samples and put together another migration guide.

marcelofabri:
Another advantage of this is that if another parameter is added (or removed), people won't have to change every `response` call.

tobiasoleary:
@macelofabri agreed! Recently upgraded a project to Swift 2.0. Even with regular expressions updating the code for the latest version of Alamofire was a pain.

kcharwood:
I like this change as well. Will definitely make it easier to modify the object in the future if we need too, although it will be one more round of upgrades to every call site @tobiasoleary :beers:

kylef:
:+1:

cristeahub:
Quick question here.

The `Alamofire.response` function doesnt use this system here as seen in the [documentation](https://github.com/Alamofire/Alamofire#response-handler) and [here](https://github.com/Alamofire/Alamofire/blob/238ebc3f9af52910fa69bed404a816d75d89cfc2/Source/ResponseSerialization.swift#L86). Was this only intended for `responseJSON` and similar?

cnoon:
It was definitely intended @cristeahub. The `response` method simply forwards all the data collected from the `SessionDelegate` without any additional processing. The others are all customized to process the data in a specific way.

Issue #794, opened 2015-09-20
Make Automated Testing Easier


troya2:
When writing an API with Alamofire as the enabling technology, I'd like to be able to write tests using something like DVR (http://blog.soff.es/network-testing-in-swift-with-dvr).  For that to work, there needs to be a way to replace the NSURLSession with a subclass of our own design.

Can anyone provide a suggestion on how to test with this method?  Alternatively, would it make sense to supply a NSURLSession to the Manager's initializer?

jshier:
From my limited understanding, it seems like initializing a `Manager` with a custom session is essentially impossible, as Alamofire takes over the session's delegate. This is core to the Alamofire architecture and likely can't be worked around.  

For my own testing, I use `OHHTTPStubs`. Recording is manual, but no modification to the networking code is necessary.

kylef:
You could also use something like [Mockingjay](https://github.com/kylef/Mockingjay) as that instead registers a custom NSURLProtocol to the `NSURLSession` that Alamofire would use.

cnoon:
Hi @troya2, thanks for opening this issue. While Mockingjay and OHHTTPStubs certainly have their place and are good solutions in most cases, I don't think that relying on an `NSURLProtocol` should be the only way to perform this type of customization.

Therefore, I have made some changes in #795 to make this possible. I've actually been meaning to rework the initialization of the `Manager` for a bit anyways. You opening this ticket gave me enough of a push to dig in and get it updated. The changes in #795 now use DI to inject a custom `NSURLSession` or a `SessionDelegate`, depending on your use case. If everyone could take a look at that PR and give feedback, I'd greatly appreciate it.

I'm going to close this issue out. Please direct all future comments to #795.

Thanks again! 🍻

PR #795, opened 2015-09-20
Feature - Manager Initialization

cnoon:
This PR opens up the `Manager` initialization to allow users to use dependency injection for both the underlying `NSURLSession` and `SessionDelegate`. 

#### Using Dependency Injection for the URL Session

This change was really made due to #794. While libraries like Mockingjay and OHHTTPStubs are one way to perform automation, Alamofire should also allow a bit more control over the URL session creation to allow approaches like DVR to be possible. This PR adds a failable initializer allowing this to be possible.

#### Using Dependency Injection for the Session Delegate

The current design of the `Manager` initialization is tailored to starting up your app, setting up a `Manager` instance and start making requests. It does not however support background session callbacks very well. 

The problem is that you need to configure the override closures on the `SessionDelegate` before actually creating the NSURLSession. Otherwise, you could start receiving delegate callbacks before you've had a chance to hook up the override closures. This PR adds the ability to use dependency injection to pass in a pre-configured `SessionDelegate` that already has the override closures set.

### DIFF
--- a/Source/Manager.swift
+++ b/Source/Manager.swift
@@ -114,10 +114,12 @@ public class Manager {
     // MARK: - Lifecycle
 
     /**
-        Initializes the `Manager` instance with the given configuration and server trust policy.
+        Initializes the `Manager` instance with the specified configuration, delegate and server trust policy.
 
         - parameter configuration:            The configuration used to construct the managed session. 
                                               `NSURLSessionConfiguration.defaultSessionConfiguration()` by default.
+        - parameter delegate:                 The delegate used when initializing the session. `SessionDelegate()` by
+                                              default.
         - parameter serverTrustPolicyManager: The server trust policy manager to use for evaluating all server trust 
                                               challenges. `nil` by default.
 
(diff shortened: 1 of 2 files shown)


cnoon:
cc @kylef and @kcharwood

kcharwood:
:+1:

kylef:
:+1:

dgdosen:
So - I'm trying to use this in the context of passing a custom NSURLSession (using DVR) when I instantiate an Alamofire.Manager.  I only want to bring my custom NSURLSession to the party.

I can pass in my session, but I can't get a handle on a Manager.SessionDelegate object for the init.  I can't get the SessionDelegate off the default AlamofireManager - and can't pass one in because the class isn't public. 

Should your updated API allow for passing in an NSURLSessionDelegate instead of a Manager.SessionDelegate?  I might be missing something - Thanks for any clarification...

cnoon:
I don't understand why the second initializer doesn't do exactly what you need @dgdosen.

```swift
    public init?(
        session: NSURLSession,
        delegate: SessionDelegate,
        serverTrustPolicyManager: ServerTrustPolicyManager? = nil)
    {
```

Just create a session delegate, then create your custom session, and pass both in. You have references to both the delegate and the session. Win win.

PR #810, opened 2015-09-25
Updated README.md

dfmuir:
In Xcode 7, the base64EncodedStringWithOptions method no longer compiles with a nil argument. The comments associated with that method suggest: "Use zero or one of the following...". This pull request updates the documentation to pass a NSDataBase64EncodingOptions value of 0 to this method. I would love to see something that doesn't involve using rawValue, but this seems to be the best option at this time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ let user = "user"
 let password = "password"
 
 let credentialData = "\(user):\(password)".dataUsingEncoding(NSUTF8StringEncoding)!
-let base64Credentials = credentialData.base64EncodedStringWithOptions(nil)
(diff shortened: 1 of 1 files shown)


jshier:
@dfmuir Given that it's an `OptionSetType` under Swift 2, the proper way to set no options is to use `[]`, I believe.

dfmuir:
@jshier Thank you for pointing this out. I have added a new commit to the pull request. I think this is the cleanest way to do it.

cnoon:
Thanks for your PR @dfmuir! I merged a slightly modified version of your changes into `master` in 13925497.

Cheers 🍻

PR #812, opened 2015-09-25
Import MobileCoreServices for tvOS

petereddy:
Importing MobileCoreServices for os=tvOS allows Alamofire to compile on tvOS


cnoon:
You should be using the `tvOS` branch. We won't be supporting tvOS in the `master` branch once Xcode 7.1 is out of beta.

PR #827, opened 2015-09-30
Update README.md

Screon:
It isn't need to check for document path in Swift 2.0


cnoon:
Thanks for your PR @Screon! I merged a slightly modified version of your PR in 401e76a9. Thanks again!

PR #843, opened 2015-10-06
Updated description and instructions for Carthage

Ashton-W:
I felt the description and instructions for *Carthage* weren't accurate.

- reworded the introduction to *Carthage* to be closer to how it describes itself.
- added project integration step instruction for *Carthage* based on its documentation.


cnoon:
Thanks @Ashton-W!

PR #861, opened 2015-10-14
Making boundary changeable

FrankJaeger:
I've got a project where I need be able to change a boundary. Is there a reason to prohibit that ?


cnoon:
The reason it is not exposed is that we didn't think there should ever be a need to specify your own value. The current implementation follows the RFC specs which are a bit vague. Could you shed some more light on what the server is doing so we can better understand the use case? Opening this up is something I'd rather avoid unless there is a valid reason to do so.

cnoon:
Hi @FrankJaeger, I'm going to close this PR for now. If you end up posting a valid case for making this change, I may re-open and merge. Cheers.

PR #885, opened 2015-10-24
replace unnecessary usage of NSString with String

Nirma:
Sorry if this is too much a of nitpick, but I noticed this section while I was reading the project source and thought it would be better to use the string class directly instead. 

Hope it helps!

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -518,7 +518,7 @@ extension Request: CustomDebugStringConvertible {
 
         if let
             HTTPBodyData = request.HTTPBody,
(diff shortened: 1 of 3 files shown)


cnoon:
It's never too much of a nitpick @Nirma! Always always always submit a PR or open an issue with a question if you notice anything that could be improved. Thank you for doing so! 

I'm fairly certain these have not already been changed because this didn't used to work in Swift 1.2. Somewhere along the Swift 2.0 or 2.1 releases the Foundation extensions on String from NSString were implemented properly. Either way, this is a great change. Thanks for putting it together!

Cheers 🍻

PR #889, opened 2015-10-25
Feature - Added 204 Support to Response Serializers

cnoon:
We've been tracking this closely in our Trello project. Alamofire could greatly benefit from better 204 support. Therefore, we've been discussing the options in how to actually handle this. This PR puts forward the most sensible, non-invasive support for this behavior IMO.

The only other thing I question about this approach is whether we should require validation on the 204 as an acceptable status code before allowing this check in the response serializers. That would better help identify intent. I also hate to require that though as I feel it is a bit heavy handed.

I'm very interested to get everyone else's thoughts on this change. Anyone out there that has had to work through a 204 use case, please get your feedback in now!

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -261,8 +261,8 @@ extension Request {
 
             guard response?.statusCode ?? -1 != 204 else { return .Success(NSNull()) }
 
-            guard let validData = data else {
-                let failureReason = "JSON could not be serialized. Input data was nil."
+            guard let validData = data where validData.length > 0 else {
+                let failureReason = "JSON could not be serialized. Input data was nil or zero length."
                 let error = Error.errorWithCode(.JSONSerializationFailed, failureReason: failureReason)
                 return .Failure(error)
             }
@@ -317,8 +317,8 @@ extension Request {
 
             guard response?.statusCode ?? -1 != 204 else { return .Success(NSNull()) }
 
-            guard let validData = data else {
-                let failureReason = "Property list could not be serialized. Input data was nil."
+            guard let validData = data where validData.length > 0 else {
+                let failureReason = "Property list could not be serialized. Input data was nil or zero length."
                 let error = Error.errorWithCode(.PropertyListSerializationFailed, failureReason: failureReason)
                 return .Failure(error)
             }
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -145,11 +145,13 @@ extension Request {
         - returns: A data response serializer.
     */
     public static func dataResponseSerializer() -> ResponseSerializer<NSData, NSError> {
-        return ResponseSerializer { _, _, data, error in
+        return ResponseSerializer { _, response, data, error in
             guard error == nil else { return .Failure(error!) }
 
-            guard let validData = data where validData.length > 0 else {
-                let failureReason = "Data could not be serialized. Input data was nil or zero length."
+            guard response?.statusCode ?? -1 != 204 else { return .Success(NSData()) }
+
+            guard let validData = data else {
+                let failureReason = "Data could not be serialized. Input data was nil."
                 let error = Error.errorWithCode(.DataSerializationFailed, failureReason: failureReason)
                 return .Failure(error)
             }
@@ -190,8 +192,10 @@ extension Request {
         return ResponseSerializer { _, response, data, error in
             guard error == nil else { return .Failure(error!) }
 
-            guard let validData = data where validData.length > 0 else {
-                let failureReason = "String could not be serialized. Input data was nil or zero length."
+            guard response?.statusCode ?? -1 != 204 else { return .Success("") }
+
+            guard let validData = data else {
+                let failureReason = "String could not be serialized. Input data was nil."
                 let error = Error.errorWithCode(.StringSerializationFailed, failureReason: failureReason)
                 return .Failure(error)
             }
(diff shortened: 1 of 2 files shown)


cnoon:
cc @jshier, @kcharwood, @kylef

Tests/ResponseSerializationTests.swift
```diff
@@ -266,8 +348,8 @@ class ResponseSerializationTestCase: BaseTestCase {
         XCTAssertNotNil(result.error, "result error should not be nil")
 
         if let error = result.error {
-            XCTAssertEqual(error.domain, Error.Domain, "error domain should match expected value")
-            XCTAssertEqual(error.code, Error.Code.JSONSerializationFailed.rawValue, "error code should match expected value")
+            XCTAssertEqual(error.domain, "NSCocoaErrorDomain", "error domain should match expected value")
```
kylef: Can't we use the constant for `NSCocoaErrorDomain`?

cnoon: We definitely can! Will fix.

cnoon: Addressed in 304af310.

Tests/ResponseSerializationTests.swift
```diff
@@ -266,8 +348,8 @@ class ResponseSerializationTestCase: BaseTestCase {
         XCTAssertNotNil(result.error, "result error should not be nil")
 
         if let error = result.error {
-            XCTAssertEqual(error.domain, Error.Domain, "error domain should match expected value")
-            XCTAssertEqual(error.code, Error.Code.JSONSerializationFailed.rawValue, "error code should match expected value")
+            XCTAssertEqual(error.domain, NSCocoaErrorDomain, "error domain should match expected value")
+            XCTAssertEqual(error.code, 3840, "error code should match expected value")
```
kylef: Isn't this `NSPropertyListReadCorruptError`? Seems like a Foundation bug raising a property list error for JSON.

cnoon: Yeah it definitely is an `NSPropertyListReadCorruptError`. Looks like they used the same errors for JSON as they did plists rather than redefining special error codes for JSON. With that said, should we keep the `data.length > 0` check to instead through an actual Alamofire JSON error instead? I think it would make sense in both JSON and plist cases to keep that check. Thoughts?

Source/ResponseSerialization.swift
```diff
@@ -145,11 +145,13 @@ extension Request {
         - returns: A data response serializer.
     */
     public static func dataResponseSerializer() -> ResponseSerializer<NSData, NSError> {
-        return ResponseSerializer { _, _, data, error in
+        return ResponseSerializer { _, response, data, error in
             guard error == nil else { return .Failure(error!) }
 
-            guard let validData = data where validData.length > 0 else {
-                let failureReason = "Data could not be serialized. Input data was nil or zero length."
+            guard response?.statusCode ?? -1 != 204 else { return .Success(NSData()) }
```
jshier: How do you feel about:
```swift
guard let response = response where response.statusCode != 204 else { return .Success(NSData()) }
```
It makes more sense to me than coalescing into a -1.

cnoon: Yup...make way more sense to me too! 😉🍻 Will change.

cnoon: Yeah now I remember why I didn't do this...if response is nil then the `guard` will fail and it will return a `.Success` case which is definitely not what we want. I don't see an cleaner way to do it than I already have. Open to suggestions...

Tests/ResponseSerializationTests.swift
cnoon: Been thinking about this and I think it would be much better to throw an actual Alamofire error instead of the 3840 which can be misleading. Updated in 83b32faa.

aryaxt:
This is great :+1:

damienrambout:
This is great :) I just don't like the `NSData()` and `NSNull()` that much. I guess it can't be done anyway else because `.Success` is expecting a non-nil value. And I'm pretty sure you wouldn't be a big fan of `ResponseSerializer<AnyObject?, NSError>` returned as the JSON serializer.
Anyway, it 's way better than getting an error. `NSNull` can be checked by the receiver at runtime and transformed to `nil` when needed/appropriate.

Source/ResponseSerialization.swift
cristeahub: I think writing this with parenthesis like this:
```swift
guard (response?.statusCode ?? -1) != 204 else { return .Success(NSData()) }
```
might look better. At a glace I got confused with the precedence of `??` and `!=`.

jshier: If you aren't opposed to using an if, this will do what you want:
```swift
if let statusCode = response?.statusCode where statusCode == 204 { return .Success(NSData()) }
```

cnoon:
That's exactly right @damienrambout. These changes have a big set of tradeoffs. I'm not a huge fan of having to return the `NSData()` and `NSNull()` objects either, but that's the only option without making MAJOR changes to the other underlying types.

Source/ResponseSerialization.swift
cnoon: Yeah that's fair @jshier. Will change.

cnoon: Addressed in e5dfdb1a.

RafaelPlantard:
@cnoon How I can do a request when I wait for 204 response? Can you show me a sample?

PR #890, opened 2015-10-25
Feature - Content Type Validation on Empty Data

cnoon:
Content Type validation now always succeeds when server data is nil or empty. Content type validation should NEVER fail when there is no data to actually validate against. For example, if someone specifies they want to use `application/json` validation for all their JSON requests, this check would always fail on DELETE requests that most likely will return a 204. This change ensures this common use case is better supported.

### DIFF
--- a/Tests/ValidationTests.swift
+++ b/Tests/ValidationTests.swift
@@ -220,12 +220,6 @@ class ContentTypeValidationTestCase: BaseTestCase {
             .validate(contentType: [])
             .response { _, response, data, responseError in
                 error = responseError
-
-                print(response)
-                if let data = data {
-                    print(String(data: data, encoding: NSUTF8StringEncoding)!)
-                }
-
                 expectation.fulfill()
             }
 
--- a/Tests/ValidationTests.swift
+++ b/Tests/ValidationTests.swift
@@ -208,6 +208,33 @@ class ContentTypeValidationTestCase: BaseTestCase {
         }
     }
 
+    func testThatValidationForRequestWithNoAcceptableContentTypeResponseSucceedsWhenNoDataIsReturned() {
+        // Given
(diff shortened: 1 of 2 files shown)


cnoon:
cc @jshier, @kylef and @kcharwood

Tests/ValidationTests.swift
```diff
@@ -208,6 +208,33 @@ class ContentTypeValidationTestCase: BaseTestCase {
...
+
+        var error: NSError?
+
+        // When
+        Alamofire.request(.GET, URLString)
+            .validate(contentType: [])
+            .response { _, response, data, responseError in
+                error = responseError
+
+                print(response)
```
kylef: Was these left in from debugging?

cnoon: Grrr...yes, my bad. Will remove.

cnoon: Addressed in 12c97b76.

jshier:
Just wanted to point out that, like some backends require a `Content-Type` even when there's no content (I'm developing against one right now, :sob:), some users may be using this validation not as an actual check against the data returned but merely to ensure that the `Content-Type` header is returned with a particular value. This isn't best practice, of course, but it may be something to call out, especially for users updating from an older version of Alamofire. Perhaps calling it out in the Validation section of the README would be a good idea?

cnoon:
I think that's a good point @jshier, but that seems like a very uncommon case to me. Most users would use the `validation(contentType:)` method to ensure the data returned the correct type. Doesn't make sense to me why the client would want to validate the content type header and not care about the actual data returned. That seems more like a set of tests that should live at the server level.

cnoon:
We could certainly add some documentation there, but I think it would just confuse people since it seems like such a strange way to me to perform validation as a client.

PR #898, opened 2015-10-30
Update deprecated playground function

paulthorsteinson:
```XCPSetExecutionShouldContinueIndefinitely()``` should now be ```XCPlaygroundPage.currentPage.needsIndefiniteExecution = true``` as the former was deprecated.


cnoon:
Thanks for the PR @paulthorsteinson. Rather than merging this in, I removed the playground completely in 7582758d due to instability reasons. Thanks again.

PR #903, opened 2015-11-01
Don't code sign for iOS simulator SDK builds

JRHeaton:
Hi there. I've made a small change to the project settings to not code sign the iOS framework target when built against the simulator SDK. The reason for this is because when built with carthage on a CI server with no code signing identities, it defaults to building with the simulator SDK which doesn't require code signing like the iphoneos SDK does. Other frameworks supporting Carthage installs have been doing this as well it seems (example: https://github.com/ReactiveCocoa/ReactiveCocoa/commit/f685971d32d9035df40eb0a7dd49f2224e8cab1b)


cnoon:
Thanks for this @JRHeaton! I merged a slightly modified version of your changes in cbac897b and also put together 2159ce3c for `watchOS` and `tvOS`. Both are available in `master`.

PR #904, opened 2015-11-02
Added robustness to TLSEvaluationTests in timeout case

cotsog:
You need to declare your expectation variables as "weak" variables so that the function does not keep a reference which was causing your tests to hang.

You can see this StackOverflow question for more details: http://stackoverflow.com/questions/27555499/xctestexpectation-how-to-avoid-calling-the-fulfill-method-after-the-wait-contex

### DIFF
--- a/Alamofire.xcodeproj/xcshareddata/xcschemes/Alamofire iOS.xcscheme
+++ b/Alamofire.xcodeproj/xcshareddata/xcschemes/Alamofire iOS.xcscheme
@@ -52,11 +52,6 @@
                BlueprintName = "Alamofire iOS Tests"
                ReferencedContainer = "container:Alamofire.xcodeproj">
             </BuildableReference>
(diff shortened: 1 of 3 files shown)


cnoon:
Thanks for putting this together @cotsog! This has been a really tricky problem for us to track down. I can see how this solution would in fact fix the problem. The one caveat that I would like to point out is that this solution is NOT thread-safe. Anywhere where this solution is used, we need to make sure the async callback is executing on the main thread. Otherwise, we can run into a race condition where the callback is trying to call the expectation while the time is expiring.

cnoon:
Okay so I ended up unwinding the merge and squashing the commits into e4320bd5. Thanks again for putting this together!

cotsog:
No problem @cnoon! Don't hesitate to reach out again if you need help.

PR #912, opened 2015-11-10
Replace http to https on README

tmnb:
Sample code does not work, because of ATS in iOS9. I made a change in README as follows; replace http with https.


kylef:
Thanks @tmnb

PR #913, opened 2015-11-10
Sets UIRequiredDeviceCapabilities for tvOS target

simonbs:
arm64 must be set in UIRequiredDeviceCapabilities for tvOS targets when uploading to iTunes Connect.


cnoon:
Thanks for this @simonbs!

PR #949, opened 2015-12-05
change escape method

sun-fox-cj:
change escape method, cuz there are some custom chars need to check


cnoon:
Thank you for your PR @sun-fox-cj!

Unfortunately, we think this is probably not the best change for the community. The `escape` method is intended only for escaping query parameter key value pairs. It is not for escaping other parts of a URL. If you find the need to customize the escaped characters, then you will need to use your own custom implementation. We have MANY tests around this functionality to ensure we're following the spec as closely as possible for everyone.

Thanks again. 🍻

PR #953, opened 2015-12-08
Disable testability for release config and disable tests for non-test builds.

jedlewison:
Currently, the ENABLE_TESTABILITY flag is enabled for release builds. This appears to have been turned on to support building with Carthage, but setting up the projects schemes so that they don't build tests except when tests are being run achieves the same goal without incurring the performance impact that enabling testability can have. This PR leaves the ENABLE_TESTABILITY flag in the default state -- on for debug, off for release.

It's possible there's some other consideration of which I'm unaware, but at least with respect to Carthage builds, this configuration works equally well without the downside of always having ENABLE_TESTABILITY on.


jedlewison:
Adding -- These build settings are the current default for new projects in Xcode.

cnoon:
Hi @jedlewison, thanks for putting this together! Unfortunately, we won't be merging these changes in. The project was definitely configured this way by design. We want to support Carthage as well as always have the project compile the test code with any code changes. It helps ensure we never introduce compiler issues in the test suite when fixing bugs or developing new features. I understand that this introduces a slight overhead when using Carthage to compile Alamofire, but IMO I'd always error on the side of stability here rather than Carthage performance.

Thanks again. 🍻

jedlewison:
I'm curious -- how does this impact the ability to run Alamofire tests? I ran the projects test suite before and after with the exact same results. Essentially all this would do is use the standard Xcode project defaults for a new project with unit tests.

Perhaps my description of the pull request set off alarm bells that it shouldn't have? Either that, or there is an issue with running tests of which I'm not aware.

jedlewison:
(Adding: Thanks for taking a look at the PR and for the work you do on Alamofire.)

jedlewison:
@cnoon Adding one more thing -- the project currently works fine with Carthage. The motivation here is to disable testability in release mode because it negatively impacts performance.

Here's a description (from [Apple](https://developer.apple.com/library/ios/releasenotes/DeveloperTools/RN-Xcode/Chapters/xc7_release_notes.html)) describing the issue (emphasis added):

> Testability: Tests of Swift 2.0 frameworks and apps are written without having to make internal routines public.
> Use @testable import {ModuleName} in your test source code to make all public and internal routines usable. The app or framework target needs to be compiled with the Enable Testability build setting set to Yes. ***The Enable Testability build setting should be used only in your Debug configuration, because it prohibits optimizations that depend on not exporting internal symbols from the app or framework.*** (17732115)

cnoon:
Okay, thanks for all this info @jedlewison. I'm going to re-open this and investigate further. I wasn't aware that flag prohibiting the compiler optimizations. It makes sense now that I'm thinking about it that way.

Thanks for pressing on us here. 👍🏼

jedlewison:
You bet, @cnoon  -- thanks for taking another look.

cnoon:
Okay @jedlewison, after digging in further, I think your change is overall the best course of action. The reason I've been so hesitant on this change is that it requires us to disable the Travis-CI builds for the release configuration. Apple doesn't allow you to have an optimized build with the `ENABLE_TESTABILITY` flag disabled and still use `@testable` in your test target.

While that's certainly a bummer, it's not a blocker. I ended up finding another workaround for this in 39ca953d. What I did was create a `ReleaseTest` configuration that is configured exactly like the `Release` configuration, except has the `ENABLE_TESTABILITY` flag enabled. Since Carthage defaults to the `Release` configuration, this should be a sufficient way to attack the problem. This allows us to run our test suite against the `ReleaseTest` configuration on Travis-CI, while still allowing all our Carthage users to optimize their Alamofire frameworks.

Thanks again for your due diligence here. Alamofire is certainly much better off for it!

Cheers. 🍻

jedlewison:
@cnoon Thanks so much for not only considering this PR, but also taking the time and effort to to deal with the side effects. I'm happy to have helped in a small way, but the bigger thanks go to you and the rest of the Alamofire crew! We use it extensively in our Swift code at Best Buy and it works incredibly well.

cnoon:
That's awesome! Glad to hear it. 👍🏼

PR #954, opened 2015-12-08
'URL' ParameterEncoding handles empty parameter dictionary

AnthonyMDev:
When passing in an empty dictionary for `parameters`:

*Current:*

    let query = request.URL?.query    // "" (empty string)
    let URLString = request.URL?.absoluteString    // "http://www.example.com/&" (URL has "&" appended)

*This PR:*

    let query = request.URL?.query    // nil
    let URLString = request.URL?.absoluteString    // "http://www.example.com"

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -90,7 +90,11 @@ public enum ParameterEncoding {
 
         switch self {
         case .URL, .URLEncodedInURL:
-            func query(parameters: [String: AnyObject]) -> String {
+            func query(parameters: [String: AnyObject]) -> String? {
+                guard parameters.count > 0 else {
+                    return nil
+                }
+                
                 var components: [(String, String)] = []
 
                 for key in parameters.keys.sort(<) {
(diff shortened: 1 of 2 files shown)


cnoon:
Hi @AnthonyMDev, 

Thanks for putting this together...much appreciated! Overall I really like the goal of this change that you broke down nicely in your description. However, I think there's a MUCH easier approach. What about the following?

```swift
public func encode(
    URLRequest: URLRequestConvertible,
    parameters: [String: AnyObject]?)
    -> (NSMutableURLRequest, NSError?)
{
    var mutableURLRequest = URLRequest.URLRequest

    guard let parameters = parameters where !parameters.isEmpty else { return (mutableURLRequest, nil) }

    var encodingError: NSError? = nil

    // etc. etc.
}          
```

That's probably what this method should have been doing all along. If your down with this approach, do you think you could rework this PR to handle the change this way while keeping the test as part of the PR?

AnthonyMDev:
@cnoon Yeah, I'll do that at some point today for you.
I didn't want to touch anything affecting any of the other encoding methods, as I hadn't tested how they are handling the scenario. When making changes to a library I don't own, I traditionally like to make my changes as conservative as possible, but if you are good with changing this for all encoding methods, I'll happily make the change.

AnthonyMDev:
@cnoon Refactor completed.

cnoon:
Thanks for putting this together @AnthonyMDev...much appreciated! I also pushed b0c6a9bf to fix up a failing custom parameter encoding test case. Normally I wouldn't modify a test like this so a change that did break the test suite could move though. However, when looking at the failing test, it wasn't really a good test. Therefore, I updated it.

Thanks again!

PR #955, opened 2015-12-09
Update README.md

vlastachu:
Sorry for such little fix. I'm not sure that this is ok.
Search problem takes about five minutes, but it is much easier to see commands that just doing the right thing.


cnoon:
I like it!

PR #956, opened 2015-12-11
Update certificatesInBundle - addl. extensions, case sensitivity

jacobjennings:
In certificatesInBundle, check for well-formed certificates with
extensiions .cer, .crt, .der, and check for upper and lower case
extensions (OSX filesystem may or may not be case sensitive. iOS
always is)

DER is a binary encoding format
PEM is an ASCII encoding format
.crt is the most common certificate extension, may be DER or PEM
.cer is a Microsoft extension convention, may be DER or PEM
.der is a (poor) extension referring to encoding instead of
contents, these may contain a certificate and/or a key.

Apple's SecCertificateCreateWithData only accepts DER-encoded certificates


cnoon:
Thanks for putting this together @jacobjennings...much appreciated! I made a few small formatting changes to your PR and pushed it up in 234568d4 still giving you attribution.

Thanks again! 🍻

PR #958, opened 2015-12-14
Add optional queue: argument to responseJSON

nvh:
This makes it possible to specify a dispatch queue when using responseJSON.

Also it's not possible to get this behaviour with the current version of Alamofire, because the property `response` from `Request` hides the function 
```swift
func response<T: ResponseSerializerType>(
        queue queue: dispatch_queue_t? = nil,
        responseSerializer: T,
        completionHandler: Response<T.SerializedObject, T.ErrorObject> -> Void)
        -> Self
```

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -285,11 +285,13 @@ extension Request {
         - returns: The request.
     */
     public func responseJSON(
-        options options: NSJSONReadingOptions = .AllowFragments,
+        queue queue: dispatch_queue_t? = nil,
+        options: NSJSONReadingOptions = .AllowFragments,
         completionHandler: Response<AnyObject, NSError> -> Void)
         -> Self
     {
         return response(
+            queue: queue,
             responseSerializer: Request.JSONResponseSerializer(options: options),
             completionHandler: completionHandler
         )


cnoon:
Thanks for the PR @nvh. Unfortunately, I don't think this change is the correct way to go with the `responseJSON` serializer. The statement you made about the `response` method being hidden doesn't seem to be true. The following compiles without any issues on the `master` branch:

```swift
Alamofire.request(.GET, "https://httpbin.org/get")
    .response(
        queue: dispatch_get_main_queue(),
        responseSerializer: Request.JSONResponseSerializer(),
        completionHandler: { response in
            print(response)
        }
)
```

If you need to use a custom queue, this is the recommended way to do it. This is also a great example of why the `ResponseSerializerType` protocol exists making it easy to swap serialization logic.

Thanks again! 🍻

nvh:
I tried to put it in an extension:
```
extension Alamofire.Request {
    public func responseJSON(
        queue queue: dispatch_queue_t? = nil,
        options: NSJSONReadingOptions = .AllowFragments,
        completionHandler: Response<AnyObject, NSError> -> Void)
        -> Self
    {
        return response(
            queue: queue,
            responseSerializer: Request.JSONResponseSerializer(options: options),
            completionHandler: completionHandler
        )
    }
}
```
But that didn't work:
screen shot 2015-12-16 at 11 43 49
Hence my PR

Turns out, I had a `Request` struct in the project that was hiding Alamofire's `Request` class. After prefixing `Request` in the function with `Alamofire.` it works. So it was just a unhelpful error message from Xcode. Thanks!

PR #959, opened 2015-12-14
Add AlamofireObjectMapper to README

phatmann:
Add info about AlamofireObjectMapper to new Third-Party Libraries section


cnoon:
Hi @phatmann...thanks for putting this together. Unfortunately, that section of the README is only for promoting libraries officially supported by the [AlamofireSF](https://github.com/Alamofire/Foundation). We want to avoid promoting libraries in which we don't have any control over.

Thanks anyways! 🍻

Issue #965, opened 2015-12-17
Feature request: customise the bahaviour for parameter array


demetrio812:
Hello,
currently if I use an array for a parameter Alamofire URL method encodes it with square bracket like:

    &foo[]=bar&foo[]=bar2

I understand the meaning and it is also specified that:

    Since there is no published specification for how to encode collection types, the convention of appending `[]` to the key for array values (`foo[]=1&foo[]=2`), and appending the key surrounded by square brackets for nested dictionary values (`foo[bar]=baz`)

However, even some modern servers that uses RestEasy with JAX-RS accept this format for arrays:

    &foo=bar&foo=bar2

For this reason, I'd like to propose to have the ability to customise this behaviour with an option in order to support both cases.

It's a simple change that I can help doing and I think it will add flexibility to the library and will avoid the usage of the Custom method for this simple case.

What do you think?

Thanks,
Dem

DavdRoman:
+1. Exactly the issue I'm dealing with currently. I'm pretty sure this would be pretty straightforward to implement.

cnoon:
I'm going to go ahead and close this issue out. Please see my comments in #966. Cheers. 🍻

demetrio812:
Needless to say I don't agree that it's not commonly used, maybe in your experience it's not but I'm using a commonly used modern java framework so it's not that unusual. Also, beside the pull proposal from DavidRoman, it can be implemented in a backward compatible way (default config is like it is now, and you can change if you want) and you will support both commonly used use cases.

Obviously you (the team) decide what you want to do, but saying "it's not commonly used" is not an acceptable answer in my opinion.

Same thing for multiple encoding methods in a single request (https://github.com/Alamofire/Alamofire/issues/374): you (the team) say it's not common but again it's not common for you, it's a very common pattern that happens regularly and it feels like Alamofire doesn't support basic patterns out of the box.

Cheers,
Dem

cnoon:
Just to be clear, at no point did I claim that the PR was declined because "it's not commonly used". I declined it for other reasons which were very cleared documented. Additionally, I think you are assuming a bit too much here. At no point have we claimed we'll never support this feature. It's actually in our Trello backlog and has been for a while. I think I could have done a better job explaining that in my previous comment to avoid confusion...my bad there.

[...]

Issue #329, opened 2015-02-03
Get request and array params


solomon23:
It doesn't seem like GET requests are translating array params correctly.

        let params: [String: AnyObject] = ["q": ["one", "two"], "target": "en"]
        Alamofire.request(.GET, "https://www.googleapis.com/language/translate/v2", parameters: params)

generates the url:

q%5B%5D=one&q%5B%5D=two&target=en

I think it's trying to added []'s for the POST case.  But in the GET case it should be q=one&q=two

JRG-Developer:
This is intended behavior. Per the README.md,

>Since there is no published specification for how to encode collection types, the convention of appending [] to the key for array values (foo[]=1&foo[]=2), and appending the key surrounded by square brackets for nested dictionary values (foo[bar]=baz) [is used].

Instead of using GET, you should consider using POST or PUT and passing values via JSON, XML, or another well-defined format. This could require server side changes obviously.

If server side is out of your control, you should consider manually encoding these parameters instead.

solomon23:
Yea it's the google api endpoint, so it's out of my control.  The google API doesn't like that format.  I can construct it manually, thought I'd file it though so you guys are aware.

mattt:
@JRG-Developer is correct—this is intended behavior. Use custom parameter encoding if you need anything other than the provided options.

PR #966, opened 2015-12-17
Merge URL and URLEncodedInURL into URL(URLEncodedInURL, squareBrackets)

DavdRoman:
In the intent of addressing #965, I saw an opportunity to merge URL and URLEncodedInURL to simplify the encoding process even more and give some extra facilities such as the [so](#251) [requested](#329) "square bracket-less" array parameters support.


cnoon:
Thank you for putting this together @DavdRoman. We always appreciate updates from the community. 

With that said, I'm afraid we won't be adopting these changes into Alamofire. The main reason is complexity. There is no official RFC spec for any of the collection formatting. Our official Alamofire position is that we support one of the most commonly used formats when it comes to URL encoded collection formatting. If you need to customize the format in a different way, then you need to use a `.Custom` parameter encoding.

The main reason for this is to make it easy for the community to adopt, yet flexible enough to handle any use case. This is our philosophy when designing any of the Alamofire APIs. These changes make things more complicated without a high enough ROI to consider it a worthwhile change. Additionally, these changes are not in any way backwards compatible and would require a major version bump.

Thanks again. 🍻

DavdRoman:
Thanks @cnoon, I totally get your reasoning. I just decided to fork Alamofire to adjust it to my current necessities and sent this pull requests as a suggestion for future versions.

Thanks for your answer 🍻

sharoni:
>There is no official RFC spec for any of the collection formatting

Brackets are illegal characters for parameter names (regardless if it is a collection or not) according to https://tools.ietf.org/html/rfc6570#section-2.3

PR #979, opened 2015-12-25
Add support for Swift Package Manager (SPM)

kylef:
There are various warnings when building Alamofire with SPM, however it does build fine. I'd suggest we add initial support using this PR and then work on the warnings closer to the next Swift release that goes into Xcode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,8 @@ DerivedData
 *.ipa
 *.xcuserstate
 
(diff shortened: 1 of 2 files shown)


cnoon:
Awesome! Thanks for putting this together @kylef. I agree, let's get it in now and continue to iterate closer to the next Swift release. 👍🏼

PR #981, opened 2015-12-29
Background session enhancements

mwoollard:
I like much of Alamofire but with background sessions there isn't any build it features to simplify re-connecting to background session tasks after an app restart. I've in the past used a serialisable delegate class that methods for the various delegates involved that was passed as part of a request. This would be serialised to the file system for requests associated with background session. These would automatically be reloaded on app restart and the idea being they would contain all the information necessary to continue tracking the task. This doesn't however map that well to Alamofire's design. So instead what I've done is add a dictionary to Request that will be serialised if Request is associated with background session. This means the data can be read after an app restart. I've also made it so that Request get re-created when referenced by background session task after an app restart. Add to this two methods to allow Requests created this way to be re-connected to delegate closures and managing background sessions is simpler. This all could have been done with code in the client app using existing Alamofire framework but better to have some support built in I think. Check out the example app additions for how I envisage it could be used. Go into background downloads, click the + a few times, exit page and then go back - this re-connects to new VC - terminate app by terminating session under debugger - restart and it all reconnects if you go back in the page.

This is just presented as a concept and am sure needs further discussion and adjustment but let me know what you think and whether you feel its an approach that makes sense for you. Happy to continue in this direction to incorporate your feedback if you wish.

All the best
Mark

### DIFF
--- a/Example/AppDelegate.swift
+++ b/Example/AppDelegate.swift
@@ -21,11 +21,15 @@
 // THE SOFTWARE.
 
 import UIKit
+import Alamofire
+
+public let BackgroundSessionId = "com.alamofire.example"
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate, UISplitViewControllerDelegate {
 
     var window: UIWindow?
+    var manager: Manager?
 
     // MARK: - UIApplicationDelegate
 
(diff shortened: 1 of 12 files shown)


cnoon:
Thanks for putting this together @mwoollard! This is going to take me a while to dig through everything you have put together and wrap my head around everything. Please be patient with us while we evaluate this PR. Thanks!

mwoollard:
No problem - any questions let me know !

cnoon:
I haven't forgotten about this @mwoollard. Hope to get to it soon!

cnoon:
Okay @mwoollard, I finally managed to make it through your PR. Unfortunately, I don't think these changes or this approach should be adopted into Alamofire. These changes are quite involved and don't solve the core issue which is that all the closure logic is still being dropped.

When we add better support to Alamofire for background sessions, we have to address the fact that all the public APIs currently only serve `default` and `ephemeral` configurations. Closures simply don't work with background sessions. Just FYI, we've slated better background session support for Alamofire 4 in our Trello backlog.

Thanks again for all your work here. All the best. 🍻

mwoollard:
No problem - not unsurprising as big change as you say - look forward to v4 - happy to contribute / give feedback as framework moves forward. All the best Mark

PR #982, opened 2015-12-31
Use functional approach to generate acceptLanguage

dapenggao:
Use `map` instead of `for` loop and mutation. `prefix(5)` is equivalent to `if q <= 0.5 { break }` in the loop.


cnoon:
Thanks for putting this together @dapenggao! I merged a slightly modified version of your changes in fa5a10d2 while still giving you attribution. You need a `prefix(6)` to maintain the same logic. I also pulled the `quality` into a separate local variable to make it a bit easier to understand.

Thanks again! 🍻

dapenggao:
Thanks a lot!

PR #983, opened 2015-12-31
Fix the crash problem in debugDescription

tokorom:
I hope that debugDescription of Request with an invalid URL do not crash.

I added the guard statement so as not to crash even when the URL was invalid.


cnoon:
Thanks for putting this together @tokorom...much appreciated! I merged a slightly modified variation of your changes in 08cf2695 while still giving you attribution. Thanks again for putting this together! 🍻

PR #984, opened 2016-01-01
Support the Custom encoding without parameters

tokorom:
I need to use `.Custom` encoding without parameters.

This is a specific case in Dropbox's iOS SDK.

https://github.com/dropbox/SwiftyDropbox/blob/master/Source/Client.swift#L177-L181

```swift
// params: JSON
let request = client.manager.request(.POST, url, parameters: [:], headers: headers, encoding: ParameterEncoding.Custom {(convertible, _) in
    let mutableRequest = convertible.URLRequest.copy() as! NSMutableURLRequest
    mutableRequest.HTTPBody = dumpJSON(params)
    return (mutableRequest, nil)
})
```

In this case, parameters is not required but, the Custom encoding closure is not called.

I hope to fix it so that it does not check the parameters for only case of `.Custom`.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -82,8 +82,15 @@ public enum ParameterEncoding {
     {
         var mutableURLRequest = URLRequest.URLRequest
 
-        guard let parameters = parameters where !parameters.isEmpty else {
-            return (mutableURLRequest, nil)
+        let validParameters: [String: AnyObject]
+
+        if case .Custom = self {
+            validParameters = parameters ?? [:]
+        } else {
+            guard let parameters = parameters where !parameters.isEmpty else {
+                return (mutableURLRequest, nil)
+            }
+            validParameters = parameters
         }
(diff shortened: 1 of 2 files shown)


cnoon:
Hi @tokorom,

This behavior is intended. The `.Custom` parameter encoding case should only be used for creating a custom parameter encoding for the URL request, nothing else. It doesn't logically make sense to be customizing all different parts of the URL request inside of a parameter encoding method. It should only customize the parameters.

> Please see #954 for more info about the change.

Instead, you should create your own URL request manually and customize as needed. You can still use the `ParameterEncoding` encode method, then customize the returned mutable URL request as needed. Once you are finished, you can pass it to the `Manager.request` method to execute the request.

Hopefully that all makes sense.

Cheers. 🍻

tokorom:
Thanks for your polite reply :bow:
I understood the behavior of `.Custom` and I was fixed SwiftyDropbox.

PR #993, opened 2016-01-05
experimental implementation of Request.retry(count)

itchingpixels:
Hi!

This is a silly pull request, please, feel free to close it immediately if you find it intrusive or anything like that.
This contains a hacky implementation of .retry(count), that you can use to chain any Request on.

It doesn't include any tests and it shouldn't be merged.
It includes a modified Example-iOS project that demonstrates its usage.

I'd really appreciate if I could get some feedback on this.
I read that the reason why retry() is left out of this library is the extra clutter it would bring to the public API. I'm definitely not saying that this simplistic implementation (with basically zero configuration options) would do the trick, but it would have definitely helped me in certain scenarios. I also feel like the fluent interface of Request makes it an ideal candidate to choose between retry policies (including custom ones) in the future.

I am motivated here by the poor retry implementations I've seen in different apps I worked on. It would be awesome to give people a tool to manage it better!:) And I'm happy to contribute in the proper implementation of this feature.

Thanks!

### DIFF
--- a/Example/MasterViewController.swift
+++ b/Example/MasterViewController.swift
@@ -64,23 +64,23 @@ class MasterViewController: UITableViewController {
                 switch segue.identifier! {
                 case "GET":
                     detailViewController.segueIdentifier = "GET"
-                    return Alamofire.request(.GET, "https://httpbin.org/get")
+                    return Alamofire.request(.GET, "https://httpbin.org/get").retry(5)
                 case "POST":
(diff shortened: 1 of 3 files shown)


jshier:
@itchingpixels Thanks for this! However, we'll need to do a much more comprehensive investigation to support this feature. We are tracking automatic retry as a feature in our backlog, so it's possible we'll support this in the future.

PR #1008, opened 2016-01-14
Add @warn_unused_result attribute

samirGuerdah:
Add @warn_unused_result attribute to trigger a warning if the result of a function or method isn't used


cnoon:
Hi @samirGuerdah...thanks for the pull request! Unfortunately, these changes are too heavy handed and should not be merged. This attribute is only to be used in cases where you actually want to throw up a warning if the return value is not used. For example, we don't want to throw a warning on the `Alamofire.request()` method if you don't use the returned `Request`.

Thanks again. 🍻

PR #1013, opened 2016-01-16
Added a completionQueue to the manager class.

hwittland:
This change adds a completionQueue to the manager class.
It changes the initializer of the request class to get the manager object instead of an NSURLSession and uses the new completionQueue in responsehandling.

If set the completenQueue will be used to dispatch the completenHandler to in the response methods.

### DIFF
--- a/Source/Download.swift
+++ b/Source/Download.swift
@@ -42,7 +42,7 @@ extension Manager {
             }
         }
 
-        let request = Request(session: session, task: downloadTask)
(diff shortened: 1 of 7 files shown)


cnoon:
Thank you for the pull request @hwittland. Unfortunately, we don't think this is a good change to make to the core Alamofire logic. It already defaults to the main queue when a custom one is not provided. Additionally, the queue can easily be overridden already by calling the `response` method with a queue and response serializer.

Thanks again. 🍻

PR #1027, opened 2016-01-25
Fix Broken CocoaDocs Generation

### LINKED ISSUE #1025: Cocoadocs aren't generated properly
If you go to [Alamofire's Cocoadocs page](http://cocoadocs.org/docsets/Alamofire/3.1.5/), you'll see there's an error parsing the pod. I ran into the same problem with my project, [ForecastIO](https://github.com/sxg/ForecastIO). The issue is that there are two .xcodeproj files at the root of the repo ("Alamofire.xcodeproj" and "iOS Example.xcodeporoj"). In order for Cocoadocs to properly generate docs for Alamofire, "iOS Example.xcodeproj" needs to be moved out of the root.

jshier:
Move iOS Example project into Examples to fix CocoaDocs generation. Fixes #1025.


cnoon:
Hey @jshier, do you think you could take a look at the jazzy.yaml file approach instead? That approach would let us organize our repo however we want, yet still be able to properly generate jazzy docs.

cnoon:
Thanks for putting this together @jshier! Since I can't find any great documentation on Jazzy yaml files at the moment, let's roll with this solution for now. We can always change it later. Maybe we could get some help from some Jazzy experts out there?!?!

PR #1030, opened 2016-01-27
Fixed build warnings in Xcode 7.3 beta 2 for Swift 2.2

Baza207:
Nothing major changed here really, and mostly ran with the suggestions from Xcode.

I'm not sure if this needs to go on a `swift-2.2` branch for now, but I though I'd make this for now to try and help out.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -236,7 +236,7 @@ public enum ParameterEncoding {
             while index != string.endIndex {
                 let startIndex = index
                 let endIndex = index.advancedBy(batchSize, limit: string.endIndex)
-                let range = Range(start: startIndex, end: endIndex)
+                let range = startIndex ..< endIndex
 
                 let substring = string.substringWithRange(range)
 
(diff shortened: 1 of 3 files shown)


Source/ParameterEncoding.swift
```diff
@@ -236,7 +236,7 @@ public enum ParameterEncoding {
             while index != string.endIndex {
                 let startIndex = index
                 let endIndex = index.advancedBy(batchSize, limit: string.endIndex)
-                let range = Range(start: startIndex, end: endIndex)
+                let range = Range(startIndex ..< endIndex)
```
kylef: Wouldn't `startIndex ..< endIndex` return a range? Passing it down to the Range initialiser shouldn't be necessary

Baza207: True, I'll change, test and push an update.

cnoon:
Thanks for putting this together @Baza207! I cherry picked the commit into the new `swift_2.2` branch in 3cf3614c. We'll maintain this separate branch until Xcode 7.3 is officially released.

Thanks again! 🍻

PR #1053, opened 2016-02-06
Feature - Network Reachability Manager

cnoon:
This PR adds the `NetworkReachabilityManager` class to the project along with full coverage of the APIs. All public APIs have docstrings and I also added a section to the README with tips and tricks. The class is fully available on `iOS`, `OSX` and `tvOS` but not `watchOS`. The podspec has been updated to reflect the `watchOS` omission.

One important callout is that the default socket address uses IPv6. Other than that, the logic is fairly similar to `AFNetworking`. I added a bit more functionality around the `Listener` closure by allowing you to specify the `dispatch_queue_t` the closure is called on.


cnoon:
cc @jshier, @kcharwood, @kylef for review

Source/NetworkReachabilityManager.swift
```diff
@@ -0,0 +1,208 @@
...
+    /// network reachability status.
+    public typealias Listener = NetworkReachabilityStatus -> Void
+
+    // MARK: - Properties
+
+    /// Whether the network is currently reachable.
+    public var isReachable: Bool { return isReachableOnWWAN || isReachableOnWiFi }
+
+    /// Whether the network is currently reachable over the WWAN interface.
+    public var isReachableOnWWAN: Bool { return networkReachabilityStatus == .ReachableOnWWAN }
```
kylef: From what I understand, this is only possible on iOS? It might be nice to have a comment stating it doesn't apply to OS X / tvOS etc

Source/NetworkReachabilityManager.swift
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    deinit {
+        stopListening()
+    }
+
+    // MARK: - Listening
+
+    /**
+        Starts listening for changes in network reachability status.
```
kylef: It might be a good idea to mention that if `NetworkReachabilityManager` isn't retained, it will automatically stop listening. That `NetworkReachabilityManager` must be retained while you are listening.

Source/NetworkReachabilityManager.swift
```diff
@@ -0,0 +1,208 @@
...
+        previousFlags = flags
+
+        let networkReachabilityStatus = networkReachabilityStatusForFlags(flags)
+
+        listener?(networkReachabilityStatus)
+
+        dispatch_async(dispatch_get_main_queue()) {
+            NSNotificationCenter.defaultCenter().postNotificationName(
+                NetworkReachabilityStatusDidChangeNotification,
+                object: networkReachabilityStatus.rawValue
```
kylef: Shouldn't the object be `self` aka, the manager and the status in user info?

Otherwise it would be impossible to determine which manager sent this status.

Source/NetworkReachabilityManager.swift
jshier: Yes, much of this API will have to be platform specific, perhaps even the definition of `NetworkReachabilityStatus`, since, for OS X, it would be more appropriately defined as:
```swift
public enum NetworkReachabilityStatus: Int {
    case Unknown
    case NotReachable
    case Reachable
}
```
Since OS X doesn't distinguish reachability by interface.

Source/NetworkReachabilityManager.swift
cnoon: Will do. Amazing how many people get bit by this. 👍🏼

Source/NetworkReachabilityManager.swift
cnoon: Good points gents. Let me look into creating a separate status declaration for `OSX` and `tvOS`.

Source/NetworkReachabilityManager.swift
cnoon: That it would be! Will fix.

Source/NetworkReachabilityManager.swift
```diff
@@ -0,0 +1,208 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+import SystemConfiguration
+
+/// Notification posted when network reachability status changes. The notification `object` contains the update
+/// network reachability status as an `NSNumber` which will need to be converted.
+public let NetworkReachabilityStatusDidChangeNotification = "com.alamofire.network.reachability.status.did.change"
```
tkrajacic: If we use reverse domain name notation here wouldn't it make more sense to have the string be: `"com.alamofire.network.reachability.statusDidChange"` or if you must: `"com.alamofire.network.reachability.status.didChange"`

I'd actually rather suggest and follow Apple's guidelines here which make the string equal to the notification name. See [https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html](url) (look under _Notifications and Exceptions_ )

All of Apple's notifications are made up this way as well.

Source/NetworkReachabilityManager.swift
cnoon: Ended up removing notifications altogether.

Source/NetworkReachabilityManager.swift
cnoon: Ended up removing notifications altogether for this.

tt3tt3tt:
in iOS8.1, NetworkReachabilityStatus is always notReachable, how to fix it, thanks

Issue #1029, opened 2016-01-25
Collecting latency metrics 


alimoeeny:
I want to collect server response latency metrics. 
As far as I can see there is no built in support for this in Alamofire.
I have two questions:
1- Is this something that (if done properly) would be of interest to the community?
2- Am I right in assuming the best place to do this is in the `Request` object? comparing the `init` timestamp with the `cancel` or different `taskdelegate` methods?

cnoon:
Hi @alimoeeny, thanks for the awesome idea! I started playing around with different ways to make something like this to work and it turned into PR #1054! If you could check that out, I'd certainly appreciate it.

I'm going to close this issue out for now. Please redirect all comments to #1054 from here on out. Thanks! 🍻

PR #1054, opened 2016-02-07
Feature - Request Timelines

cnoon:
These changes started with me experimenting with the ideas in #1029, then became something pretty awesome!

This PR adds the ability for Alamofire to track the timings of a `Request` throughout it's entire lifecycle with virtually zero overhead. The `Timeline` object stores all the timing info about the `Request` and is delivered as part of the `Response` object. The timings are collected at various points throughout the lifecycle of the request.

Alamofire can now report the latency, request duration and total duration of any `Request`. I also added `CustomStringConvertible` and `CustomDebugStringConvertible` conformance to the `Timeline` to make it really easy to print out the info.

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -36,6 +36,9 @@ public struct Response<Value, Error: ErrorType> {
     /// The result of response serialization.
     public let result: Result<Value, Error>
 
+    /// The timeline of the complete lifecycle of the `Request`.
+    public let timeline: Timeline
+
     /**
         Initializes the `Response` instance with the specified URL request, URL response, server data and response
         serialization result.
@@ -44,14 +47,22 @@ public struct Response<Value, Error: ErrorType> {
         - parameter response: The server's response to the URL request.
         - parameter data:     The data returned by the server.
         - parameter result:   The result of response serialization.
-    
+        - parameter timeline: The timeline of the complete lifecycle of the `Request`.
+
         - returns: the new `Response` instance.
     */
-    public init(request: NSURLRequest?, response: NSHTTPURLResponse?, data: NSData?, result: Result<Value, Error>) {
+    public init(
+        request: NSURLRequest?,
+        response: NSHTTPURLResponse?,
+        data: NSData?,
+        result: Result<Value, Error>,
+        timeline: Timeline)
+    {
         self.request = request
         self.response = response
         self.data = data
         self.result = result
+        self.timeline = timeline
     }
 }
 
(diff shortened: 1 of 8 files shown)


cnoon:
cc @jshier, @kcharwood, @kylef for review

cnoon:
I'll add a section to the README about this once everyone is on board that this is a good set of changes to put in.

Source/Response.swift
```diff
@@ -77,6 +88,7 @@ extension Response: CustomDebugStringConvertible {
         output.append(response != nil ? "[Response]: \(response!)" : "[Response]: nil")
         output.append("[Data]: \(data?.length ?? 0) bytes")
         output.append("[Result]: \(result.debugDescription)")
+        output.append("[Timeline]: \(timeline.debugDescription)")
```
jshier: There's a new line between this and the `[Result]` output. Perhaps there's an extra newline in `result.debugDescription`? Trivial, but didn't quite match the rest of the output.

Source/Timeline.swift
```diff
@@ -0,0 +1,104 @@
...
+    public let serializationCompletedTime: NSDate
+
+    /// The time interval in seconds from the time the request started to the initial response from the server.
+    public let latency: NSTimeInterval
+
+    /// The time interval in seconds from the time the request started to the time the request completed.
+    public let requestDuration: NSTimeInterval
+
+    /// The time interval in seconds from the time the request started to the time response serialization completed.
+    public let totalDuration: NSTimeInterval
```
jshier: Perhaps we could additionally break serialization duration out into it's own property?

Source/Response.swift
cnoon: Ah bummer. I'll check it out.

Source/Timeline.swift
cnoon: For sure...will add. 👍🏼

alimoeeny:
Wow Dude @cnoon , this is much more complete than what I had in mind, thank you. 
I need to try it and see how I can collect all the metrics (`timelines` I guess) in one location to log or send back to a log aggregator.

Source/Response.swift
cnoon: This is possible if the value stored in the result contains newlines. There's nothing to fix here.

Source/Timeline.swift
cnoon: Added in 0a2e6a21.

cnoon:
Thanks @alimoeeny! It was definitely fun to put together.

jshier:
:+1:

@alimoeeny I do logging in my custom `.responseObject` extension, taking the closure that comes in, adding my logging call, and passing the result the actual serializer. I imagine you could do the same, especially if you have a custom response closure already.

alimoeeny:
thanks @jshier I think I'd do the same.

PR #1064, opened 2016-02-14
Simplify project setup, reduce number of framework and test targets

DenHeadless:
Hey!

This PR aims to make Alamofire universal framework, that can be built for all 4 platforms using a single target. This is achieved using https://github.com/mrackwitz/xcconfigs.

Basically, three xcconfigs are created to explicitly set SDK needed to build framework and test target. Then, when CI is building and running tests, it is enough to specify destinations, and Alamofire will be built for specific platform.

There are couple of things to consider here. First of all, all three dependency managers should still work as before. CocoaPods and Swift Package Manager do not rely on Xcode setup, Carthage supports uninversal frameworks, and successfuly builts 4 frameworks for all platforms. 

Second - CI setup should work correctly on all previously included configurations and OS-es.

The only downside i was able to find with this kind of Xcode setup is that example project target get's confused, and allows to select Mac, tvOS and watchOS target, even though it is explicitly configured to work only on iOS. But this probably is not a blocker.

Example of such Xcode setup would be [Lyft's Mapper](https://github.com/lyft/mapper) framework - it is configured exactly like this.

What do you think?


cnoon:
Hi @DenHeadless,

Generally I'm against moving to a universal framework. We've had all sorts of issues with this in the past. Here are some other obvious downsides that I can see.

* Example project doesn't seem to work (obviously this is a huge problem if true)
* Another issue could be in other frameworks that use Alamofire as a dependency (such as AlamofireImage)
    * We've run into all sorts of weird dependency issues with Xcode and multiple OS's
* Can no longer not include certain files in certain targets and everything would need to be managed by `#if os()` flags.

@kylef, I'd love to get your take on this if possible.

cnoon:
Thanks for putting this PR together @DenHeadless, but we've decided to not accept these changes. We're going to continue using multiple targets due to all the OS specific code in Alamofire.

Thanks again. 🍻

DenHeadless:
@cnoon No problem) Just out of curiosity, what was the main deciding point? I'm thinking too about including such setup in my own oss solutions.

cnoon:
Just complexity. We already have all the kinks worked out by having multiple frameworks nested through several layers of dependencies. I'm not sure how using this approach would affect other libraries like AlamofireImage and all the other libraries out there that use Alamofire as a dependency and support multiple OS's.

PR #1066, opened 2016-02-16
tag 3.2.0 with ios7 workaround

acecilia:


cnoon:
Hi @acecilia, Alamofire does not support iOS 7 and hasn't for some time. If you need to support iOS 7, then you'll need to maintain your own Alamofire fork.

Cheers. 🍻

PR #1078, opened 2016-02-23
Add codebeat badge

marcinwyszynski:
Because as everyone knows "more badges" >  "fewer badges".

medals


cnoon:
Also closed in #1085.

PR #1081, opened 2016-02-24
moved a guarded variable to a specific use case for that var

nicktrienensfuzz:
extension Request: CustomDebugStringConvertible 
Moved the URL.host form the guard to where it is actually needed, in my testing this was safe and allowed more requests to be converted.

not sure why URL.host was not returning a valid value but the requests succeed, so I'd love to have the curl command.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -470,8 +470,7 @@ extension Request: CustomDebugStringConvertible {
 
         guard let
             request = self.request,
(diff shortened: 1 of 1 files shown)


cnoon:
Thanks for the suggestion here @nicktrienensfuzz, but we'll keep the logic as is. It was originally written to ensure all three variables were available before processing the request.

Issue #1095, opened 2016-03-02
No access to completionHandler from URLSession(didReceiveChallenge: completionHandler:)


JoachimK:
I would like to implement my own handler for the authentication challenge to be able to ask the user for a username and password when a challenge is received. This would be possible in the original `NSURLSessionDelegate` by putting up a dialog for the user when the challenge is received, waiting for them to input username+password and calling the `completionHandler` from `URLSession(didReceiveChallenge: completionHandler:)` once that is done. 

This is not possible with Alamofire since both the `taskDidReceiveChallenge` and the `sessionDidReceiveChallenge` blocks are synchronous and just expect the implementation to return a credential directly and do not give access to the actual `completionHandler`. If I had the credential at that point in time I would have supplied it via `authenticate`. ;-)
The implementations in the `Request.TaskDelegate` and `Manager.SessionDelegate` also always call the `completionBlock`, so even if I got access to the `completionBlock` it would be called to early.

I originally thought I could just supply my own `NSURLSessionDelegate` to the `Manager.init` but this is not possible as it expects a `Manager.SessionDelegate` and not just any delegate protocol implementation.

Next I tried to subclass the `Manager.SessionDelegate` but *this* is not possible because it was made final.

The `Request.TaskDelegate` is not a final class but the `Request.delegate` property is `let` so I can't change that either. 

Oh, and I can't supply my own session with my session delegate the [`Manager.init` checks that it's actually a `SessionDelegate`](https://github.com/Alamofire/Alamofire/blob/master/Source/Manager.swift#L149).

So I can't find a way to access the `completionHandler ` to delay the response for the authentication challenge, not even considering how much effort it would be if there is one. And all that even though the default `NSURLSessionDelegate` has a perfectly nice way to delay the response to the authentication challenge.

Why did you lock it down so much (e.g. why can't I use my own `NSURLSessionDelegate`?) and why did you remove access to the `completionHandler` when responding to an authentication challenge.

I think the original `URLSession(didReceiveChallenge: completionHandler:)` method is awesome, because it allows you to 

1. only ask the user for a username and password when needed
2. enables you to react to changed authentication information correctly (if doing several requests in a row and the user's password changed in between, the authentication token expires or whatever)
3. Will be called again if the user enters a wrong password and the server still returns `401`. This way you get retry of incorrect passwords basically for free (the `challenge` object even tells you how many failed tries you had already) without having to do any error-code-checking-and-rescheduling-the-request.

[...]

PR #1099, opened 2016-03-06
Feature - Override Closures with Completion

cnoon:
This PR extends the `SessionDelegate` override closures to include variants that contain the `completionHandler` parameter. In certain cases, such as reported in #1095, Alamofire was not providing enough flexibility to allow a user to override the implementation and perform some async operation before calling the `completionHandler`. These changes allow users to fully override the entire set of functionality inside any of the `SessionDelegate` APIs.

> For more details about a specific use case, please refer to #1095.

An alternative approach to this PR would be to remove the `final` modifier from the `SessionDelegate` declaration. This would allow users to subclass the `SessionDelegate` which would remove the need for these variant closures. I feel though that this is not a good idea since the `SessionDelegate` does not publicly expose the subdelegates and opening this up could lead to misuse when subclassing.

cc @jshier, @kcharwood, @kylef, @AnthonyMDev

### DIFF
--- a/Source/Manager.swift
+++ b/Source/Manager.swift
@@ -254,6 +254,9 @@ public class Manager {
         /// Overrides default behavior for NSURLSessionDelegate method `URLSession:didReceiveChallenge:completionHandler:`.
         public var sessionDidReceiveChallenge: ((NSURLSession, NSURLAuthenticationChallenge) -> (NSURLSessionAuthChallengeDisposition, NSURLCredential?))?
 
+        /// Overrides all behavior for NSURLSessionDelegate method `URLSession:didReceiveChallenge:completionHandler:` and requires the caller to call the `completionHandler`.
+        public var sessionDidReceiveChallengeWithCompletion: ((NSURLSession, NSURLAuthenticationChallenge, (NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Void) -> Void)?
+
         /// Overrides default behavior for NSURLSessionDelegate method `URLSessionDidFinishEventsForBackgroundURLSession:`.
         public var sessionDidFinishEventsForBackgroundURLSession: ((NSURLSession) -> Void)?
 
@@ -281,6 +284,11 @@ public class Manager {
             didReceiveChallenge challenge: NSURLAuthenticationChallenge,
             completionHandler: ((NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Void))
         {
+            guard sessionDidReceiveChallengeWithCompletion == nil else {
+                sessionDidReceiveChallengeWithCompletion?(session, challenge, completionHandler)
+                return
+            }
+
             var disposition: NSURLSessionAuthChallengeDisposition = .PerformDefaultHandling
             var credential: NSURLCredential?
 
@@ -321,11 +329,23 @@ public class Manager {
         /// Overrides default behavior for NSURLSessionTaskDelegate method `URLSession:task:willPerformHTTPRedirection:newRequest:completionHandler:`.
         public var taskWillPerformHTTPRedirection: ((NSURLSession, NSURLSessionTask, NSHTTPURLResponse, NSURLRequest) -> NSURLRequest?)?
 
+        /// Overrides all behavior for NSURLSessionTaskDelegate method `URLSession:task:willPerformHTTPRedirection:newRequest:completionHandler:` and
+        /// requires the caller to call the `completionHandler`.
+        public var taskWillPerformHTTPRedirectionWithCompletion: ((NSURLSession, NSURLSessionTask, NSHTTPURLResponse, NSURLRequest, NSURLRequest? -> Void) -> Void)?
+
         /// Overrides default behavior for NSURLSessionTaskDelegate method `URLSession:task:didReceiveChallenge:completionHandler:`.
         public var taskDidReceiveChallenge: ((NSURLSession, NSURLSessionTask, NSURLAuthenticationChallenge) -> (NSURLSessionAuthChallengeDisposition, NSURLCredential?))?
 
+        /// Overrides all behavior for NSURLSessionTaskDelegate method `URLSession:task:didReceiveChallenge:completionHandler:` and 
+        /// requires the caller to call the `completionHandler`.
+        public var taskDidReceiveChallengeWithCompletion: ((NSURLSession, NSURLSessionTask, NSURLAuthenticationChallenge, (NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Void) -> Void)?
+
         /// Overrides default behavior for NSURLSessionTaskDelegate method `URLSession:session:task:needNewBodyStream:`.
-        public var taskNeedNewBodyStream: ((NSURLSession, NSURLSessionTask) -> NSInputStream!)?
+        public var taskNeedNewBodyStream: ((NSURLSession, NSURLSessionTask) -> NSInputStream?)?
+
+        /// Overrides all behavior for NSURLSessionTaskDelegate method `URLSession:session:task:needNewBodyStream:` and 
+        /// requires the caller to call the `completionHandler`.
+        public var taskNeedNewBodyStreamWithCompletion: ((NSURLSession, NSURLSessionTask, NSInputStream? -> Void) -> Void)?
 
         /// Overrides default behavior for NSURLSessionTaskDelegate method `URLSession:task:didSendBodyData:totalBytesSent:totalBytesExpectedToSend:`.
         public var taskDidSendBodyData: ((NSURLSession, NSURLSessionTask, Int64, Int64, Int64) -> Void)?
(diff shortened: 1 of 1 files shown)


JoachimK:
I'm wondering whether just adding blocks for all the delegate methods is a good approach. The "old" cocoa-pattern of having a delegate possibly implement a set of methods and using default behavior if a method is not implemented achieves basically the same. But since the declarations for all the delegates you need (session delegate, task delegate, etc.) are already done you don't need to write them, in contrast to the block. 

So two more high-level questions:
1. How about giving the session- and task-delegate each a delegate of same kind that users of Alamofire could implement. If a method is implemented, call it, if not use Alamofire's default implementation.
2. Why not allow an NSURLSessionDelegate instead of a `SessionDelegate` to be passed to the Manager initializer. What is special about Alamofire's SessionDelegate that would take much work to reproduce in an own delegate and everyone needs?

Also, if I see it correctly, I could implement my own delegate and make it instantiate an instance of `SessionDelegate` itself to forward methods to that if I don't want to provide a custom implementation. This way, the implementation of the `SessionDelegate` wouldn't even be lost, but could be reused.

Source/Manager.swift
```diff
@@ -281,6 +284,11 @@ public class Manager {
             didReceiveChallenge challenge: NSURLAuthenticationChallenge,
             completionHandler: ((NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Void))
         {
+            guard sessionDidReceiveChallengeWithCompletion == nil else {
+                sessionDidReceiveChallengeWithCompletion?(session, challenge, completionHandler)
```
JoachimK: shouldn't the `?` be a `!` here as `sessionDidReceiveChallengeWithCompletion` must be `!=nil`?

A way to let the compiler know that would be to use

```objc
if let didReceiveChallengeHandler == sessionDidReceiveChallengeWithCompletion {
    didReceiveChallengeHandler(session, challenge, completionHandler)
    return
}
```

Not sure whether this would be better. Has the advantage of not using `!` but the disadvantage of not ensuring we have `return` in the if-block.
So I'd probably stay with the guard but use `!` to get a runtime error instead of simply nothing happening (and the completion block never being called) 

Same argument in the case of `taskWillPerformHTTPRedirectionWithCompletion`, `taskDidReceiveChallengeWithCompletion`, `taskNeedNewBodyStreamWithCompletion `, `dataTaskDidReceiveResponseWithCompletion` and `dataTaskWillCacheResponseWithCompletion` of course

AnthonyMDev:
@cnoon I think this is a great change. The example in #1095 convinces me that this is a necessary addition. This also makes it easier to create plugins that build on top of `Alamofire`(ex. A networking logger). I also agree that this is a better solution than allowing the `SessionDelegate` to be subclassed. 

@JoachimK I'm personally a big fan of using block variables as an alternative to the delegate pattern. It feels much more Swifty to me, and is becoming a pretty popular pattern across many open source libraries.
The delegate pattern is clunky. Optional delegate methods require either `@objc`, forcing your delegate to subclass `NSObject`, or a protocol extension, which abstracts default implementation out of the delegate's definition, making it harder to understand. And I could give you 10 other reasons that I don't love the delegate pattern.

JoachimK:
Another thing about the challenge/authentication specifically: If we are going to provide our "own" API (be that in blocks or as delegate) instead of using Apple's declared delegate protocols, we could try to seize the opportunity and improve upon it. 

One of my main issues with the `didReceiveChallenge` method is that is is called both to authenticate the client to the server and to let the client verify the server, which is quite confusing.

An option to improve this is to expose the different cases in your implementation for customization. E.g. have separate delegate methods/blocks for `func verifyServerCredentials` (basically the current `evaluateServerTrust()`-method) and `func provideCredentialsForChallenge(session:NSURLSession, challenge: NSURLAuthenticationChallenge, completionBlock: (NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Void) -> Void` 

E.g., I like the verification of server certificates you build and would only need an async way to fetch a credential to authenticate the user if needed. :-)

Not sure whether this would be too specific for Alamofire to build it this way, but to me it seems like you would usually only want to override either the server trust verification OR providing credentials for the user. In addition, the verification of servers is already quite customizable in Alamofire.

JoachimK:
@AnthonyMDev Thanks a lot for your input! I'm a bit wary of using blocks/closures everywhere just because it's Swift. When Apple introduced blocks to Objective-C there was also a big hype around them and everyone suddenly wanted to use blocks for everything and get rid of delegates. However, they also have many disadvantages like leading to inadvertent retain-cycles more easily so the enthusiasm died down a bit again.

The memory management improved with Swift, making it [easier to declare how the parameters of a block should be captured](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/AutomaticReferenceCounting.html#//apple_ref/doc/uid/TP40014097-CH20-ID48), however, you still need to declare it and I'm not sure how many people even know this functionality exists in Swift. 

I'm usually using blocks for callbacks, generally for stuff that is called just once but at a later time. As soon as you need to store a block (e.g. because it's called several times) I'll usually switch to delegates.

I'm also using delegates, if there are several methods that are called and they belong together/are usually implemented in the same class/object. I think this would be the case here. 
If the normal use case is to set all the blocks of the `SessionDelegate` in the same place in the code (and I think it is), then a delegate might be better suited for the job.


Regarding the "delegate is clunky" part: We basically already have an abstract implementation of the delegate protocols with the `SessionDelegate` and the `TaskDelegate`. All they would have to do is instead of ` guard sessionDidReceiveChallengeWithCompletion == nil` do a check whether the delegate response to the appropriate delegate method.


Having said that, please take this as a recommendation only. You surely have more knowledge about how Alamofire is actually used and whether the arguments above make sense in this use-cases. ;-)

AnthonyMDev:
@JoachimK Those are all compelling arguments! Thanks for taking the time to articulate that. 

> [blocks] also have many disadvantages like leading to inadvertent retain-cycles more easily so the enthusiasm died down a bit again.

I was under the impression that delegates create more accidental retain-cycles because people may not know to make the delegate weak. In this case, using blocks should only create a retain-cycle if you are referencing an object in the block.

> If the normal use case is to set all the blocks of the `SessionDelegate` in the same place in the code (and I think it is), then a delegate might be better suited for the job.

I think you are right that, generally, you will set any applicable `SessionDelegate` blocks in the same place, but I don't think you will necessarily want to set ALL of them. You will likely only use one or two in any use case.

> All they would have to do is instead of guard `sessionDidReceiveChallengeWithCompletion == nil` do a check whether the delegate response to the appropriate delegate method.

This requires the use of optional protocol functions, right? (unless I'm confused on what you are proposing.) The problem with optional functions on protocols is that they are only supported in Objective-C, and we don't want to require that our delegate be an `NSObject` subclass.

Source/Manager.swift
cnoon: It will still be called due to the check above it. I generally stay away from all `!` usage unless it's absolutely necessary to do so.

cnoon:
@AnthonyMDev and @JoachimK, thanks for the healthy debate here. I think it always helps improve functionality when looking at approaches from many different angles. You have both certainly helped do that.

With that said, I tend to agree with @AnthonyMDev on most all the discussion here @JoachimK. You've made many interesting statements here, and I'll do my best to reply to the ones that I think could use some additional clarification.

---

> How about giving the session- and task-delegate each a delegate of same kind that users of Alamofire could implement. If a method is implemented, call it, if not use Alamofire's default implementation.

This is exactly what the override closures allow you to do. The difference is that this approach let's you override only one of the methods, not all. The approach you are suggesting is relying on @objc protocols which we're obviously going to try to avoid in Alamofire. Especially given all the changes currently taking place in Swift 3.0.

> Why not allow an NSURLSessionDelegate instead of a SessionDelegate to be passed to the Manager initializer. What is special about Alamofire's SessionDelegate that would take much work to reproduce in an own delegate and everyone needs?

Look at the implementation of the `SessionDelegate`. It does MUCH more than simply act as an `NSURLSessionDelegate`. It stores a bunch of state as well as a special `TaskDelegate` for each request. There's much more going on in there.

> Also, if I see it correctly, I could implement my own delegate and make it instantiate an instance of SessionDelegate itself to forward methods to that if I don't want to provide a custom implementation. This way, the implementation of the SessionDelegate wouldn't even be lost, but could be reused.

You can absolutely do this, but again I'd question why. Most of the time, you just want to use the override closures. There are other things that happen in those methods (such as cleanup) that you'll have to do yourself if you completely override the method.

> Another thing about the challenge/authentication specifically: If we are going to provide our "own" API (be that in blocks or as delegate) instead of using Apple's declared delegate protocols, we could try to seize the opportunity and improve upon it.

Generally I agree with this philosophy. However, I don't think it is wise in this particular case. The reason is that the delegate APIs are so complicated and come with so much documentation, that I hate to deviate too far from them. We don't want to have to re-create all the documentation that already exists for all the `NSURLSessionDelegate` protocols and all the more specific ones. Instead, Alamofire strives to provide a layer of convenience so the general user never even has to know about the delegate calls. Yet for power users, they have the ability to customize all the `NSURLSessionDelegate` calls if absolutely necessary.

---

Overall, thank you for all your feedback here @JoachimK. You've certainly helped us improve Alamofire significantly by exposing all the completion handlers directly in the override closures.

Cheers. 🍻

Issue #1097, opened 2016-03-05
Mechanism to Retry Requests


KingOfBrian:
Hello 
    I'd like to add the ability to retry requests transparently on token authentication failures, and wanted to check to see if it was functionality that might be accepted before going off and writing it. This is not a request to add oauth support, but functionality that could be used to easily add oauth support. The public API I see (names are probably not great) would look like:

```swift
extension Request {

    public typealias RetryCallback = NSURLRequest? -> Void
    public typealias RetryCheck = (NSURLRequest?, NSHTTPURLResponse?, RetryCallback) -> Void

    func check(check: RetryCheck) -> Self {
    }
}
```

This would be consumed by the client like:

```swift
request.check() {HTTPRequest, HTTPResponse, callback in
     if let HTTPResponse = HTTPResponse where HTTPResponse.statusCode == 401 {
         self.refreshAuthenticationToken() { token in
            let newRequest = HTTPRequest.mutableCopy()
            newRequest?.allHTTPHeaderFields["Authorization"] = "Bearer \(token)"
            callback(newRequest)
         }
     }
     callback(nil)
}
// Normal response handler
request.responseObject(completion)
```

*Background:*
Transparent token refresh is almost always desired when interacting with OAuth. This forces every request to re-implement painful token refresh that has to recapture all of the inputs and perform the logic above. It would be awesome to easily add this refresh behavior to the existing chain.

*Implementation Approach*
I believe this would be relatively easy to implement. First the operation queue would be suspended. If nil is passed to the callback, the operation queue would be resumed, and the remaining response chain would run. If a request was passed to the callback, all of the state of the request object would be cleared out and the new request would over-write the old one. The request would then be ran again with a new token, and hopefully succeed.

I was hoping to be able to create a request copy and copy over all of the response chain, but the response chain is built by grabbing the request via operation block closures, so I don't think that approach can work.
 
Any thoughts?

Thanks!

Brian

KingOfBrian:
So, I was looking through the implementation approach above, and it looks pretty bad. A request really doesn't want to retry, it's one request.

However, it may be easy to make the responder chain copyable. Right now it appears that the dispatch queue that represents the responder chain grabs the request state via closures. If that relation ship was stored as arrays of `ResponderChain` objects that conformed to a protocol like `func respondWithRequest(request: Request)`, a retry could move the responder chain to a new request and function as though it was one request from the API consumer perspective.

[...]

PR #1109, opened 2016-03-11
Feature - Retry Support

KingOfBrian:
I never heard back on #1097 to add support for a retry mechanism, but I thought a PR with some more tangible code could move the conversation forward. This PR is to show what that might look like, but it is probably not ready for merging. There is a test that shows what the functionality looks like [here](https://github.com/KingOfBrian/Alamofire/blob/feature/retry/Tests/RetryTests.swift#L33)

The core retry functionality is implemented by being able to copy the response chain of a request object. The request object still represents just one request, but the handler chain can be detached and moved to a new request object. This is done by passing all the response handlers through a closure that takes a request, instead of passing the closure onto the operation queue with a captured self.

### Ugly # 1
Retry functionality only works on operations that are passed through [-response](https://github.com/KingOfBrian/Alamofire/blob/feature/retry/Source/ResponseSerialization.swift#L98). All of the library handlers are passed through this method. The ugly part is that there is a public operation queue. Anything that is added to the queue by the client will not be executed if the request is retried. I'm not really sure what to do with this except document the caveat.

### Ugly # 2
The requests need to be stored in the operation queue and in an array. The value of the operation queue is now very minimal.

### Ugly # 3
Naming. I don't really like the name of either of the methods I added.


cnoon:
Thanks for putting this together @KingOfBrian. Unfortunately, we won't be taking in these changes. For more details, please see my response to #1097.

Thanks again.

PR #1112, opened 2016-03-11
Adding the possibility to receive the callback to a different queue for all responses

lucatorella:
Adding the possibility to receive the callback to a different queue to all `response*` methods.

I'm not sure why it's not like that at the moment, maybe I'm missing something 🙃

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -176,8 +176,12 @@ extension Request {
 
         - returns: The request.
     */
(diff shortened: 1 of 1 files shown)


cnoon:
Thanks for putting this together @lucatorella! You are absolutely right, these should have always been exposed. For some reason, I've always used the `response` method and passed my custom queue along with the response serializer I needed. I'm not really sure why I didn't think to add them to the top-level APIs. Thankfully you did!

Cheers. 🍻

Issue #1105, opened 2016-03-09
Issue with isReachable in NetworkReachabilityManager


IvanZhouGitHub:
I found a issue with isReachable in NetworkReachabilityManager when network condition is 100% loss.
My code :
let isReachable = NetworkReachabilityManager(host:"192.168.3.1:3000")!.isReachable
In ios8 isReachable is false
In ios9 isReachable is true
I hope you can solve my doubts.

cnoon:
Marking as a duplicate of #1086.

cnoon:
I'm going to close this issue out since it's a duplicate of #1086 @IvanZhouGitHub. Please redirect all comments to that issue instead.

Thanks! 🍻

Issue #1086, opened 2016-02-26
Reachability seems buggy on iOS 8


kchromik:
As the title already says, the new reachability feature seems a bit buggy on iOS 8. The `listener` seems to not be called each time the network status changes. Furthermore, the `network reachability status` return by the `listener` closure doesn't return the correct status.

The `status` values seem to be very random and don't represent the current state. I've tested on several iPhone devices and simulator. On iOS 9 everything works as expected.

cnoon:
Interesting. This will take some investigation on our end. I don't have access to an iOS 8.x device right now, so I'll only be able to test with the simulator which isn't as helpful.

kchromik:
Btw - I switched back to https://github.com/ashleymills/Reachability.swift, and there it works properly (Simulator & Device). I have access to iOS 8 devices, so let me know if there is something I could test.

cnoon:
Okay @kchromik, I've switched over to using IPv4 on iOS 8.x in ef756b00. Would you be able to test against the latest changes on `master` to see if they resolve your problems?

KingOfBrian:
Also, can we test that on an iOS8 device built on the iOS9 SDK? I don't think you want the compile-time check, but a runtime check (Which is why I thought it would be better to stick  with ipv4 for all).

cnoon:
The `#available` does provide a runtime check as well @KingOfBrian. The availability checks provide both compile time checks for API usage as well as runtime logic branching behavior.

KingOfBrian:
Ah sweet, I did not know that! Takes time to un-learn my obj-c ways.

cnoon:
Okay, I finally managed to track down an iOS 8.2 device. Looks like the IPv4 changes for iOS 8 were the correct solution. Reachability monitoring is now working as expected on both iOS 8 and 9. 👍🏼

PR #1113, opened 2016-03-12
Bugfix - Reachability Issue

KingOfBrian:
I was looking to clean up the unit test failures and the reachability issues reported in  #1086 and #1105 are causing unit test failures on iOS 8.x. It appears that reachability via IPv6 and `sockaddr_in6()` on iOS 8 is buggy. This updates the reachability check to use IPv4 and `sockaddr_in()` instead. That's the theory anyway, we'll see what the CI build says. I confirmed that the Reachability package that @kchromik confirmed as working is using this address.

This appears to be the simplest approach. I looked at doing some #if checks to only use ipv4 on iOS8, but it seemed to be a bit weird and inconsistent so I thought I'd go simple first. If IPv6 reachability was really desired an enum with IPV4 and IPV6 that is passed to the initializer with a default value of IPV4 would be easy.

### DIFF
--- a/Source/NetworkReachabilityManager.swift
+++ b/Source/NetworkReachabilityManager.swift
@@ -114,14 +114,14 @@ public class NetworkReachabilityManager {
     }
 
     /**
-        Creates a `NetworkReachabilityManager` instance with the default socket address (`sockaddr_in6`).
+        Creates a `NetworkReachabilityManager` instance with the default socket address (`sockaddr_in`).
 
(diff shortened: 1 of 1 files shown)


cnoon:
Thanks for investigating here @KingOfBrian. I've gone ahead and added the availability checks in ef756b00. It does seem to resolve a large portion of the test issues. I don't think we're completely in the clear yet, but it's looking much better than it did. I'm going to close this PR out and we'll continue to monitor in the two issues that are open.

Thanks again! 🍻

Issue #1072, opened 2016-02-19
No map support on Result


KingOfBrian:
I'm looking to use map on a Result to clean up some of my error processing in my response callback. Would a PR with map support be something that this project would like to include?

cnoon:
Could you put together a quick code sample demonstrating your use case? That would make it much easier for us to come to a decision.

Thanks!

fchaillou:
Hello,
What i would suggest is use this Result implementation : 
https://github.com/antitypical/Result

It has map and flatMap and is already used by other swift frameworks (Reactive Cocoa for example),
This would allow to more easily interact with it.

Thanks !

cnoon:
We will not be adding a third party library as a dependency @fchaillou. That has already been discussed in great detail throughout other issues.

@KingOfBrian, I'm going to close this issue out for now. If you post feedback here, then I can gladly re-open later.

Cheers. 🍻

fchaillou:
@cnoon noted, I disagree with your decision but i get it and you are the one making working on the project (thanks for it by the way)

For anybody who wants to use the other Result as we do here is a gist with a protocol extension :
https://gist.github.com/fchaillou/c9f798a747b939728aca

Hopes this help.
Cheers

KingOfBrian:
Sorry for the late follow up, but this is the sort of behavior that I'd like:

```swift
request.responseJSON() { response in
    let modelResult = response.result.map { value in
        guard
            let json = value as? Dictionary<String, AnyObject>,
            let token = json["token"] as? String,
            let refreshToken = json["refresh_token"] as? String
            else
        {
            return .Failure(ServiceError.errorWithCode(.InvalidResponseContent))
        }
        return OAuthResponse(token, refreshToken)
    }
```
It's just slightly nicer than checking for error in this case, but as there are multiple transformations reads much better. I agree with you on project dependencies btw, but I think the map behavior in the atypical result object could be copied over here pretty easily.

jshier:
@KingOfBrian I've implemented something similar as an extension on `Result`, and without further abusing `map`.
```swift
extension Alamofire.Result {
    typealias SuccessClosure = (value: Value) -> Void
    typealias FailureClosure = (error: Error) -> Void
    
    // TODO: Change back to onSuccess once compiler stops being dumb.
    func ifSuccess(@noescape success: Closure) -> Result {
        if isSuccess { success() }
        
        return self
    }
    
    func onSuccess(@noescape success: SuccessClosure) -> Result {
[...]

PR #1115, opened 2016-03-13
Feature - Result.ResultType conformance

KingOfBrian:
I brought up lack of map support in #1072 and there was some discussion on adding some functionality to result. I was implementing the Result functionality from antitypical, when I realized we can add support for `Result.ResultType` to `Alamofire.Result` without actually depending on the protocol. Then consumers that desire map support can add the Result framework and a small extension:

```swift
extension Alamofire.Result : ResultType {}
```

Obviously we don't need to do this and we could just tell people to add the extension this PR adds
 to `Alamofire.Result`, but the initializers and the analysis functions are helpful on their own. Also from looking at the issues, this seems to be a re-occurring request, and having a documented answer to it would probably make your life easier.

### DIFF
--- a/Source/Result.swift
+++ b/Source/Result.swift
@@ -99,3 +99,31 @@ extension Result: CustomDebugStringConvertible {
         }
     }
 }
+
+// MARK: ResultType Conformance
+
+extension Result {
+
+    /// Constructs a success wrapping a `value`.
+    public init(value: Value) {
+        self = .Success(value)
+    }
+
+    /// Constructs a failure wrapping an `error`.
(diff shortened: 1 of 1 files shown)


KingOfBrian:
Hey -- I'm going to close this. It turned out to be a bit painful without higher kinded types. There's a lot of effort the user has to take to follow to make sure Result<> implementation the developer is refering to.

cnoon:
Sounds good @KingOfBrian. I'm currently working through some solutions to try to make the `Result` type more functional friendly. More to come...

PR #1116, opened 2016-03-13
Bugfix - Kill simulator before running tests

KingOfBrian:
This will kill the simulator before running the tests. This may fix a common CI failure where xcodebuild exits with code 65.

Sorry for making the PR just to test it out. If the failure persists, I'll close this out. I don't know travis enough to do the tests locally.


KingOfBrian:
After looking a bit closer it appears to be xcpretty not catching a test failure.

PR #1120, opened 2016-03-14
ability to specify queue on which the serializer will be dispatched

krezzoid:
Added ability to specify queue for serializer, for me its highly usable if you have framework (like Realm) which don't allow to pass serialized objects between threads.

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -97,8 +97,8 @@ extension Request {
 
     /**
         Adds a handler to be called once the request has finished.
-
-        - parameter queue:              The queue on which the completion handler is dispatched.
+        - parameter queue_in            The queue on which the serializer is dispatched.
+        - parameter queue_out           The queue on which the completion handler is dispatched.
         - parameter responseSerializer: The response serializer responsible for serializing the request, response, 
                                         and data.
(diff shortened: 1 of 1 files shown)


cnoon:
Hi @krezzoid,

Thanks for putting this PR together. Unfortunately we don't think this is a good set of changes to include in Alamofire itself. This deviates greatly from the delegate queue serial processing pattern leveraged heavily by the method chaining pattern. If you need this type of functionality, then I'd suggest you create your own extensions on the `Request` object to do this.

Cheers. 🍻

krezzoid:
Hi @cnoon,

unfortunately I can't create my own `Request` extension, because in this way I will not be able to access 
`                self.delegate.data,
                self.delegate.error
` etc.,
from my extension (they are don't have public declaration and will not be visible outside Alamofire framework). Or you suggestion was about to create new extension for `Request` as a part of Alamofire framework?

cnoon:
Ah ic. In this case I think you'll want to create a custom response serializer that takes in the queue you want to run the serialization on and does it that way. For example, just execute something like `responseData` on the `TaskDelegate` queue, then process the data into your Realm model object on your provided queue. That should allow you to do what it is you're trying to accomplish.

Issue #1110, opened 2016-03-11
misusage of NSStream code in MultipartFormData


wizzardchao:
The NSStream code is synchronized. As for the doc of Apple, we needn't schedule the NSInput/OutputStream on a runloop, which will make the stream event delivered to the delegate of NSStream (which is the stream itself as we haven't set it).

KingOfBrian:
Can you explain this a bit more? I'm not sure what problem you're referring to, but I'm seeing some intermittent encoding errors that could occur from a deadlock.

wizzardchao:
I think we are not talking about the same issue. But from the document of Apple, NSStream can work on 2 mode: sync or async.
If you schedule the stream on a runloop as you can see in Alamofire, it's async. We can get notified and get data from the delegate method.
But Alamofire use the blocking api -[NSStream read:maxLength:]. In such case, we need not schedule it on a runloop.

I don't know if it will cause the deadlock issue, but it's obvious a misunderstanding of NSStream.

Correct me if I'm wrong.
Thanks!

KingOfBrian:
Yea, I see what you're saying. I don't see any reason to schedule the stream in the runloop either. Not sure it will actually cause problems, but there's no reason for it. There's a good chance the runloop doesn't even notice that it was added. 

I removed the runloop scheduling and all the unit tests still passed.

wizzardchao:
hi, @KingOfBrian . So if it's a issue, I think we should open a PR to fix it. Will you do it? Or I will do it tomorrow.

cnoon:
@wizzardchao could you post a link to the docs you're referring to?

wizzardchao:
@cnoon. You can find the example code from Apple's Stream Programming Guide.
The Run-Loop Scheduling code can be found in:
https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Streams/Articles/WritingOutputStreams.html#//apple_ref/doc/uid/20002274-BAJCABBC
The Polling code can be found in:
https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/Streams/Articles/PollingVersusRunloop.html#//apple_ref/doc/uid/20002275-CJBEDDBG

You can see there is no need to schedule on a runloop using the polling mode stream .

Hope this can help you!

jcj52436999:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

[...]

PR #1121, opened 2016-03-15
Stream Usage

KingOfBrian:
This is a small fix discovered by @wizzardchao in #1110, where the `NSInputStream` and `NSOutputStream` are used via the poll method and scheduled in the current runloop. The stream programming guide states that the stream should be used in one method or the other:

https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/Streams/Articles/PollingVersusRunloop.html#//apple_ref/doc/uid/20002275-CJBEDDBG

It doesn't appear to be causing any issues, but it should just use one method for clarity sake.


cnoon:
This looks good @KingOfBrian. Thanks for putting this together! 🍻

PR #1141, opened 2016-03-25
Fixed redirect with completion property not being called

kcharwood:
The new completion properties weren't getting called due to the bad `respondsToSelector` check.

I updated the test suite to provide test coverage for this property, as well as made that test suite a little cleaner by using a unique manager each time through.

@cnoon The same treatment may be needed here for the other properties you added to 3.3, but I'm not familiar with those yet.


cnoon:
This looks great @kcharwood. Thanks for putting it together. I'm going to merge these changes then get the other closures fixed up as well. 🍻

PR #1144, opened 2016-03-28
Allow default delegation controlled via property on SessionDelegate

taquitos:
By default, the session delegate executes a delegate block property if it exists
and then stops. If we set delegateBlocksPreventsDefaultDelegation to 'true'
delegation will continue after calling the block. 'false' is the default value.


taquitos:
Sorry, pushed to wrong repo. #lrn2Git

Issue #1145, opened 2016-03-28
Allow default delegation in SessionDelegate (with attached pr)


taquitos:
By default, the session delegate executes a delegate block property if it exists and then stops. I would like to be able to set various delegate closure properties that get executed but also continue with default handling. I need to be notified of a few methods being called (along with their args), but I want the default behavior to continue. I have a PR on a forked Alamofire open: https://github.com/taquitos/Alamofire/pull/1/files

I'm open to suggestions on how to accomplish, or make it better. I know that it isn't perfect as we can't add this feature to every delegate callback (because some require completion blocks to be called as part of the delegate callback), but I think keeping clear documentation and comments is the best solution.

Note: I didn't include unit tests yet as I'd like to make sure this approach is something people agree on before I invest the time. Also note that It seems like there are no tests that ensure the block properties are called in the SessionDelegate yet.

taquitos:
Anyway we can discuss this suggestion?

cnoon:
Hi @taquitos, my apologies for not getting back to you sooner.

Your use case here is completely valid, and is definitely something that you can't do with the current implementation. Essentially you'd like to be able to proxy a `SessionDelegate` API or two somewhere to perform some additional operations, yet still be able to use the default logic. The problem is that you can't actually do this with the basic closures or even the closures with completion.

Now I'm not entirely sure I follow what you're attempting to do in your PR, but I do have a simple solution here. I have been debating this for quite some time, and I think it might just be time to do this. If we remove the `final` signature on the `SessionDelegate`, users are free to subclass the `SessionDelegate` and proxy the delegate APIs as much as necessary for the use case. Here's what that "could" look like:

```swift
public class SessionDelegate: 
    NSObject, 
    NSURLSessionDelegate, 
    NSURLSessionTaskDelegate, 
    NSURLSessionDataDelegate, 
    NSURLSessionDownloadDelegate 
{
    // All the current implementation...
}

public class CustomSessionDelegate: Manager.SessionDelegate {
    public override func URLSession(
        session: NSURLSession, 
        dataTask: NSURLSessionDataTask, 
        didReceiveData data: NSData) 
    {
        print("I received some data!")
        super.URLSession(session, dataTask: dataTask, didReceiveData: data)
    }
}
```

[...]

PR #1161, opened 2016-04-05
Allow default delegation controlled via property on SessionDelegate

taquitos:
By default, the session delegate executes a delegate block property if it exists
and then stops. If we set delegateBlocksPreventsDefaultDelegation to 'false'
delegation will continue after calling the block (new behavior). 'true' is the
default value, which is the old behavior. The main purpose of this change is to
allow the various session delegate properties to be called with custom logic that 
**also** allows the default behavior to continue after.
See issue:  https://github.com/Alamofire/Alamofire/issues/1145 for discussion.


cnoon:
Declining this in favor of #1172. Please redirect all future comments to #1172.

PR #1164, opened 2016-04-07
Requests with invalid credentials should terminate with a 401 code

### LINKED ISSUE #1159: -999 instead of 401
When doing a .GET with authentication, things work fine if I give the right user/pwd.  If I give the incorrect user or pwd with Alamofire 3.2.1, response.response from the responseJSON method is getting nil, and the error being passed to the .Failure enum is -999.

My request manager is NOT going out of scope.

0xced:
As defined in RFC 7235 https://tools.ietf.org/html/rfc7235#section-3.1

Fixes #1159

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -284,7 +284,7 @@ public class Request {
                 }
             } else {
                 if challenge.previousFailureCount > 0 {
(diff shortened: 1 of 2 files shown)


cnoon:
This looks good @0xced...thanks for putting it together! 🍻

PR #1166, opened 2016-04-08
Better errors

0xced:
Here’s a proposition to improve how some errors are handled.


cnoon:
I really like this approach @0xced! I made a few alterations to your PR and pushed 8d9f79a8 and 2b91b8fc while maintaining your attribution.

Thanks again! 🍻

ricardopereira:
Ok, `Error.errorWithCode` is deprecated so I should replace it with what?

README should be updated:
> `let error = Error.errorWithCode(.DataSerializationFailed, failureReason: failureReason)`

PR #1172, opened 2016-04-09
Feature - Non-Final SessionDelegate

cnoon:
This PR removes the `final` keyword from the `SessionManager` declaration. These changes allow advanced users to subclass the `SessionDelegate`, but should only be used when absolutely necessary. A simple example of such a use case would be to use a proxy pattern to emit notifications or log messages when certain APIs are called. The override closures do not allow you to do this and still use the default implementation.

I also added a section to the README detailing the different ways to use the override closures vs. subclassing. Hopefully this provides enough info to allow users to make the best decision for their use case.


cnoon:
cc @taquitos, @AnthonyMDev, @jshier, @kcharwood, @kylef

README.md
```diff
@@ -1023,6 +1023,74 @@ enum Router: URLRequestConvertible {
 Alamofire.request(Router.ReadUser("mattt")) // GET /users/mattt
 ```
 
+### SessionDelegate
+
+By default, an Alamofire `Manager` instance creates an internal `SessionDelegate` object to handle all the various types of delegate callbacks that are generated by the underlying `NSURLSession`. The implementations of each delegate method handle the most common use cases for these types of calls abstracting the complexity away from the top-level APIs. However, advanced users may find the need to override the default functionality for various reasons.
```
taquitos: delegate method handle**s**

jshier:
:+1: On this, though an actual example of improper `subdelegates` usage might be a good idea, if it can be such an issue.

taquitos:
This looks great to me.

AnthonyMDev:
I'm happy with this too. I agree with @jshier about adding an example of improper usage. Otherwise, great job @cnoon!

README.md
jshier: I believe the singular is correct, as `handle` is referring do the single method being overridden, not all of the implementations together.

taquitos: #lackOfCoffeeError.
carry on :)

cnoon:
Okay awesome guys, thanks for all the feedback! I'll go ahead and merge this now and put together another PR with the example of improper usage.

PR #1187, opened 2016-04-14
Add `basicAuthenticationHeader` function

0xced:
This addresses the issue that the `Request.authenticate` function can’t be used to authenticate on servers that don’t send an authentication challenge.

This new function is roughly the equivalent of AFNetworking’s -[AFHTTPRequestSerializer setAuthorizationHeaderFieldWithUsername:password:]

Related to #1159.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -72,6 +72,16 @@ public class Request {
 
     // MARK: - Authentication
 
+    public static func basicAuthenticationHeader(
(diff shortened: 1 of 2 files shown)


cnoon:
Thanks for putting this together @0xced! I made a few small changes to the API and pushed into `master` in c2a282cc while still keeping your attribution. Thanks again!

Cheers. 🍻

PR #1188, opened 2016-04-14
Set indentation settings to four spaces at the project level

0xced:


cnoon:
Thanks for putting this together @0xced!

PR #1226, opened 2016-05-05
Adjust the order

broccolii:
Analyzing statement must before assignment


cnoon:
Thanks for this @broccolii. I modified the commit message and pushed into `master` in d3211ea9 while maintaining your attribution.

Thanks again! 🍻

PR #1229, opened 2016-05-08
README: remove content-type header field in GET request

haoxins:
It's confused to contain content-type header field in GET request


cnoon:
Hi @haoxins,

Thanks for submitting this pull request...much appreciated! I completely agree with you, that was not a good header to use as an example. Rather than remove the second header altogether as you have in this PR, I decided to change it to `Accept: application/json` instead in 20bc682d. Thanks again!

Cheers. 🍻

PR #1230, opened 2016-05-08
Define some notifications name as Enum

totomo:
**IMO** I think this PR provides following advantages.

* Ensure unique notification name
* If receive multiple notifications in one method we can write more efficiently by using switch statement like below.
```swift
func observeNotification(notification: NSNotification) {
    guard let taskName = Notifications.Task(notification.name) else {
        return
    }
    switch taskName {
        case .DidResume:
            //Do anything ...
        case .DidSuspend:
            //Do anything ...
        ...
    }
}
```

Please review when you have enough time 🙇

### DIFF
--- a/Source/Manager.swift
+++ b/Source/Manager.swift
@@ -488,7 +488,7 @@ public class Manager {
                 delegate.URLSession(session, task: task, didCompleteWithError: error)
             }
 
-            NSNotificationCenter.defaultCenter().postNotificationName(Notifications.Task.DidComplete, object: task)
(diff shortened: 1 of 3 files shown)


cnoon:
Hi @totomo,

Thank you for submitting this PR...much appreciated! Unfortunately, we cannot accept this PR at this time b/c it's not backwards compatible. I have however added a task to our Trello backlog for Alamofire 4.0 to reinvestigate whether we would like to go this route.

Thanks again! 🍻

PR #1248, opened 2016-05-13
Convert URL date parameters to ISO8601 format

jamaalalleyne:


cnoon:
Hi @jamaalalleyne,

It is not valid to try to encode an `NSDate` using `NSJSONSerialization`. Instead of adding support like this into Alamofire, you should instead encode the date into a string, then pass that string into the parameters argument.

Cheers. 🍻

jamaalalleyne:
Hi @cnoon,

Are you guys planning to support the ISO8601 date format for query params?

jamaal

cnoon:
No, you need to convert the `NSDate` into a `String` using whatever format you need to send to your server, then pass that `String` into Alamofire to encode it into JSON.

jamaalalleyne:
@cnoon thank you for your help. I really appreciate you taking the time to respond to me.

Ok...If I have to, i will go about converting NSdate to String. But for now I am going to continue using my forked version of alamofire, which is formatting dates in ISO8601 format.

**Note well - the remainder of this statement refers to Swagger and its use of Alamofire. I am just trying to shed some light on what I am doing**

Also, I am using a Swift code swagger client which has Alamofire as a dependency. So I am using the swagger generated client to make the necessary requests(with the dates)

Also I didn't have this issue when i was using Objective C swagger client which was using CFNetworking, which i believe Alamofire uses(CFNetworking that is?...or i could be wrong)

Ok thank you.
Jamaal

PR #1257, opened 2016-05-20
Update 'Generic Response Object Serialization' section on README by using protocol extension

raphaeloliveira:
Hi guys, I'm wondering if would be a good idea on README to use a protocol extension instead of a direct implementation of `ResponseCollectionSerializable ` on `User`. This way it's easier to make any type conform to `ResponseCollectionSerializable`.

If you guys feel like a more functional approach I can change to 

```Swift
extension ResponseCollectionSerializable where Self: ResponseObjectSerializable {
    static func collection(response response: NSHTTPURLResponse, representation: AnyObject) -> [Self] {
        guard let representation = representation as? [[String: AnyObject]] else { return [] }
        return representation.flatMap { Self(response: response, representation: $0) }
    }
}
```  

Happy to hear your thoughts. 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -809,6 +809,22 @@ public protocol ResponseCollectionSerializable {
     static func collection(response response: NSHTTPURLResponse, representation: AnyObject) -> [Self]
 }
 
+extension ResponseCollectionSerializable where Self: ResponseObjectSerializable {
+    static func collection(response response: NSHTTPURLResponse, representation: AnyObject) -> [Self] {
+        var collection = [Self]()
+        
+        if let representation = representation as? [[String: AnyObject]] {
+            for itemRepresentation in representation {
+                if let item = Self(response: response, representation: itemRepresentation) {
+                    collection.append(item)
+                }
+            }
+        }
+        
+        return collection
+    }
+}
+
 extension Alamofire.Request {
     public func responseCollection<T: ResponseCollectionSerializable>(completionHandler: Response<[T], NSError> -> Void) -> Self {
         let responseSerializer = ResponseSerializer<[T], NSError> { request, response, data, error in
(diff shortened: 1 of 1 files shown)


cnoon:
I think this is a great improvement @raphaeloliveira...thanks for putting this together! I rebased your change and pushed in master as dae3ed1b while keeping your attribution.

Cheers. 🍻

giofid:
I think that in the [Generic Response Object Serialization](https://github.com/Alamofire/Alamofire#generic-response-object-serialization) section on README
```
struct User: ResponseObjectSerializable, ResponseCollectionSerializable, CustomStringConvertible {
    let username: String
    let name: String

    var description: String {
        return "User: { username: \(username), name: \(name) }"
    }

    init?(response: HTTPURLResponse, representation: Any) {
        guard
            let username = response.url?.lastPathComponent,
            let representation = representation as? [String: Any],
            let name = representation["name"] as? String
        else { return nil }

        self.username = username
        self.name = name
    }
}
```
should be
```
struct User: ResponseObjectSerializable, CustomStringConvertible {
    let username: String
    let name: String

    var description: String {
        return "User: { username: \(username), name: \(name) }"
    }

    init?(response: HTTPURLResponse, representation: Any) {
        guard
            let username = response.url?.lastPathComponent,
            let representation = representation as? [String: Any],
            let name = representation["name"] as? String
        else { return nil }

        self.username = username
        self.name = name
    }
}
```

`User` shouldn't  adopt the `ResponseCollectionSerializable` protocol.

raphaeloliveira:
Hi @giofid, why are you suggesting `User` should not be able to be parsed as a collection of `User`s?

giofid:
Hi @raphaeloliveira, you're absolutely right! My mistake, sorry.

Issue #1232, opened 2016-05-10
Memory leak when print request and response debugDescription


sosng:
Hello,
  I noticed that in my code , when use 

```swift
print(response.debugDescription) 
or
print(request.debugDescription)
```
There will be a memory leak shows in instrument.

Here is my code:
```swift
 Alamofire.request(.GET, "https://www.github.com").responseJSON { (response) in
    print(response.debugDescription)
}
```
and this is the screenshoot of instrument
![](https://raw.githubusercontent.com/sosng/AlamofireMemoryLeakSample/master/screenshoot.png)

you can see more in my [repo](https://github.com/sosng/AlamofireMemoryLeakSample)

thanks a lot ~

cnoon:
Hi @sosng,

Thanks for putting this issue together along with the sample project. I'm able to reproduce the leaks you're referring to. They have to do with leaks in string interpolation for building the `debugDescription` strings. I should have a fix in for this over the next few days.

Thanks again for reporting! 🍻

cnoon:
I've opened up PR #1262 to address this issue. Please direct all future comments to #1262.

sosng:
Good Job!

PR #1262, opened 2016-05-22
Fix - String Interpolation Leaks

cnoon:
This PR fixes memory leaks that were occurring in the string interpolation for the `description` and `debugDescription` implementations in the `Timeline` struct. None of these changes "should" be necessary, but unfortunately are at the moment. The main change was to switch over from string interpolation to string concatenation. I've been debugging with the sample project provided in #1232 and everything seems to be resolved with these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1250,6 +1250,7 @@ There are some important things to remember when using network reachability to d
 The following rdars have some affect on the current implementation of Alamofire.
 
 * [rdar://21349340](http://www.openradar.me/radar?id=5517037090635776) - Compiler throwing warning due to toll-free bridging issue in test case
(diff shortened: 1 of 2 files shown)


jshier:
👍 

Though, perhaps make a `TODO` or otherwise indicate that the code should be moved back to interpolation when the leaks are fixed. Also, was there a bug filed with Apple for this issue? Radar or Swift project would work.

cnoon:
I agree @jshier. I put together the [Rdar-StringInterpolationMemoryLeak](https://github.com/Alamofire/Rdar-StringInterpolationMemoryLeak) project and filed [rdar://26761490](https://openradar.appspot.com/radar?id=5010235949318144) with Apple. I also added `NOTE`s to the code and added the rdar to the README in 6c1cbdd7.

Issue #1122, opened 2016-03-16
SSL Connection does not work when using a Background Session Configuration


matthiaslau:
When using a Background Session Configuration instead of the default to call a HTTPS URL the serverTrustPolicyManager is never used and the download fails with "An SSL error has occured and a secure connection to the server cannot be made.".

I create my Manager like this
```
let serverTrustPolicies: [String: ServerTrustPolicy] = [
    Consts.host: .PinCertificates(
        certificates: ServerTrustPolicy.certificatesInBundle(),
         validateCertificateChain: true,
         validateHost: false
     )
]
let configuration = NSURLSessionConfiguration.backgroundSessionConfigurationWithIdentifier("projectid.background")
let downloadManager = Alamofire.Manager(
    configuration: configuration,
    serverTrustPolicyManager: ServerTrustPolicyManager(policies: serverTrustPolicies)
)
```

The `didReceiveChallenge` of the SessionDelegate is never called. Using `NSURLSessionConfiguration.defaultSessionConfiguration()` everything works as expected and the SSL certificate is successfully validated. Does anyone have an idea why the didReceiveChallenge is not called when using a background configuration? I couldn´t figure it out yet.

I am using XCode 7.2.1 testing on iOS 9.2.1.

cnoon:
This is a great question @matthiaslau. Personally I haven't tried to use cert pinning yet with a background session. If anyone else has any info here, please share!

Are you running into this problem when the app is running in the background, or when it's running in the foreground? There are some obvious problems right now with the Alamofire design working with background sessions. Such as trying to reconnect your session and all the response closures being dropped.

Could you give me a more detailed explanation of your use case?

matthiaslau:
This problem directly occurs regardless of whether in foreground or background. In my usecase we are communicating with a server running on a hardware we are producing using certificate pinning for security reasons. When we are downloading files from this server I want the app to continue in the background so the download is finished even when the user leaves the app.

The interesting part is that without putting the app into background `didReceiveChallenge`is not called just because the session configuration is different.

@cnoon: do you have a rough idea where I can look at for further investigation? Thanks for your help! :-D

cnoon:
Hey @matthiaslau, have you made any progress here? 

[...]

PR #1263, opened 2016-05-22
Test call of session:didReceiveChallenge in background session

bclausdorff:
A test that will fail in order to investigate #1122 

=> session:didReceiveChallenge is being called when using default configuration but won't be called when using background session configuration


cnoon:
Hi @bclausdorff,

Thank you for taking the time to put this together. I'm going to go ahead and close this out since there's nothing we can do here. Please see my reasoning in #1122 for further details.

Cheers. 🍻

PR #1288, opened 2016-06-02
Update README.md

lferro9000:
Update cocoapod install to be compatible with cocoapod 1.x as it now requires a concrete target.


cnoon:
Thanks for putting this together @lferro9000!

PR #1292, opened 2016-06-05
[1269] Update generation of the OS version part of the user agent to use non-localized sources.

### LINKED ISSUE #1269: User-Agent is localized
Hi, I see that `User-Agent` in `Manager` https://github.com/Alamofire/Alamofire/blob/master/Source/Manager.swift#L64 is localized

because of this 

```swift
let os = NSProcessInfo.processInfo().operatingSystemVersionString
```

> The operating system version string is human readable, localized, and is appropriate for displaying to the user. This string is not appropriate for parsing.

is this good?

jshier:
Fixes #1269. Updates the user agent’s OS version to use non-localized sources. It also includes the actual OS name in the string.

### DIFF
--- a/Source/Manager.swift
+++ b/Source/Manager.swift
@@ -61,9 +61,29 @@ public class Manager {
                 let executable = info[kCFBundleExecutableKey as String] as? String ?? "Unknown"
                 let bundle = info[kCFBundleIdentifierKey as String] as? String ?? "Unknown"
                 let version = info[kCFBundleVersionKey as String] as? String ?? "Unknown"
-                let os = NSProcessInfo.processInfo().operatingSystemVersionString
+                let osNameVersion: String = {
+                    let version = NSProcessInfo.processInfo().operatingSystemVersion
+                    let versionString = "\(version.majorVersion).\(version.minorVersion).\(version.patchVersion)"
+                    let osName: String = {
+                        #if os(iOS)
+                            return "iOS"
+                        #elseif os(watchOS)
+                            return "watchOS"
+                        #elseif os(tvOS)
+                            return "tvOS"
+                        #elseif os(OSX)
+                            return "OS X"
+                        #elseif os(Linux)
+                            return "Linux"
+                        #else
+                            return "Unknown"
+                        #endif
+                    }()
+                    
+                    return "\(osName) \(versionString)"
+                }()
 
-                var mutableUserAgent = NSMutableString(string: "\(executable)/\(bundle) (\(version); OS \(os))") as CFMutableString
+                var mutableUserAgent = NSMutableString(string: "\(executable)/\(bundle) (\(version); \(osNameVersion))") as CFMutableString
                 let transform = NSString(string: "Any-Latin; Latin-ASCII; [:^ASCII:] Remove") as CFString
 
                 if CFStringTransform(mutableUserAgent, UnsafeMutablePointer<CFRange>(nil), transform, false) {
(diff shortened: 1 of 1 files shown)


cnoon:
Okay so I've spent some time making sure that we get this exactly right. Here's a few examples of what we get if we don't change anything.

**English**
* Process Info: `Version 9.3 (Build 13E230)`
* User Agent: `appname/bundle_id (1606122225; OS Version 9.3 (Build 13E230))`

**Japanese**
* Process Info: `バージョン 9.3（ビルド 13E230）`
* User Agent: `appname/bundle_id (1606122229; OS bajon 9.3(birudo 13E230))`

**Spanish**
* Process Info: `Versión 9.3 (Fase 13E230)`
* User Agent: `appname/bundle_id (1606122227; OS Version 9.3 (Fase 13E230))`

**Russian**
* Process Info: `Версия 9.3 (Выпуск 13E230)`
* User Agent: `appname/bundle_id (1606122226; OS Versia 9.3 (Vypusk 13E230))`

**Thai**
* Process Info: `เวอร์ชั่น 9.3 (บิลด์ 13E230)`
* User Agent: `appname/bundle_id (1606122228; OS wexrchan 9.3 (bi ld 13E230))`

Therefore, we definitely need to make this change! 

Now what's interesting is that we no longer need to run the `CFStringTransform`. It's not possible to have a non-latin character come through as the app name or the bundle id. I spent some time trying this out, and Apple replaces any non-latin characters in the app name or bundle id with nothing or a `-`. Therefore, we no longer need to run the `CFStringTransform` on this string to be safe. I'll update the PR based on this info.

cnoon:
Removed the `CFStringTransform` logic in 3591bfac and then rebased.

Source/Manager.swift
```diff
@@ -61,14 +61,31 @@ public class Manager {
...
                 let version = info[kCFBundleVersionKey as String] as? String ?? "Unknown"
-                let os = NSProcessInfo.processInfo().operatingSystemVersionString
 
-                var mutableUserAgent = NSMutableString(string: "\(executable)/\(bundle) (\(version); OS \(os))") as CFMutableString
-                let transform = NSString(string: "Any-Latin; Latin-ASCII; [:^ASCII:] Remove") as CFString
-
-                if CFStringTransform(mutableUserAgent, UnsafeMutablePointer<CFRange>(nil), transform, false) {
-                    return mutableUserAgent as String
-                }
+                let osNameVersion: String = {
```
onmyway133: @jshier thanks, but we miss the OS build version, right?

cnoon: Yes, the build number is no longer going to be present because we cannot extract it in a non-localized manner.

bgerstle:
@cnoon I would love to know how you tested this. I [mentioned](https://github.com/Alamofire/Alamofire/issues/1269#issuecomment-224777738) that I tried a few things, but couldn't get the `operatingSystemVersionString` value to change.  Thanks for providing a fix we can all benefit from!

cnoon:
Hi @bgerstle,

One of the apps I work on is localized in 20 languages. All I did was drop some log messages into our `AppDelegate` printing out the `operatingSystemVersionString` values and ran it in the different languages posted above as an experiment. 

You can easily add additional localizations to a sample project by adding a new localization in your project's `Info` settings. Here's a [link](http://www.appcoda.com/localization-tutorial-ios8/) to get you started if you haven't done this before. Then you can switch between locales when launching by switching Scheme's runtime language which can be done by `Edit Scheme` -> `Run` -> `Options` -> `Application Language`.

Cheers. 🍻

bgerstle:
Thanks! But the strange thing is, we've had reports of this happening in
languages our app isn't localized in 😖. Anyway, thanks again for providing
more info!

PR #1305, opened 2016-06-10
Add indexing to queryComponents with Array as value

CelpanDmitrii:
When encoding(URL or URLEncodeInURL) a dictionary with array as value
the result combinations of main dictionary key and array elements must
be indexed.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -191,8 +191,8 @@ public enum ParameterEncoding {
                 components += queryComponents("\(key)[\(nestedKey)]", value)
             }
         } else if let array = value as? [AnyObject] {
-            for value in array {
-                components += queryComponents("\(key)[]", value)
+            for (index, value) in array.enumerate() {
+                components += queryComponents("\(key)[\(index)]", value)
             }
         } else {
             components.append((escape(key), escape("\(value)")))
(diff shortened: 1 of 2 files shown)


CelpanDmitrii:
When using next structure as parameter
{"filter": [ {"field": "user_id",
  			  "type": "eq",
  			  "value": 2 },
 			 {"field": "someField",
  			  "type": "eq",
  			  "value": 0 }]}
then the formatted string for url(before encoding) must be

filter[0][field]=user&filter[0][value]=7&filter[0][type]=eq&filter[1][field]=someField&filter[1][type]=eq&filter[1][value]=0

and not

filter[][field]=user&filter[][value]=7&filter[][type]=eq&filter[][field]=someField&filter[][type]=eq&filter[][value]=0

cnoon:
Hi @CelpanDmitrii,

Thank you for putting together this pull request...much appreciated! 

In this case, you'll want to use a `.Custom` parameter encoding since we don't want to change the default behavior. Since there's no documented spec on how to encode collection types with URL encoding, Alamofire uses one of the most common approaches. Anything outside this approach MUST be handled with a `.Custom` encoding. This is documented in the ParameterEncoding [docstrings](https://github.com/Alamofire/Alamofire/blob/master/Source/ParameterEncoding.swift#L41-L47).

Thanks again. 🍻

Issue #1312, opened 2016-06-14
Swift 2.3


KingOfBrian:
Turns out there are a few nullability changes that are causing build failures with Swift 2.3.

1 - A number of `NSURL` functions return nullable `NSURL` instead of nonnull. 
2 - `SecTrustResultType` moved to an enum from a collection of global Ints.
3 - `SecTrustSetPolicies(trust, policies)` no longer compiles: `Argument type '[SecPolicy?]' does not conform to expected type 'AnyObject'`

davetroy:
@KingOfBrian Just a field report on Xcode 8/iOS 10 — I am using your forked copy (feature/Swift2.3 branch) and after using the code migrator (to 2.3) and then compiling I am able to make it run, but HTTP requests do not seem to be completing properly. Lots of logging when running in Simulator. Unable to make it work properly on either simulator or device.

These are the last few log lines when running in Simulator:
```
2016-06-14 21:21:30.531880 NewApp[26994:303708] [] nw_endpoint_flow_protocol_connected [21.1 54.243.148.90:443 in_progress socket-flow (satisfied)] Output protocol connected
2016-06-14 21:21:30.532781 NewApp[26994:303708] [] nw_endpoint_flow_connected_path_change [21.1 54.243.148.90:443 ready socket-flow (satisfied)] Connected path is satisfied
2016-06-14 21:21:30.533201 NewApp[26994:303708] [] nw_connection_endpoint_report [21.1 54.243.148.90:443 ready socket-flow (satisfied)] reported event flow:finish_connect
2016-06-14 21:21:30.533567 NewApp[26994:303708] [] nw_connection_endpoint_report [21 e.crashlytics.com:443 ready resolver (satisfied)] reported event flow:finish_connect
2016-06-14 21:22:01.235541 NewApp[26994:303708] [] nw_endpoint_handler_cancel [21 e.crashlytics.com:443 ready resolver (satisfied)]
2016-06-14 21:22:01.236215 NewApp[26994:303708] [] nw_endpoint_handler_cancel [21.1 54.243.148.90:443 ready socket-flow (satisfied)]
2016-06-14 21:22:01.236853 NewApp[26994:303708] [] __nw_socket_service_writes_block_invoke sendmsg(fd 15, 31 bytes): socket has been closed
[...]

PR #1313, opened 2016-06-14
[wip] Swift 2.3 Compatibility

KingOfBrian:
This PR fixes some of the issues identified by #1312. This is going to need a lot of #availability wrappers, but thought I'd put it up here in case any one was curious how the rest of their code was doing on Swift 2.3

# Code Changed
The following changes were made aligned with the issues in #1312:

1. Explicitly unwrap NSURL as needed. I don't like doing this, but not sure what else to do.
2. Moved to the Enum type.
3. Explicitly unwrap SecPolicy creations. It appears that arrays with nil objects will not bridge to AnyObject's, that CFType expects to be an array.

# Unknown's
I'm not sure what the proper fix is for Swift 2.3. Changing the project file will prevent this project from supporting 2.3 and 3.0 without a branch. I imagine a lot of people will be thinking about the best solution here, the easy quick fix was to change the project file.

# New Test Failures
There are a few new test failures that need further investigation.

1. Three redirection tests are failing
2. testThatUploadingMultipartFormDataOnBackgroundSessionWritesDataToFileToAvoidCrash fails
3. testThatBackgroundConfigurationHeadersAreSentWithRequest fails

Brian


cnoon:
This is a great start @KingOfBrian...thanks for putting all this together! The `NSURL` changes are really rough. We'll need to spend some time getting familiar with the failure cases of the new optionality to understand if force unwrapping is really the correct solution.

avaidyam:
I would like to point out that I've got a Swift 3 compiling branch over [here.](https://github.com/avaidyam/Alamofire/tree/swift-3.0)

KingOfBrian:
I had a coworker ask about the nullability at the labs. I'm not sure it's a complete or correct answer, but I thought I'd include it for reference.

```
File reference URLs persist while the file moves. 
Trying to do append or other operations on a file reference URL will return nil if the file has been deleted. 
The only times the NSURL methods can return nil: dealing with reference URLs, or sometimes with an empty path.
```

Honestly the persist while-moving doesn't make sense to be, but it appears a file reference URL has some magic which forced these changes.  Given that this is the case, it appears http/https URL objects don't have this concern. But the 'sometimes with an empty path' doesn't instill too much confidence.

ZevEisenberg:
I'm the coworker. Going to follow up in the labs. When appending a path component to a file reference URL (those look like `file://some-identifier-string`), it `stat`s the file to see if it exists, and if it doesn't, it returns nil. Apparently.

aaroncrespo:
> Explicitly unwrap SecPolicy creations. It appears that arrays with nil objects will not bridge to AnyObject's, that CFType expects to be an array.

How about instead of explicit unwrapping SecPolicy you try:
```
let policy = SecPolicyCreateSSL(true, validateHost ? host as CFString : nil)
SecTrustSetPolicies(serverTrust, [policy].flatMap { $0 })
```
Would that allow `trustIsValid(serverTrust)` to return expected results for policy unhappy path?

brentleyjones:
Before this is merged can it use `#ifswift`? I would love to have my Swift 2.2 and Swift 2.3 code be the same code, so I can use Xcode 8 as my main driver and submit with Xcode 7.3.

ZevEisenberg:
I was told in the labs that it's not really supported to target multiple versions of Xcode or the SDK :\

kcharwood:
@brentleyjones I just pushed up a branch named `swift2.3` that uses the `#if swift` macros.

I think its important we do this for the AF 3.x branch so people can easily transition to swift 2.3 if they choose, with the same version of Alamofire.

Obviously, that won't be possible with Swift 3.0 given the huge diff.

kcharwood:
> There are a few new test failures that need further investigation.

> testThatUploadingMultipartFormDataOnBackgroundSessionWritesDataToFileToAvoidCrash fails
> testThatBackgroundConfigurationHeadersAreSentWithRequest fails

Both of these are in relation to a radar I filed. 

```
26870455: NSURLSession with background configuration does not function on the iOS simulator inside of test targets
````

**Comment by gottsohn** [comment]
@KingOfBrian Any updates to your PR as regards the failing tests?

From TravisCI, the unwrapping causes the error, in line _313_ of _Uploader.swift_

```
let directoryURL = tempDirectoryURL.URLByAppendingPathComponent("com.alamofire.manager/multipart.form.data")!
```

cnoon:
Thanks for everyone's amazing work throughout this issue...very much appreciated! 🍻

We now have our `swift2.3` branch is tip-top shape, building against Swift 2.2 and 2.3. We also have our travis-ci tests running against both Xcode 7 and Xcode 8 properly. At this point, I'm going to go ahead and close this issue out. We'll open new issues if necessary when the next round of betas come out.

Thanks again everyone!

PR #1315, opened 2016-06-16
Add 'public' to enum 'BackendError'

adrianbrink:
Add the 'public' identifier to the 'BackendError' enum in the part about using Generics for Serialization. 

Reason:

The extension of Request throws an error ("Method cannot be declared public because its parameter uses an internal type") unless the enum is public.


cnoon:
Thanks for putting this together @adrianbrink! I made some small tweaks to your commit message and pulled the change into `master` in a53cc42b while keeping your attribution.

Thanks again! 🍻

PR #1317, opened 2016-06-19
Use set operations to check for matching certificates and keys

marius-serban:
When searching for a match between the pinned certificates and the server certificates, using set operations would express the logic in a simpler way. Same thing goes for public key matching.


marius-serban:
Just realised unit tests fail on iOS 8, will close this PR for now

PR #1318, opened 2016-06-20
JSON parameters shouldn't be expected to be in a particular order

marius-serban:
This fixes the unit [test failure](https://travis-ci.org/Alamofire/Alamofire/jobs/137143030) on iOS 8.1 and 8.2 simulator.

The cause of the issue was that the order of items in a dictionary is different between iOS 8 and iOS 9.

### DIFF
--- a/Tests/RequestTests.swift
+++ b/Tests/RequestTests.swift
@@ -597,12 +597,10 @@ class RequestDebugDescriptionTestCase: BaseTestCase {
             "command should contain 'application/json' Content-Type"
         )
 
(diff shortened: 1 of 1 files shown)


kcharwood:
This is also different in iOS 10 as well.

cnoon:
Thanks for putting this together @marius-serban! I wish I would have done it this way in the first place...would have saved us a bunch of headaches. Anyways, I've made a few small tweaks to this logic and pushed it into `master` in 32ecfd16 while keeping your attribution. I'm also going to make sure these changes get cherry picked into the `swift2.3` and `swift3` branches.

Thanks again @marius-serban! 🍻

PR #1322, opened 2016-06-24
Swift lint parsing

gauravds:


cnoon:
Hi @gauravds, thank you for putting this PR together...much appreciated!

While I completely agree the whitespace should be cleaned up, I don't agree with some of the formatting changes in this PR. Because of this, I've added a `Cleanup Whitespace` aggregate target to Alamofire that will remove all the extra tabs and spaces at the end of any line. I added the target in 7789baaa and ran it in c5fe5da0. I'll be adding this new target to both AFI and AFNAI as well to keep things tidy moving forward.

Thanks again. 🍻

PR #1330, opened 2016-07-03
hold the original NSMutableURLRequest's allHTTPHeaderField and cacheP…

passchaos:
I think the original NSMutableURLRequest's property should hold, not create just from request URL!


cnoon:
Thanks for putting this together @passchaos...I couldn't agree more! I made some changes to your implementation and pushed it to `master` in bb55b866 while maintaining your attribution.

Thanks again! 🍻

PR #1368, opened 2016-07-25
Update swift3 conditions for "if let"

mendesbarreto:
When you try to use the Alamofire from branch swift3 with Xcode Version 8.0 beta 2 (8S162m), the compiler complains about the old statements used. Like:  

Old one:
guard let validData = self.delegate.data, validData.count > 0 else { return .success }

New one
guard let validData = self.delegate.data where validData.count > 0 else { return .success }

This pull request updates all statements on this branch.

### DIFF
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -268,7 +268,7 @@ public class MultipartFormData {
         var isDirectory: ObjCBool = false
 
         guard
-            let path = fileURL.path,
+            let path = fileURL.path where
             FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory) && !isDirectory else
         {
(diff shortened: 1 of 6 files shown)


cnoon:
The `swift3` branch now requires Xcode 8 beta 3+. We've made these changes on purpose. Please update to the beta 3 build and everything will work properly.

Cheers. 🍻

mendesbarreto:
Hello Cnoon, How it's going?

Thank you for your replay, but what is the purpose of the comma instead of where?

ikesyo:
`where` in this case is deprecated now.

mendesbarreto:
Thanks for all fellows..

Gerst20051:
https://github.com/apple/swift-evolution/blob/master/proposals/0099-conditionclauses.md

PR #1375, opened 2016-07-29
fix a swift bug,which make the request error

zkh90644:
When I want to get the image from my school website,there are some error.
I found if the url with **Chinese words(maybe other languages)**,the init of NSURL return **nil**.(It's the swift bug,not your code )
So I add an String extension in the file to make the framework stronger, and run the demo,the Problem solved .I'm sorry I cant read all files at once ,so maybe It's not the best position to code,but I wish it will help you .Thanks

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -59,6 +59,15 @@ extension NSURLRequest: URLStringConvertible {
     public var URLString: String { return URL!.URLString }
 }
 
+extension String {
+    func urlEncode() -> String? {
+        let unreserved = "!=:-._~/?"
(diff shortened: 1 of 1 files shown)


cnoon:
Hi @zkh90644,

This is something that you need to keep in your own logic. The `URLRequestConvertible` protocol clearly calls out in the documentation that your `NSURL` instances MUST conform to RFCs [1738](https://tools.ietf.org/html/rfc1738) and [1808](https://tools.ietf.org/html/rfc1808) to be used safely inside Alamofire. Since there are MANY different ways to safely escape your URLs, we leave this to the client implementations.

Cheers. 🍻

zkh90644:
Thank you so much to tell me these things.

PR #1386, opened 2016-08-08
Feature - Moving SessionDelegate to Store Requests

cnoon:
This PR modifies the `SessionDelegate` to store `Request`s internally instead of only the task delegates of the `Request` in preparation for more advanced features such as retry logic.

> There are no functional changes to the behavior in this PR.


Source/Request.swift
```diff
@@ -35,7 +35,16 @@ public class Request {
...
+    public internal(set) var delegate: TaskDelegate {
+        get {
+            taskDelegateLock.lock() ; defer { taskDelegateLock.unlock() }
+            return taskDelegate
+        }
+        set {
+            taskDelegateLock.lock() ; defer { taskDelegateLock.unlock() }
+            taskDelegate = newValue
+        }
+    }
```
cnoon: This may be overkill to wrap the change here in a lock, but I figure it's better safe than sorry. The only time this can ever change is when a data task is converted to a download task. Previously we were changing the subdelegate and not actually updating the request. That then created a mismatch between what the `Request` delegate reported and what was actually called.

cnoon: If we decide this is unnecessary, then it's really simple. We just need to remove the `taskDelegate` and the `taskDelegateLock` and revert this change while maintaining the `internal(set) var delegate` change.

Source/TaskDelegate.swift
```diff
@@ -33,7 +33,7 @@ public class TaskDelegate: NSObject {
     /// The serial operation queue used to execute all operations after the task completes.
     public let queue: OperationQueue
 
-    let task: URLSessionTask
+    var task: URLSessionTask
```
cnoon: This is the most important part of this change. By allowing the underlying `task` to be swapped on the `TaskDelegate`, we can essentially defer the suspended queue operations while swapping out the `task` inside the same `Request`. This will allow us to create a new task from the `Request` and swap the task in the `TaskDelegate`.

cnoon:
Closing this out in favor of #1391.

PR #1392, opened 2016-08-10
Swift 3 compliance

florentbruel:
ServerTrustPolicy has some compile errors due to swift3 updates.

function "public static fun certificatesInBundle" does not compile due to "bundle.paths(forResourcesOfType: fileExtension, inDirectory: nil)". This call is now to be replaced by bundle.pathsForResources(ofType: fileExtension, inDirectory: nil)


jshier:
Current `swift3` branch is up to date with beta 5.

PR #1395, opened 2016-08-10
Update MultipartFormData.swift - fix parameter: withN -> withName

jhudsonWA:
fix MultippartFormData.append overload parameter name withN -> withName. This matches the rest of the append overloads in this file


cnoon:
Thanks for putting this together @jhudsonWA! It looks like we just missed this in the refactor.

PR #1399, opened 2016-08-11
Swift3

JourneyWind:
Refactored All APIs for Swift 3


cnoon:
Hi @JourneyWind,

We won't be merging into `master` until the Xcode 8 GM drops. At that time, we'll be releasing both the `swift2.3` branch as AF 3.5.0 and the `swift3` branch as AF 4.0.0. This is our current plan, but we're trying to squeeze in some pretty big changes for AF 4.0 that may cause us to delay slightly.

Cheers. 🍻

PR #1401, opened 2016-08-11
Update podspec to 4.0.0, update deployment targets.

jshier:
Updates the podspec to version 4.0.0 and sets the deployment targets to match the project’s new deployment targets.

Also fixes a build issue with the iOS Example app.

### DIFF
--- a/Alamofire.podspec
+++ b/Alamofire.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name = 'Alamofire'
-  s.version = '3.4.1'
+  s.version = '4.0.0'
   s.license = 'MIT'
   s.summary = 'Elegant HTTP Networking in Swift'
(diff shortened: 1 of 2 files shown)


cnoon:
Looks good to me @jshier.

rs10693:
Hello @cnoon,

I am using Alamofire but right now i am facing issue on swift 3. so please help me. Here i am adding my pod file text and also error.

My Deployment target is 8.0 

source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '8.0'
use_frameworks!
target 'project' do

pod 'Alamofire', '~> 4.0'

end
If i am wrong then suggest me.
error

cnoon:
You cannot use Alamofire 4 with a deployment target of iOS 8.0 @rs10693.

rs10693:
so i want  to use alamofire library in deployment target 8.0 and also use in swift 3, so plz suggest better option. @cnoon

PR #1402, opened 2016-08-12
Swift3

JourneyWind:
Refactored All APIs for Swift 3


jshier:
We'll be merging the `swfit3` branch into `master` at some point in the future, but not until closer to the Xcode 8 release date.

PR #1412, opened 2016-08-16
Xcode 8 Beta 6 compatibility

littleowl:
Brings Swift 3 into balance with Xcode Beta 6.
All Tests pass.
Also, brings back iOS 8 compatibility via @availability for streaming functions and properties.


cnoon:
Thanks for putting this together @littleowl. We've completely the port to beta 6 in #1413. Additionally, we won't be modifying the deployment targets for the Alamofire 4.0.0 release.

Cheers. 🍻

PR #1419, opened 2016-08-22
New Error Handling

jshier:
This PR refactors Alamofire's error handling to be based around a new `AFError` type rather than `NSError`, as Swift 3 has pushed the system frameworks and consumers toward using `Error` for everything. 

`AFError` consists of a few base cases with reason types for each. Convenience booleans and other properties have been added to make interacting with type easier in end use scenarios.

This PR is a works in progress as the inline and separate documentation still needs to be completed and the tests need to be updated for the new error type.


jshier:
@cnoon One unanswered question to also consider is whether the response serialization errors should wrap the underlying decoding errors (e.g. the `JSONSerialization` and `PropertyListSerialization` errors).

Source/MultipartFormData.swift
```diff
@@ -232,19 +231,22 @@ open class MultipartFormData {
...
         //              Check 2 - is file URL reachable?
         //============================================================
-
-        let isReachable = (fileURL as NSURL).checkPromisedItemIsReachableAndReturnError(nil)
-
-        guard isReachable else {
-            setBodyPartError(withCode: NSURLErrorBadURL, failureReason: "The file URL is not reachable: \(fileURL)")
+        
+        do {
+            let isReachable = try fileURL.checkPromisedItemIsReachable()
```
cnoon: Newline?

Source/ResponseSerialization.swift
```diff
@@ -32,31 +32,31 @@ public protocol ResponseSerializerType {
     associatedtype SerializedObject
 
     /// The type of error to be created by this `ResponseSerializer` if serialization fails.
-    associatedtype ErrorObject: Error
+//    associatedtype ErrorObject: Error
```
cnoon: Can probably revert this now.

Source/SessionDelegate.swift
```diff
@@ -29,94 +29,94 @@ open class SessionDelegate: NSObject {
...
+    /// Overrides default behavior for URLSessionStreamDelegate method `urlSession(_:writeClosedFor:)`.
     open var streamTaskWriteClosed: ((URLSession, URLSessionStreamTask) -> Void)?
 
-    /// Overrides default behavior for NSURLSessionStreamDelegate method `URLSession:betterRouteDiscoveredForStreamTask:`.
+    /// Overrides default behavior for URLSessionStreamDelegate method `urlSession(_:betterRouteDiscoveredFor:)`.
     open var streamTaskBetterRouteDiscovered: ((URLSession, URLSessionStreamTask) -> Void)?
 
-    /// Overrides default behavior for NSURLSessionStreamDelegate method `URLSession:streamTask:didBecomeInputStream:outputStream:`.
+    /// Overrides default behavior for URLSessionStreamDelegate method `urlSession(_:streamTask:didBecome:outputStream:)`.
     open var streamTaskDidBecomeInputStream: ((URLSession, URLSessionStreamTask, InputStream, OutputStream) -> Void)?
```
cnoon: Nice! That had to take a while to fix. 😕

Source/TaskDelegate.swift
```diff
@@ -137,7 +137,8 @@ open class TaskDelegate: NSObject {
         completionHandler(bodyStream)
     }
 
-    func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: NSError?) {
+    @objc(URLSession:task:didCompleteWithError:)
```
cnoon: Are these really necessary? Same with the redirection API above.

Source/Validation.swift
```diff
@@ -72,22 +72,8 @@ extension Request {
...
-                    userInfo: [
-                        NSLocalizedFailureReasonErrorKey: failureReason,
-                        ErrorUserInfoKeys.StatusCode: response.statusCode
-                    ]
-                )
-
-                return .failure(error)
-            }
+            return acceptableStatusCode.contains(response.statusCode) ?
+                .success : .failure(AFError.responseValidationFailed(reason: .unacceptableStatusCode(code: response.statusCode)))
```
cnoon: This is long enough I'd probably just use an `if` / `else` condition rather than a multi-line ternary. Thoughts?

Source/Validation.swift
```diff
@@ -133,45 +119,27 @@ extension Request {
...
-                for contentType in acceptableContentTypes {
-                    if let acceptableMIMEType = MIMEType(contentType), acceptableMIMEType.matches(responseMIMEType) {
-                        return .success
-                    }
-                }
-            } else {
+            
+            guard
+                let responseContentType = response.mimeType,
+                let responseMIMEType = MIMEType(responseContentType) else {
```
cnoon: Could we pull the `else` down on the next line?

```swift
guard
    let responseContentType = response.mimeType,
    let responseMIMEType = MIMEType(responseContentType) 
else {
```

Source/Validation.swift
```diff
@@ -133,45 +119,27 @@ extension Request {
...
-                domain: ErrorDomain,
-                code: ErrorCode.contentTypeValidationFailed.rawValue,
-                userInfo: [
-                    NSLocalizedFailureReasonErrorKey: failureReason,
-                    ErrorUserInfoKeys.ContentType: contentType
-                ]
-            )
-
+            
+            let error = AFError.responseValidationFailed(reason: .contentTypeMismatch(acceptable: acceptableContentTypes as! [String], actual: responseContentType))
```
cnoon: This is long enough, can we break this up into a couple lines? Also, is the `as!` guaranteed to be safe here?

Source/Error.swift
```diff
@@ -24,41 +24,428 @@
...
+        case contentTypeMissingAndNotWildcard
+        case unacceptableStatusCode(code: Int)
+    }
+    
+    case multipartEncodingFailed(reason: MultipartEncodingFailureReason)
+    case inputStreamReadFailed(inputStream: InputStream)
+    case outputStreamWriteFailed(reason: OutputStreamFailureReason)
+    case responseValidationFailed(reason: ValidationFailureReason)
+    case responseSerializationFailed(reason: SerializationFailureReason)
+}
```
cnoon: I think this is an AWESOME start @jshier. I spent some time working through an alternate version of these with some different groupings.

```swift
public enum AFError: Error {
    public enum MultipartEncodingFailureReason {
        case bodyPartNotFile(at: URL)
        case bodyPartFilenameInvalid(at: URL)
        case bodyPartFileNotReachable(at: URL)
        case bodyPartFileIsDirectory(at: URL) // should probably lump together with `.bodyPartNotFile`
        case bodyPartAttributesNotAvailable(at: URL)
        case bodyPartInputStreamCreationFailed(at: URL)

        case outputStreamCreationFailed(at: URL)
        case outputStreamFileAlreadyExists(at: URL)
        case outputStreamFileIsInvalid(at: URL)
        case outputStreamWriteFailed(for: OutputStream)
    }

    public enum ValidationFailureReason {
        case missingContentType(accepted: [String])
        case unacceptableContentType(accepted: [String], response: String)
        case unacceptableStatusCode(code: Int)
    }

    public enum SerializationFailureReason {
        case inputDataNil
        case inputDataZeroLength
        case stringSerializationFailed(encoding: String.Encoding)
        case jsonSerializationFailed(error: NSError)
        case propertyListSerializationFailed(error: NSError)
    }

    case requestFailed(error: NSError) // Not sure if we want to wrap all NSURLSession errors (seems like a good idea)
    case multipartEncodingFailed(reason: MultipartEncodingFailureReason)
    case validationFailed(reason: ValidationFailureReason)
    case serializationFailed(reason: SerializationFailureReason)
}
```

Would love to see what you think here. I made the `MultipartEncoding` errors MUCH more specific and lumped the `.multipartEncodingFailed`, `.inputStreamReadFailed` and `.outputStreamWriteFailed` errors all into one. They don't really stand on their own IMO.

Then I reworked the other two a bit. I think we should definitely wrap JSON and plist serialization errors. I even think we should seriously consider wrapping all request / network errors inside an `AFError` as well. That would make it MUCH easier for the consumer to switch on the overall error type. Most consumers won't probably need to dig into the reason, but could leverage the top four buckets.

Source/Error.swift
```diff
@@ -24,41 +24,428 @@
...
+    
+    var isMultipartFailedToFetchAttributesError: Bool {
+        guard case let .multipartEncodingFailed(reason) = self else { return false }
+        return reason.isFailedToFetchAttributes
+    }
+    
+    var isMultipartFailedToCreateInputStreamError: Bool {
+        guard case let .multipartEncodingFailed(reason) = self else { return false }
+        return reason.isFailedToCreateInputStream
+    }
```
cnoon: I'm on the fence with these properties. I like the ones for the top level cases, but I question creating properties for every `XXXFailureReason`. It seems a bit excessive. Are we really thinking that people won't want to use switches? These are just so verbose it takes a while to figure out what one you'd even want to call.

cnoon:
@jshier...I'd say there's a couple open questions about the errors themselves. What to do with JSON and plist errors and also what to do with request errors. In both cases, I'm leaning towards wrapping them up in an `AFError`. It makes it MUCH easier for the consumer to switch on the top-level types of errors without have to switch between `AFError` and `NSError`. If they really need to drop down another level, they easily can.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+        case jsonSerializationFailed(error: Error)
+        case propertyListSerializationFailed(error: Error)
+    }
+
     case multipartEncodingFailed(reason: MultipartEncodingFailureReason)
-    case inputStreamReadFailed(inputStream: InputStream)
-    case outputStreamWriteFailed(reason: OutputStreamFailureReason)
     case responseValidationFailed(reason: ValidationFailureReason)
     case responseSerializationFailed(reason: SerializationFailureReason)
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+    var isUnacceptableContentType: Bool {
+        if case let .responseValidationFailed(reason) = self, reason.isUnacceptableContentType { return true }
+        return false
+    }
+
+    var isUnacceptableStatusCode: Bool {
+        if case let .responseValidationFailed(reason) = self, reason.isUnacceptableStatusCode { return true }
+        return false
     }
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
-    
-    var isOutputStreamWriteFailedError: Bool {
-        guard case let .outputStreamWriteFailed(reason) = self else { return false }
-        return reason.isWriteFailed
+
+}
+
+extension AFError {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
-    
-    var isUnacceptableStatusCode: Bool {
-        if case .contentTypeMismatch = self { return true }
+
+    var isInputStreamReadFailed: Bool {
+        if case .inputStreamReadFailed = self { return true }
         return false
     }
-    
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
 }
 
 extension AFError.SerializationFailureReason {
-    
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+    var isJSONSerializationFailed: Bool {
+        if case .jsonSerializationFailed = self { return true }
+        return false
+    }
+
+    var isPropertyListSerializationFailed: Bool {
+        if case .propertyListSerializationFailed = self { return true }
+        return false
+    }
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+}
+
+extension AFError.ValidationFailureReason {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+    var isUnacceptableContentType: Bool {
+        if case .unacceptableContentType = self { return true }
+        return false
+    }
+
+    var isUnacceptableStatusCode: Bool {
+        if case .unacceptableStatusCode = self { return true }
+        return false
+    }
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
 }
 
+// MARK: Convenience Properties
 
-// MARK: Associated Value Properties
+public extension AFError {
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+    /// The `String.Encoding` associated with a failed `.stringResponse()` call.
+    public var failedStringEncoding: String.Encoding? {
+        switch self {
+        case .responseSerializationFailed(let reason):
+            return reason.failedStringEncoding
+        default:
+            return nil
+        }
+    }
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
 }
 
-extension AFError.OutputStreamFailureReason {
-    
+extension AFError.MultipartEncodingFailureReason {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
         }
     }
-    
-    var outputStream: OutputStream? {
-        guard case let .writeFailed(outputStream) = self else { return nil }
-        
-        return outputStream
+
+    var underlyingError: Error? {
+        switch self{
```
cnoon: Need space before `{`.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+        case .bodyPartFileNotReachableWithError(_, let error), .bodyPartFileSizeQueryFailedWithError(_, let error),
+             .outputStreamWriteFailed(let error), .inputStreamReadFailed(let error):
+            return error
+        default:
+            return nil
+        }
     }
-    
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
 }
 
 extension AFError.ValidationFailureReason {
-    
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+    var responseCode: Int? {
+        switch self {
+        case .unacceptableStatusCode(let code):
+            return code
+        default:
+            return nil
+        }
+    }
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
 }
 
 extension AFError.SerializationFailureReason {
-    
-    var stringEncoding: String.Encoding? {
-        guard case let .invalidStringEncoding(encoding) = self else { return nil }
-        
-        return encoding
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+    var underlyingError: Error? {
+        switch self {
+        case .jsonSerializationFailed(let error), .propertyListSerializationFailed(let error):
+            return error
+        default:
+            return nil
+        }
     }
-    
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
 }
 
 // MARK: Error Descriptions
 
 extension AFError: LocalizedError {
-    
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
     public var errorDescription: String? {
```
cnoon: Missing docstring.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+        switch self {
+        case .multipartEncodingFailed(let reason):
+            return reason.localizedDescription
+        case .responseValidationFailed(let reason):
+            return reason.localizedDescription
+        case .responseSerializationFailed(let reason):
+            return reason.localizedDescription
+        }
+    }
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+}
+
+extension AFError.SerializationFailureReason {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+            return "Response could not be serialized, input data was nil or zero length."
+        case .stringSerializationFailed(let encoding):
+            return "String could not be serialized with encoding: \(encoding)."
+        case .jsonSerializationFailed(let error):
+            return "JSON could not be serialized because of error:\n\(error.localizedDescription)"
+        case .propertyListSerializationFailed(let error):
+            return "PropertyList could not be serialized because of error:\n\(error.localizedDescription)"
+        }
+    }
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+}
+
+extension AFError.ValidationFailureReason {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+        case .missingContentType(let types):
+            return "Response Content-Type was missing and acceptable content types (\(types.joined(separator: ","))) do not match \"*/*\"."
+        case .unacceptableContentType(let acceptableTypes, let responseType):
+            return "Response Content-Type \"\(responseType)\" does not match any acceptable types: \(acceptableTypes.joined(separator: ","))."
+        case .unacceptableStatusCode(let code):
+            return "Response status code was unacceptable: \(code)."
+        }
     }
-    
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
-//let failureReason = "Could not fetch attributes from the file URL: \(fileURL)"
-//let failureReason = "Failed to create an input stream from the file URL: \(fileURL)"
-//                let failureReason = "Response status code was unacceptable: \(response.statusCode)"
-//                failureReason = (
-//                    "Response content type \"\(responseContentType)\" does not match any acceptable " +
-//                    "content types: \(acceptableContentTypes)"
-//                )
-//failureReason = "Response content type was missing and acceptable content type does not match \"*/*\""
+extension AFError.MultipartEncodingFailureReason {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,428 +24,540 @@
...
+            return "A file already exists at the provided URL: \(url)"
+        case .outputStreamURLInvalid(let url):
+            return "The provided OutputStream URL is invalid: \(url)"
+        case .outputStreamWriteFailed(let error):
+            return "OutputStream write failed with error: \(error)"
+        case .inputStreamReadFailed(let error):
+            return "InputStream read failed with error: \(error)"
+        }
+    }
+
```
cnoon: Extra newline.

Source/MultipartFormData.swift
```diff
@@ -441,36 +433,26 @@ open class MultipartFormData {
...
-                break
+            if let error = inputStream.streamError {
+                throw AFError.multipartEncodingFailed(reason: .inputStreamReadFailed(error: error))
             }
 
             if bytesRead > 0 {
                 encoded.append(buffer, count: bytesRead)
-            } else if bytesRead < 0 {
-                error = AFError.inputStreamReadFailed(inputStream: inputStream)
-                break
```
cnoon: Hmmm...how do we know it's safe to remove this check?

Source/MultipartFormData.swift
```diff
@@ -510,14 +494,10 @@ open class MultipartFormData {
                 }
 
                 try write(&buffer, to: outputStream)
-            } else if bytesRead < 0 {
-                throw AFError.inputStreamReadFailed(inputStream: inputStream)
```
cnoon: Hmmm...how do we know it's safe to remove this check?

Source/MultipartFormData.swift
```diff
@@ -542,12 +522,8 @@ open class MultipartFormData {
             if outputStream.hasSpaceAvailable {
                 let bytesWritten = outputStream.write(buffer, maxLength: bytesToWrite)
 
-                if let streamError = outputStream.streamError {
-                    throw streamError
-                }
-
-                if bytesWritten < 0 {
-                    throw AFError.outputStreamWriteFailed(reason: .writeFailed(to: outputStream))
```
cnoon: Hmmm...how do we know it's safe to remove this check?

Source/ParameterEncoding.swift
```diff
@@ -220,9 +220,7 @@ public enum ParameterEncoding {
         let generalDelimitersToEncode = ":#[]@" // does not include "?" or "/" due to RFC 3986 - Section 3.4
         let subDelimitersToEncode = "!$&'()*+,;="
 
-        // rdar://26850776
-        // Crash in Xcode 8 Seed 1 when trying to mutate a CharacterSet with remove
-        var allowedCharacterSet = NSMutableCharacterSet.urlQueryAllowed
+        var allowedCharacterSet = CharacterSet.urlQueryAllowed
         allowedCharacterSet.remove(charactersIn: "\(generalDelimitersToEncode)\(subDelimitersToEncode)")
```
cnoon: We should figure out what characters are different in Apple's `urlQueryAllowed` character set than ours and make sure the logic we have here is exactly what we want. Doesn't need to be in this PR though.

Tests/MultipartFormDataTests.swift
```diff
@@ -22,6 +22,7 @@
 //  THE SOFTWARE.
 //
 
+@testable
 import Alamofire
```
cnoon: How about?

```swift
@testable import Alamofire
import Foundation
import XCTest
```

Tests/ResponseSerializationTests.swift
```diff
@@ -22,15 +22,15 @@
 //  THE SOFTWARE.
 //
 
+@testable
 import Alamofire
```
cnoon: Same here.

Tests/UploadTests.swift
```diff
@@ -631,7 +631,7 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
 //            }
 //        )
 //
-//        waitForExpectations(withTimeout: timeout, handler: nil)
+//        waitForExpectations(timeout: timeout, handler: nil)
```
cnoon: Thanks for fixing all these even though they're commented out! 🍻

Tests/ValidationTests.swift
```diff
@@ -176,10 +174,10 @@ class ContentTypeValidationTestCase: BaseTestCase {
...
-            XCTAssertEqual(error.domain, ErrorDomain)
-            XCTAssertEqual(error.code, ErrorCode.contentTypeValidationFailed.rawValue)
-            XCTAssertEqual(error.userInfo[ErrorUserInfoKeys.ContentType] as? String, "application/xml")
+        if let error = error as? AFError, let contentType = error.responseContentType, let acceptableTypes = error.acceptableContentTypes {
+            XCTAssertTrue(error.isUnacceptableContentType)
+            XCTAssertEqual(contentType, "application/xml")
+            XCTAssertEqual(acceptableTypes[0], "application/octet-stream")
         } else {
             XCTFail("error should not be nil")
         }
```
cnoon: This will crash if `acceptableTypes` is empty. Could we get a check for the index as well as blow out the `XCTFail` explanation to explain the other cases where this could fail?

Tests/ValidationTests.swift
```diff
@@ -205,10 +203,10 @@ class ContentTypeValidationTestCase: BaseTestCase {
...
-        if let error = error {
-            XCTAssertEqual(error.domain, ErrorDomain)
-            XCTAssertEqual(error.code, ErrorCode.contentTypeValidationFailed.rawValue)
-            XCTAssertEqual(error.userInfo[ErrorUserInfoKeys.ContentType] as? String, "application/xml")
+        if let error = error as? AFError, let contentType = error.responseContentType, let acceptableTypes = error.acceptableContentTypes {
+            XCTAssertTrue(error.isUnacceptableContentType)
+            XCTAssertEqual(contentType, "application/xml")
+            XCTAssertTrue(acceptableTypes.isEmpty)
         } else {
             XCTFail("error should not be nil")
```
cnoon: Fail does not explain all cases.

Tests/ValidationTests.swift
```diff
@@ -360,10 +358,9 @@ class MultipleValidationTestCase: BaseTestCase {
...
-        if let error = error {
-            XCTAssertEqual(error.domain, ErrorDomain)
-            XCTAssertEqual(error.code, ErrorCode.statusCodeValidationFailed.rawValue)
-            XCTAssertEqual(error.userInfo[ErrorUserInfoKeys.StatusCode] as? Int, 200)
+        if let error = error as? AFError, let statusCode = error.responseCode {
+            XCTAssertTrue(error.isUnacceptableStatusCode)
+            XCTAssertEqual(statusCode, 200)
         } else {
             XCTFail("error should not be nil")
         }
```
cnoon: We should either make the fail more descriptive, or check the status like this:

```swift
XCTAssertEqual(error.responseCode, 200)
```

I'd recommend using the check I just put above.

Tests/ValidationTests.swift
```diff
@@ -390,10 +387,10 @@ class MultipleValidationTestCase: BaseTestCase {
...
-            XCTAssertEqual(error.domain, ErrorDomain)
-            XCTAssertEqual(error.code, ErrorCode.contentTypeValidationFailed.rawValue)
-            XCTAssertEqual(error.userInfo[ErrorUserInfoKeys.ContentType] as? String, "application/xml")
+        if let error = error as? AFError, let contentType = error.responseContentType, let acceptableTypes = error.acceptableContentTypes {
+            XCTAssertTrue(error.isUnacceptableContentType)
+            XCTAssertEqual(contentType, "application/xml")
+            XCTAssertEqual(acceptableTypes[0], "application/octet-stream")
         } else {
             XCTFail("error should not be nil")
         }
```
cnoon: Same comment as before.

Tests/ValidationTests.swift
```diff
@@ -447,10 +444,9 @@ class AutomaticValidationTestCase: BaseTestCase {
         // Then
         XCTAssertNotNil(error)
 
-        if let error = error {
-            XCTAssertEqual(error.domain, ErrorDomain)
-            XCTAssertEqual(error.code, ErrorCode.statusCodeValidationFailed.rawValue)
-            XCTAssertEqual(error.userInfo[ErrorUserInfoKeys.StatusCode] as? Int, 404)
+        if let error = error as? AFError, let statusCode = error.responseCode {
+            XCTAssertTrue(error.isUnacceptableStatusCode)
+            XCTAssertEqual(statusCode, 404)
```
cnoon: Same comment as before.

Tests/ValidationTests.swift
```diff
@@ -529,10 +525,10 @@ class AutomaticValidationTestCase: BaseTestCase {
...
         XCTAssertNotNil(error)
 
-        if let error = error {
-            XCTAssertEqual(error.domain, ErrorDomain)
-            XCTAssertEqual(error.code, ErrorCode.contentTypeValidationFailed.rawValue)
-            XCTAssertEqual(error.userInfo[ErrorUserInfoKeys.ContentType] as? String, "application/xml")
+        if let error = error as? AFError, let contentType = error.responseContentType, let acceptableTypes = error.acceptableContentTypes {
+            XCTAssertTrue(error.isUnacceptableContentType)
+            XCTAssertEqual(contentType, "application/xml")
+            XCTAssertEqual(acceptableTypes[0], "application/json")
```
cnoon: Same comment as before.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
-    case jsonSerializationFailed         = -6006
-    case propertyListSerializationFailed = -6007
+/// `AFError` is the error type returned by the Alamofire framework. It encompasses a few
+/// different types of errors, each with their own associated reasons.
+///
+/// `.multipartEncodingFailed` errors are returned when some step in the multipart encoding process fails.
+///
+/// `.responseValidationFailed` errors are returned when a `validate()` call fails.
+///
+/// `.responseSerializationFailed` errors are returned when a response serializer encounters an error in the serialization process.
```
cnoon: Why no `- ` before each case? Also, let's use multiple lines here when the docstring gets too long. For example, the `. responseSerializationFailed` is probably well over that limit.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+public enum AFError: Error {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+        case bodyPartFileSizeNotAvailable(at: URL)
+        case bodyPartFileSizeQueryFailedWithError(url: URL, error: Error)
+        case bodyPartInputStreamCreationFailed(for: URL)
+
+        case outputStreamCreationFailed(for: URL)
+        case outputStreamFileAlreadyExists(at: URL)
+        case outputStreamURLInvalid(url: URL)
+        case outputStreamWriteFailed(error: Error)
+
+        case inputStreamReadFailed(error: Error)
```
cnoon: The naming convention of the associated values isn't consistent throughout these cases. Some use prepositions and some use nouns. Doesn't seem to be a convention either. IMO these should always use the noun form, but I'm not sure what the Apple guidelines recommend here. Since there aren't external parameters, it seems like they should always use the internal parameter names. Otherwise you end up with really odd internal parameter names.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+    }
+
+    public enum ValidationFailureReason {
+        case missingContentType(acceptable: [String])
+        case unacceptableContentType(acceptable: [String], response: String)
```
cnoon: These could both probably benefit from more verbosity. What about `acceptableContentTypes` and `responseContentType`?

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+        case inputDataNilOrZeroLength
+        case stringSerializationFailed(encoding: String.Encoding)
+        case jsonSerializationFailed(error: Error)
+        case propertyListSerializationFailed(error: Error)
+    }
+
+    case multipartEncodingFailed(reason: MultipartEncodingFailureReason)
+    case responseValidationFailed(reason: ValidationFailureReason)
+    case responseSerializationFailed(reason: SerializationFailureReason)
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+        case unacceptableStatusCode(code: Int)
+    }
+
+    public enum SerializationFailureReason {
+        case inputDataNil
+        case inputDataNilOrZeroLength
+        case stringSerializationFailed(encoding: String.Encoding)
+        case jsonSerializationFailed(error: Error)
+        case propertyListSerializationFailed(error: Error)
+    }
```
cnoon: Need docstrings on `MultipartEncodingFailureReason`, ValidationFailureReason` and `SerializationFailureReason`.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+    case multipartEncodingFailed(reason: MultipartEncodingFailureReason)
+    case responseValidationFailed(reason: ValidationFailureReason)
+    case responseSerializationFailed(reason: SerializationFailureReason)
+
+}
+
+// MARK: Error Booleans
+
+public extension AFError {
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+        return false
+    }
+
+    /// Returns whether the `AFError` is a response serialization error.
+    /// When true, the `failedStringEncoding` and `underlyingError` properties will contain the associated values.
+    public var isResponseSerializationError: Bool {
+        if case .responseSerializationFailed = self { return true }
+        return false
+    }
+
```
cnoon: Extra newline.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+    /// Returns whether the `AFError` is a response validation error.
+    /// When true, the `acceptableContentTypes`, `responseContentType`, and `responseCode` properties will contain the associated values.
+    public var isResponseValidationError: Bool {
+        if case .responseValidationFailed = self { return true }
+        return false
+    }
+
+    /// Returns whether the `AFError` is a response serialization error.
+    /// When true, the `failedStringEncoding` and `underlyingError` properties will contain the associated values.
```
cnoon: Why'd you start putting different sentences on different lines? Everywhere else in AF we would do the following:

```
/// Returns whether the `AFError` is a response serialization error. When true, the `failedStringEncoding` and 
/// `underlyingError` properties will contain the associated values.
```

They just go out the 120 character limit then pull down to the next line. Reasoning for this is to make it easier to review so you don't have to scroll left-to-right throughout the diff in GitHub.

Same comment applies to the two docstrings above this one as well.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+        if case .unacceptableContentType = self { return true }
+        return false
+    }
+
+    var isUnacceptableStatusCode: Bool {
+        if case .unacceptableStatusCode = self { return true }
+        return false
+    }
+
+}
```
cnoon: I think we should move all these Boolean checks into an extension that is only part of the test target. I don't see us exposing these as public any time soon. 

Why I think it would be wise to move these out of the `Error.swift` file is that they greatly convolute the jumpbar and the file itself. There are so many APIs available that it's difficult to tell what I can and can't use right now. If we pull half the APIs out into the tests, it will greatly simplify the readability of this file.

We could move all these into an `Error+AlamofireTests.swift` file. Then we could always move them back if necessary.

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+        case propertyListSerializationFailed(error: Error)
+    }
+
+    case multipartEncodingFailed(reason: MultipartEncodingFailureReason)
+    case responseValidationFailed(reason: ValidationFailureReason)
+    case responseSerializationFailed(reason: SerializationFailureReason)
+
+}
+
+// MARK: Error Booleans
```
cnoon: How about?

```swift
// MARK: - Error Booleans
```

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+    }
+
+    var isUnacceptableStatusCode: Bool {
+        if case .unacceptableStatusCode = self { return true }
+        return false
+    }
+
+}
+
+// MARK: Convenience Properties
```
cnoon: How about?

```swift
// MARK: - Convenience Properties
```

Source/Error.swift
```diff
@@ -24,41 +24,540 @@
...
+            return nil
+        }
+    }
 
-    /// The status code user info key for a `.StatusCodeValidationFailed` error stored as an `Int` value.
-    public static let StatusCode = "StatusCode"
 }
 
-// MARK: -
+// MARK: Error Descriptions
```
cnoon: How about?

```swift
// MARK: - Error Descriptions
```

Source/TaskDelegate.swift
jshier: Yes, unfortunately. I think it's a byproduct of the various subclasses protocol requirements not propagating the Swiftified version of these APIs up the hierarchy. I believe @kcharwood filed a bug about it when he did the initial Swift 3 port at WWDC. @kcharwood ?

Source/Error.swift
jshier: The docstring is provided by the protocol. If you option click this function you'll see it. Unless you want us to provide our own?

screen shot 2016-08-28 at 4 35 56 pm

jshier: I have a feeling this method name is going to change in the next beta, as I think it's supposed to be `localizedDescription`.

Source/MultipartFormData.swift
jshier: It's documented to be so. From the `InputStream` docs:
```
Return Value
A number indicating the outcome of the operation:

A positive number indicates the number of bytes read.

0 indicates that the end of the buffer was reached.

-1 means that the operation failed; more information about the error can be obtained with streamError.
```
If we want to keep this check, we'll need another error reason, as we'll catch any error as it is.

Source/MultipartFormData.swift
jshier: It's documented. See my other response.

Source/MultipartFormData.swift
jshier: Same as the others.

Tests/ValidationTests.swift
jshier: That won't work, as `XCTAssertEqual` doesn't work on optional and non-optional values.

jshier: I'll make the failure messages more descriptive and see what you think.

Source/Error.swift
jshier: Unfortunately enum cases don't offer any flexibility with associated value capture labels, otherwise some of those incongruous names would be elided and the case name would handle it. e.g. for `bodyPartURLInvalid` or the `*FailedWithError` cases, I wouldn't have a label at all, but I have to. I tried to make the cases read as grammatical phrases as much as possible while still maintaining uniqueness, but it wasn't possible in all cases. Plus there are some additional cases I had to handle that weren't in your previous rename. I added the `*FailedWithError` cases because you can't have duplicate case names that capture different values. Otherwise, like `outputStreamWriteFailed`, they wouldn't have that additional text. Unfortunately the Swift 3 naming guidelines don't offer any specific guidance for enum case names.

Source/Error.swift
jshier: Would this include documentation of all of the individual cases?

Source/Error.swift
jshier: I'll audit for 120 characters wherever possible.

Source/Error.swift
jshier: Added all docstrings.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Tests/MultipartFormDataTests.swift
jshier: Removed the `@testable`, as I moved the boolean extensions into the test target.

Tests/ResponseSerializationTests.swift
jshier: Removed the `@testable`, as I moved the boolean extensions into the test target.

Tests/UploadTests.swift
jshier: Unfortunately they're still broken.

Tests/ValidationTests.swift
jshier: I moved the acceptable type into the let statements and updated all of the failure descriptions to describe what the if statement should be unwrapping.

Tests/ValidationTests.swift
jshier: Fixed.

Tests/ValidationTests.swift
jshier: Updated the failure message, let me know what you think.

Tests/ValidationTests.swift
jshier: Updated the failure message, let me know what you think.

Tests/ValidationTests.swift
jshier: Updated the failure message, let me know what you think.

Source/Error.swift
jshier: Added, and updated to use a 120 character column.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Added verbosity.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Fixed.

Source/Error.swift
jshier: Moved into a testing extension, `AFError+AlamofireTests.swift`.

Source/Error.swift
jshier: Added.

Source/Error.swift
jshier: Added.

Source/Error.swift
jshier: Added.

Source/Error.swift
cnoon: Oh awesome! I didn't realize that if you document the protocol that you don't need to document the conformance. Good stuff.

Tests/MultipartFormDataTests.swift
cnoon: Even better!

Tests/ValidationTests.swift
cnoon: You can definitely compare optional and non-optional types using `XCTAssertEqual`. I do it all the time. I'll move this through for now and make a note to update it.

Source/Error.swift
cnoon: Yep, I completely get why you did what you did. What I'm wondering is whether we want to use grammatical naming in "most" cases, but not all. We could ignore the grammatical convention, and instead use the noun form of the internal parameter. Instead of `in`, `at` and `for` for a URL`, they would all just be `url`. I'm going to move this through as is, but I think we may still need to revisit this a bit before we lock 4.0. It also doesn't help that Apple hasn't provided any guidance in this situation. 😕

PR #1420, opened 2016-08-22
User-Agent header using build number instead of version number

antrix1989:
The User-Agent header is grabbing the infoDictionary object for the kCFBundleVersionKey key which is the app build number. It should be grabbing the object for the CFBundleShortVersionString key which is the actual version number. Same issue was fixed in AFNetworking, [issue 1907.](https://github.com/AFNetworking/AFNetworking/issues/1907)

### DIFF
--- a/Source/Manager.swift
+++ b/Source/Manager.swift
@@ -60,7 +60,7 @@ public class Manager {
             if let info = NSBundle.mainBundle().infoDictionary {
                 let executable = info[kCFBundleExecutableKey as String] as? String ?? "Unknown"
                 let bundle = info[kCFBundleIdentifierKey as String] as? String ?? "Unknown"
-                let version = info[kCFBundleVersionKey as String] as? String ?? "Unknown"
+                let version = info["CFBundleShortVersionString"] as? String ?? "Unknown"
 
                 let osNameVersion: String = {
                     let versionString: String


jshier:
This is intentionally the build number, though I guess the variable name is incorrect. The actual app version is built in the later lines, where we query `NSProcessInfo`.

jshier:
@antrix1989 Looking at it again, you're correct. However, both the app version and build number are useful bits of information, so they should likely both be in the user agent. Something like this:
```swift
    let userAgent: String = {
            if let info = NSBundle.mainBundle().infoDictionary {
                let executable = info[kCFBundleExecutableKey as String] as? String ?? "Unknown"
                let bundle = info[kCFBundleIdentifierKey as String] as? String ?? "Unknown"
                let version = info["CFBundleShortVersionString"] as? String ?? "Unknown"
                let build = info[kCFBundleVersionKey as String] as? String ?? "Unknown"
                
                let osNameVersion: String = {
                    let version = NSProcessInfo.processInfo().operatingSystemVersion
                    let versionString = "\(version.majorVersion).\(version.minorVersion).\(version.patchVersion)"
                    
                    let osName: String = {
                        #if os(iOS)
                            return "iOS"
                        #elseif os(watchOS)
                            return "watchOS"
                        #elseif os(tvOS)
                            return "tvOS"
                        #elseif os(OSX)
                            return "OS X"
                        #elseif os(Linux)
                            return "Linux"
                        #else
                            return "Unknown"
                        #endif
                    }()
                    
                    return "\(osName) \(versionString)"
                }()
                
                return "\(executable)/\(bundle) (\(version)/\(build); \(osNameVersion))"
            }
            
            return "Alamofire"
        }()
```

jshier:
Thanks for the changes @antrix1989! We made some small modifications to your PR and pushed them into `master` in commit [`31b882`](https://github.com/Alamofire/Alamofire/commit/e84a9aae9d6c2b9db68268c51102ad3b0031b882) while maintaining your attribution. Thanks again!

antrix1989:
Great! Thanks for keeping my commit :)

ekullhu:
I'm a total newbie to contributing pull requests etc. so not sure how to do this myself, but there is a bug in this commit
`return "\(executable)/\(bundle) (\(appVersion)/\(appBuild)); \(osNameVersion))"`
will return un balanced brackets i.e.: 
`app/com.app (4.0/382); iOS 10.0.0)`
which is freaking out our server parsing the user agent string.

PR #1450, opened 2016-09-04
Feature - Request Adapter and Retrier System

cnoon:
This PR adds a couple powerful protocols, `RequestAdapter` and `RequestRetrier`, that open up all sorts of possibilities including:

* Easing `Request` modifications for `Authorization` headers
* Retrying requests that encountered an error with a custom exponential backoff retry policy
* Completely thread-safe refresh systems for web services behind OAuth2 authentication

I'm sure there are all sorts of other cool things you could do with this, but these are the main things I had in mind when putting this together.

## RequestAdapter

The `RequestAdapter` protocol allows each `Request` made on a `SessionManager` to be inspected and adapted before being created. One very specific way to use an adapter is to append an `Authorization` header to requests behind a certain type of authentication.

```swift
class AccessTokenAdapter: RequestAdapter {
	private let accessToken: String
	
	init(accessToken: String) {
		self.accessToken = accessToken
	}

	func adapt(_ urlRequest: URLRequest) -> URLRequest {
	    var urlRequest = urlRequest

	    if urlRequest.urlString.hasPrefix("https://httpbin.org") {
		    urlRequest.setValue("Bearer " + accessToken, forHTTPHeaderField: "Authorization")
	    }

	    return urlRequest
	}
}

let sessionManager = SessionManager()
sessionManager.adapter = AccessTokenAdapter(accessToken: "1234")

sessionManager.request("https://httpbin.org/get", withMethod: .get)
```

## RequestRetrier

The `RequestRetrier` protocol allows a `Request` that encountered an `Error` while being executed to be retried. When using both the `RequestAdapter` and `RequestRetrier` protocols together, you can create credential refresh systems for OAuth1, OAuth2, Basic Auth and even exponential backoff retry policies. The possibilities are endless. Here's a short example of how you could implement a refresh flow for OAuth2 access tokens.

```swift
class OAuth2Handler: RequestAdapter, RequestRetrier {
    private typealias RefreshCompletion = (_ succeeded: Bool, _ accessToken: String?, _ refreshToken: String?) -> Void

    private let sessionManager: SessionManager = {
        let configuration = URLSessionConfiguration.default
        configuration.httpAdditionalHeaders = SessionManager.defaultHTTPHeaders

        return SessionManager(configuration: configuration)
    }()

    private let lock = NSLock()

    private var clientID: String
    private var baseURLString: String
    private var accessToken: String
    private var refreshToken: String

    private var isRefreshing = false
    private var requestsToRetry: [RequestRetryCompletion] = []

    // MARK: - Initialization

    public init(clientID: String, baseURLString: String, accessToken: String, refreshToken: String) {
        self.clientID = clientID
        self.baseURLString = baseURLString
        self.accessToken = accessToken
        self.refreshToken = refreshToken
    }

    // MARK: - RequestAdapter

    public func adapt(_ urlRequest: URLRequest) -> URLRequest {
        if urlRequest.urlString.hasPrefix(baseURLString) {
            var mutableURLRequest = urlRequest
            mutableURLRequest.setValue("Bearer " + accessToken, forHTTPHeaderField: "Authorization")
            return mutableURLRequest
        }

        return urlRequest
    }

    // MARK: - RequestRetrier

    public func should(_ manager: SessionManager, retry request: Request, with error: Error, completion: RequestRetryCompletion) {
        lock.lock() ; defer { lock.unlock() }

        if let response = request.task.response as? HTTPURLResponse, response.statusCode == 401 {
            requestsToRetry.append(completion)

            if !isRefreshing {
                refreshTokens { [weak self] succeeded, accessToken, refreshToken in
                    guard let strongSelf = self else { return }

                    strongSelf.lock.lock() ; defer { strongSelf.lock.unlock() }

                    if let accessToken = accessToken, let refreshToken = refreshToken {
                        strongSelf.accessToken = accessToken
                        strongSelf.refreshToken = refreshToken
                    }

                    strongSelf.requestsToRetry.forEach { $0(succeeded, 0.0) }
                    strongSelf.requestsToRetry.removeAll()
                }
            }
        } else {
            completion(false, 0.0)
        }
    }

    // MARK: - Private - Refresh Tokens

    private func refreshTokens(completion: RefreshCompletion) {
        guard !isRefreshing else { return }

        isRefreshing = true

        let urlString = "\(baseURLString)/oauth2/token"

        let parameters: [String: Any] = [
            "access_token": accessToken,
            "refresh_token": refreshToken,
            "client_id": clientID,
            "grant_type": "refresh_token"
        ]

        sessionManager.request(urlString, withMethod: .post, parameters: parameters, encoding: .json).responseJSON { [weak self] response in
            guard let strongSelf = self else { return }

            if let json = response.result.value as? [String: String] {
                completion(true, json["access_token"], json["refresh_token"])
            } else {
                completion(false, nil, nil)
            }

            strongSelf.isRefreshing = false
        }
    }
}

let baseURLString = "https://some.domain-behind-oauth2.com"

let oauthHandler = OAuth2Handler(
    clientID: "12345678",
    baseURLString: baseURLString,
    accessToken: "abcd1234",
    refreshToken: "ef56789a"
)

let sessionManager = SessionManager()
sessionManager.adapter = oauthHandler
sessionManager.retrier = oauthHandler

let urlString = "\(baseURLString)/some/endpoint"

manager.request(urlString, withMethod: .get).validate().responseJSON { response in
    debugPrint(response)
}
```

---

## Tests

I've added tests around the `RequestAdapter` to make sure all the `SessionManager` APIs actually  call the `adapter` if set. I also added tests verifying the `retrier` is called when errors occur and that the `adapter` will be called again when the request is retried.

## Internal Modifications

There are several internal changes that are worth calling out to make this work.

### Validations

The first is that `Validation` is no longer run on the delegate's `OperationQueue`. In order to determine whether a `Request` encountered an error, we need to make sure we run all the validations first, otherwise the error won't be set and the `retrier` won't be called. This was a trivial change to implement, but is an important callout.

### TaskConvertible

The `TaskConvertible` enum nested in the `Request` class allows us to store the un-adapted version of the `Request` before it is adapted and turned into a `URLSessionTask`. The `retry` method on the `SessionManager` uses the new `originalTask` property to extract the original urlRequest, adapt it if necessary, then create the new task and apply it to the `Request`. By setting a new task on the `TaskDelegate`, it automatically resets all the data tracked in the `TaskDelegate` as though it is a brand new task.

### SessionDelegate

The `SessionManager` must be passed down to the `SessionDelegate` as a `weak` property in order to allow the `SessionManager` to retry the `Request` from the `SessionDelegate`. I originally had the retry logic in the `Request` class, but it became problematic when trying to figure out how to lock down the task creation on the `SessionManager` queue. While I don't love having to store the parent reference in the child, it's not so bad since it's not exposed publicly and it doesn't result in a retain cycle since it's a `weak` reference. 

> I'm very open to suggestions here on a better approach, but I think this is about the cleanest way to do it.

## README Updates

I've added a `Adapting and Retrying Requests` section to the README to walk users through some of the ways you could use these protocols. I also created a fairly detailed `OAuth2` example to demonstrate how to start to build a thread-safe refresh system that could be shared across multiple session managers.

---

## Summary

Overall I think these are two very powerful protocols that will open up all sorts of possibilities for the Alamofire community. I'd love to see custom `RequestRetrier` implementations for linear and exponential back-off policies for things like background sync operations. Can't wait to see what everyone comes up with!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1110,6 +1110,175 @@ Generally, either the default implementation or the override closures should pro
 
 > It is important to keep in mind that the `subdelegates` are initialized and destroyed in the default implementation. Be careful when subclassing to not introduce memory leaks.
 
+### Adapting and Retrying Requests
+
+Most web services these days are behind some sort of authentication system. One of the more common ones today is OAuth. This generally involves generating an access token authorizing your application or user to call the various supported web services. While creating these initial access tokens can be laborsome, it can be even more complicated when your access token expires and you need to fetch a new one. There are many thread-safety issues that need to be considered.
+
+The `RequestAdapter` and `RequestRetrier` protocols were created to make it much easier to create a thread-safe authentication system for a specific set of web services.
+
+#### RequestAdapter
+
+The `RequestAdapter` protocol allows each `Request` made on a `SessionManager` to be inspected and adapted before being created. One very specific way to use an adapter is to append an `Authorization` header to requests behind a certain type of authentication.
+
+```swift
+class AccessTokenAdapter: RequestAdapter {
+	private let accessToken: String
+	
+	init(accessToken: String) {
+		self.accessToken = accessToken
+	}
+
+	func adapt(_ urlRequest: URLRequest) -> URLRequest {
+	    var urlRequest = urlRequest
+
+	    if urlRequest.urlString.hasPrefix("https://httpbin.org") {
+		    urlRequest.setValue("Bearer " + accessToken, forHTTPHeaderField: "Authorization")
+	    }
+
+	    return urlRequest
+	}
+}
+
+let sessionManager = SessionManager()
+sessionManager.adapter = AccessTokenAdapter(accessToken: "1234")
+
+sessionManager.request("https://httpbin.org/get", withMethod: .get)
+```
+
+#### RequestRetrier
+
+The `RequestRetrier` protocol allows a `Request` that encountered an `Error` while being executed to be retried. When using both the `RequestAdapter` and `RequestRetrier` protocols together, you can create credential refresh systems for OAuth1, OAuth2, Basic Auth and even exponential backoff retry policies. The possibilities are endless. Here's a short example of how you could implement a refresh flow for OAuth2 access tokens.
+
+```swift
+class OAuth2Handler: RequestAdapter, RequestRetrier {
+    private typealias RefreshCompletion = (_ succeeded: Bool, _ accessToken: String?, _ refreshToken: String?) -> Void
+
+    private let sessionManager: SessionManager = {
+        let configuration = URLSessionConfiguration.default
+        configuration.httpAdditionalHeaders = SessionManager.defaultHTTPHeaders
+
+        return SessionManager(configuration: configuration)
+    }()
+
+    private let lock = NSLock()
+
+    private var clientID: String
+    private var baseURLString: String
+    private var accessToken: String
+    private var refreshToken: String
+
+    private var isRefreshing = false
+    private var requestsToRetry: [RequestRetryCompletion] = []
+
+    // MARK: - Initialization
+
+    public init(clientID: String, baseURLString: String, accessToken: String, refreshToken: String) {
+        self.clientID = clientID
+        self.baseURLString = baseURLString
+        self.accessToken = accessToken
+        self.refreshToken = refreshToken
+    }
+
+    // MARK: - RequestAdapter
+
+    public func adapt(_ urlRequest: URLRequest) -> URLRequest {
+        if urlRequest.urlString.hasPrefix(baseURLString) {
+            var mutableURLRequest = urlRequest
+            mutableURLRequest.setValue("Bearer " + accessToken, forHTTPHeaderField: "Authorization")
+            return mutableURLRequest
+        }
+
+        return urlRequest
+    }
+
+    // MARK: - RequestRetrier
+
+    public func should(_ manager: SessionManager, retry request: Request, with error: Error, completion: RequestRetryCompletion) {
+        lock.lock() ; defer { lock.unlock() }
+
+        if let response = request.task.response as? HTTPURLResponse, response.statusCode == 401 {
+            requestsToRetry.append(completion)
+
+            if !isRefreshing {
+                refreshTokens { [weak self] succeeded, accessToken, refreshToken in
+                    guard let strongSelf = self else { return }
+
+                    strongSelf.lock.lock() ; defer { strongSelf.lock.unlock() }
+
+                    if let accessToken = accessToken, let refreshToken = refreshToken {
+                        strongSelf.accessToken = accessToken
+                        strongSelf.refreshToken = refreshToken
+                    }
+
+                    strongSelf.requestsToRetry.forEach { $0(succeeded, 0.0) }
+                    strongSelf.requestsToRetry.removeAll()
+                }
+            }
+        } else {
+            completion(false, 0.0)
+        }
+    }
+
+    // MARK: - Private - Refresh Tokens
+
+    private func refreshTokens(completion: RefreshCompletion) {
+        guard !isRefreshing else { return }
+
+        isRefreshing = true
+
+        let urlString = "\(baseURLString)/oauth2/token"
+
+        let parameters: [String: Any] = [
+            "access_token": accessToken,
+            "refresh_token": refreshToken,
+            "client_id": clientID,
+            "grant_type": "refresh_token"
+        ]
+
+        sessionManager.request(urlString, withMethod: .post, parameters: parameters, encoding: .json).responseJSON { [weak self] response in
+            guard let strongSelf = self else { return }
+
+            if let json = response.result.value as? [String: String] {
+                completion(true, json["access_token"], json["refresh_token"])
+            } else {
+                completion(false, nil, nil)
+            }
+
+            strongSelf.isRefreshing = false
+        }
+    }
+}
+
+let baseURLString = "https://some.domain-behind-oauth2.com"
+
+let oauthHandler = OAuth2Handler(
+    clientID: "12345678",
+    baseURLString: baseURLString,
+    accessToken: "abcd1234",
+    refreshToken: "ef56789a"
+)
+
+let sessionManager = SessionManager()
+sessionManager.adapter = oauthHandler
+sessionManager.retrier = oauthHandler
+
+let urlString = "\(baseURLString)/some/endpoint"
+
+manager.request(urlString, withMethod: .get).validate().responseJSON { response in
+    debugPrint(response)
+}
+```
+
+Once the `OAuth2Handler` is applied as both the `adapter` and `retrier` for the `SessionManager`, it will handle an invalid access token error by automatically refreshing the access token and retrying all failed requests in the same order they failed. 
+
+> If you needed them to execute in the same order they were created, you could instead sort them by their task identifiers. 
+
+The example above only checks for a `401` response code which is not nearly robust enough, but does demonstrate how one could check for an invalid access token error. In a production application, one would want to check the `realm` and most likely the `www-authenticate` header response although it depends on the OAuth2 implementation.
+
+Another important note is that this authentication system could be shared between multiple session managers. For example, you may need to use both a `default` and `ephemeral` session configuration for the same set of web services. The example above allows the same `oauthHandler` instance to be shared across multiple session managers to manage the single refresh flow.
+
+> Please note that this is not a global OAuth2 solution. It is merely an example demonstrating how one could use the `RequestAdapter` in conjunction with the `RequestRetrier` to create a thread-safe refresh system.
+
 ### Security
 
 Using a secure HTTPS connection when communicating with servers and web services is an important step in securing sensitive data. By default, Alamofire will evaluate the certificate chain provided by the server using Apple's built in validation provided by the Security framework. While this guarantees the certificate chain is valid, it does not prevent man-in-the-middle (MITM) attacks or other potential vulnerabilities. In order to mitigate MITM attacks, applications dealing with sensitive customer data or financial information should use certificate or public key pinning provided by the `ServerTrustPolicy`.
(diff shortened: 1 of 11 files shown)


cnoon:
cc @jshier and @kcharwood. I'd love to get your guys thoughts here.

README.md
```diff
@@ -1110,6 +1110,175 @@ Generally, either the default implementation or the override closures should pro
...
+
+let sessionManager = SessionManager()
+sessionManager.adapter = AccessTokenAdapter(accessToken: "1234")
+
+sessionManager.request("https://httpbin.org/get", withMethod: .get)
+```
+
+#### RequestRetrier
+
+The `RequestRetrier` protocol allows a `Request` that encountered an `Error` while being executed to be retried. When using both the `RequestAdapter` and `RequestRetrier` protocols together, you can create credential refresh systems for OAuth1, OAuth2, Basic Auth and even exponential backoff retry policies. The possibilities are endless. Here's a short example of how you could implement a refresh flow for OAuth2 access tokens.
```
jshier: Seems like there should be a simpler example or walkthrough of the `RequestRetrier` protocol before jumping directly into an OAuth2 implementation. From looking at it I can't tell what's actually required for the protocol here and what's part of the OAuth2 flow.

Source/SessionDelegate.swift
```diff
@@ -362,19 +365,64 @@ extension SessionDelegate: URLSessionTaskDelegate {
...
-            name: Notification.Name.Task.DidComplete,
-            object: self,
-            userInfo: [Notification.Key.Task: task]
-        )
+        // Run all validations on the request before checking if an error occurred
+        request.validations.forEach { $0() }
 
-        self[task] = nil
+        // Determine whether an error has occurred
+        var error: Error? = error
```
jshier: If an error occurs from validation, wouldn't it immediately then be overwritten by the `taskDelegate` error? Presumably the two conditions are mutually exclusive, but in that case, why run the validators at all?

README.md
tobiasoleary: @jshier Quick Opinion, the name RequestAdapter makes me think on the Adapter pattern in GoF's Design Patterns. I understand why you pick the name, but I think the pattern you are using is more commonly referred to as the Interceptor pattern. So RequestInterceptor might be a better name.

README.md
```diff
@@ -1110,6 +1110,175 @@ Generally, either the default implementation or the override closures should pro
...
+
+> If you needed them to execute in the same order they were created, you could instead sort them by their task identifiers. 
+
+The example above only checks for a `401` response code which is not nearly robust enough, but does demonstrate how one could check for an invalid access token error. In a production application, one would want to check the `realm` and most likely the `www-authenticate` header response although it depends on the OAuth2 implementation.
+
+Another important note is that this authentication system could be shared between multiple session managers. For example, you may need to use both a `default` and `ephemeral` session configuration for the same set of web services. The example above allows the same `oauthHandler` instance to be shared across multiple session managers to manage the single refresh flow.
+
+> Please note that this is not a global OAuth2 solution. It is merely an example demonstrating how one could use the `RequestAdapter` in conjunction with the `RequestRetrier` to create a thread-safe refresh system.
```
jshier: Perhaps we should move any disclaimers above the example, to increase the chance they are seen before people copy/paste their OAuth2 implementation. ðº

README.md
```diff
@@ -1110,6 +1110,175 @@ Generally, either the default implementation or the override closures should pro
 
 > It is important to keep in mind that the `subdelegates` are initialized and destroyed in the default implementation. Be careful when subclassing to not introduce memory leaks.
 
+### Adapting and Retrying Requests
```
jshier: In general, do we want any discussion of when to use an adapter vs. when to use a router?

Source/SessionDelegate.swift
```diff
@@ -362,19 +365,64 @@ extension SessionDelegate: URLSessionTaskDelegate {
...
+
+        if let taskDelegate = self[task]?.delegate, taskDelegate.error != nil {
+            error = taskDelegate.error
+        }
+
+        /// If an error occurred and the retrier is set, asynchronously ask the retrier if the request
+        /// should be retried. Otherwise, complete the task by notifying the task delegate.
+        if let retrier = retrier, let error = error {
```
jshier: Slightly better version with a guard for `shouldRetry`. I'd argue you could use a guard for the whole thing but recognize that there's a stylistic issue there, since the retry behavior is truly optional.
```swift
if let retrier = retrier, let error = error {
            retrier.should(sessionManager, retry: request, with: error) { [weak self] shouldRetry, delay in
                guard shouldRetry else {
                    completeTask(session, task, error)
                    return
                }
                
                DispatchQueue.utility.after(delay) { [weak self] in
                    guard let strongSelf = self else { return }

                    let retrySucceeded = strongSelf.sessionManager?.retry(request) ?? false

                    if retrySucceeded {
                        strongSelf[request.task] = request
                        return
                    } else {
                        completeTask(session, task, error)
                    }
                }
            }
        } else {
            completeTask(session, task, error)
        }
```

Source/SessionManager.swift
```diff
@@ -328,25 +362,18 @@ open class SessionManager {
...
-        case .request(let request):
-            queue.sync {
-                downloadTask = self.session.downloadTask(with: request)
-            }
-        case .resumeData(let resumeData):
-            queue.sync {
-                downloadTask = self.session.downloadTask(withResumeData: resumeData)
-            }
+        case let .request(urlRequest):
+            let urlRequest = adapt(urlRequest)
```
jshier: Might be a bit clearer here if it was `let adaptedRequest = adapt(urlRequest)` so there's so confusion between the incoming parameter and what's actually requested. Plus it matches some other areas where the adapted request is named `adaptedRequest`.

Source/SessionManager.swift
```diff
@@ -606,28 +633,22 @@ open class SessionManager {
         var HTTPBodyStream: InputStream?
```
jshier: Not part of this PR, but this needs to be renamed to `httpBodyStream`.

Source/SessionDelegate.swift
```diff
@@ -362,19 +365,64 @@ extension SessionDelegate: URLSessionTaskDelegate {
...
+            return
         }
 
-        NotificationCenter.default.post(
-            name: Notification.Name.Task.DidComplete,
-            object: self,
-            userInfo: [Notification.Key.Task: task]
-        )
+        // Run all validations on the request before checking if an error occurred
+        request.validations.forEach { $0() }
```
jshier: This means validations are synchronously now, right?

Tests/SessionManagerTests.swift
```diff
@@ -131,7 +177,7 @@ class SessionManagerTestCase: BaseTestCase {
         XCTAssertNil(manager, "manager should be nil")
     }
 
-    // MARK: Start Requests Immediately Tests
+    // MARK: Tests - Start Requests Immediately
```
jshier: These are clearly all tests, so perhaps it's not the ordering that matters but the existence of the word at all? I don't care either way.

jshier:
@cnoon Initial review complete. We discussed some of the general issues I had but it looks great otherwise!

README.md
cnoon: I agree @jshier...a simpler example is definitely warranted. I'll make a note to try to get something put together when we rework all the docs. @tobiasoleary I respectfully disagree with the naming. An `interceptor` doesn't imply the `Request` will be modified in any way.

README.md
cnoon: I couldn't agree more. I'll try to work out some better disclaimers here.

Source/SessionDelegate.swift
cnoon: This just means that they'll run on the SessionDelegate's serial queue with is running on some background thread. I'll modify this so that the completion all dispatches to the utility queue though so we don't hold up the session delegate's queue. 

We want the networking to continue to run as fast as possible.

README.md
cnoon: Definitely, I'll mark it down as a TODO for the rework of the documentation.

Source/SessionDelegate.swift
cnoon: Great question. The validators all early out if an `error` has already occurred. All this logic does is grab the resulting error if it actually occurred to be able to hand over to the retrier.

```swift
    @discardableResult
    public func validate(_ validation: Validation) -> Self {
        let validationExecution: () -> Void = {
            if
                let response = self.response,
                self.delegate.error == nil,
                case let .failure(error) = validation(self.request, response, self.delegate.data)
            {
                self.delegate.error = error
            }
        }

        validations.append(validationExecution)

        return self
    }
```

Source/SessionDelegate.swift
cnoon: That's fair, I'll change it to use your approach. 👍🏼

Source/SessionManager.swift
cnoon: Great catch! I actually deleted the variable altogether in one of the downstream PRs. I'll just leave it as is for now if that's cool.

Tests/SessionManagerTests.swift
cnoon: This is just the convention I've used throughout the rest of the test suite. It helps more clearly define what are tests and what are helper types, properties, private helper methods, etc. I'd recommend we leave this as is for now then we can revisit the whole test suite once we ship 4.0.

Source/SessionManager.swift
cnoon: Yeah...there's definitely room for improvement, which is why I reworked it in a following PR! The `adapt` API is now an extension on `URLRequest` which takes the `adapter` as an argument. That cleaned up things a bunch. I'd say let's move this through as is right now and we can revisit in a downstream PR if you still think it's a bit clunky.

Source/SessionDelegate.swift
cnoon: Addressed in 904d0d36.

README.md
cnoon: Addressed in 2f005e49. We'll get a more in-depth review of all the docs when we rework them. But this should be good for now.

cnoon:
Thanks for all the awesome comments here @jshier. Everything has been addressed at this point so I'm going to merge it in to keep the ball rollin'. 👍🏼

cnoon:
You have a couple of places you could do that @alizainprasla, the `validation` closures or in the `response` closures. The `validation` closures give you a chance to inspect the response to decide whether to throw an error or not. If you do throw an error, then the retrier will be called to give you the option to retry the request if you so choose.

jshier:
@ricardopereira It would be much easier to just allow adapter errors to trigger retry behavior. I'm not sure if that would work now, but it might become possible for Alamofire 5. Such a feature likely wouldn't ship until then anyway.

ricardopereira:
@jshier That sounds even better, would be simpler to reuse the logic behind reauthentication. Should I open a Feature request?

jshier:
Sure. At the very least it will give us a fresh issues to discuss adaptation vs. retry.

PR #1453, opened 2016-09-04
Podspec config to build target with Swift3

mherod:
Adding this Podspec configuration line means that the SWIFT_VERSION does not need to manually set to 3.0 in Xcode and is done on install. This configuration is being followed conventionally by many Pods being updated for Swift 3.

### DIFF
--- a/Alamofire.podspec
+++ b/Alamofire.podspec
@@ -14,4 +14,6 @@ Pod::Spec.new do |s|
   s.watchos.deployment_target = '2.0'
 
   s.source_files = 'Source/*.swift'
+  
+  s.pod_target_xcconfig = { 'SWIFT_VERSION' => '3.0' }
 end


jshier:
Thanks for the PR!

First, this is unnecessary, as CocoaPods as of `1.1.0.beta.1` (which you need to be using for Xcode 8 compatibility anyway) will set the `SWIFT_VERSION` of the generated `Pods` project to whatever the target project specifies. Second, it doesn't fix our current issue with publishing the `4.0.0-beta.2` spec, which you can read more about [here](https://github.com/CocoaPods/CocoaPods/issues/5796). To fix that issue requires a change on the CocoaPods side. Once we have a resolution to that issue we'll make the necessary changes, if any. Until then, using the CocoaPods beta works fine for consumers of the framework.

Cheers!

PR #1455, opened 2016-09-05
Feature - Request Subclasses and New Progress APIs

cnoon:
This PR splits up the `Request` class into `DataRequest`, `DownloadRequest`, `UploadRequest` and `StreamRequest` subclasses.

## Motivation

There were several reasons why the `Request` class could greatly benefit from being split up. 

### Progress

The first reason is that the progress APIs were stomping on each other and were confusing the users. For example, the upload progress behavior was very ambiguous considering that the same `Progress` instance was used to report both upload progress and then download progress after the upload was complete.

### Session Manager Duplication

The `feature/adapter-and-retrier` PR added the `TaskConvertible` enumeration to the `Request` class. Unfortunately, there is a large amount of duplication between the `TaskConvertible` implementation and the `Downloadable`, `Uploadable` and `Streamable` enumerations inside the `SessionManager`. Not only is there duplication, but the `Downloadable`, `Uploadable` and `Streamable` enumerations don't really belong in the `SessionManager` anyways. They're really the pieces required to construct a concrete `Request`.

By splitting up the `Request` into subclasses, we could move the `Downloadable`, `Uploadable` and `Streamable` implementations into their respective subclass and eliminate the duplication with the `TaskConvertible` enumeration in the `Request` class.

In the previous PR where the `TaskConvertible` enumeration was added, the `Downloadable`, `Uploadable` and `Streamable` enumerations

### Response Serializers

Another awesome reason for splitting up the `Request` class is to associate response serializers with a particular request type. For example, right now you can chain a `responseData` closure onto a `Request` that is backed by a `URLSessionDownloadTask` or a `ULRSessionStreamTask`. However, if you do this, what's the behavior? It currently gets really weird in all these cases when you start to consider resume data. 

Also, how does one get the destination URL of a download? Great question, right now it is WAY more difficult than it should be. By splitting up the `Request`, we can extend only certain subclasses to support specific response serializers that are tailored to the underlying type of task. These changes will be coming in a future PR.

## Solution

So now that we have some background as to why these changes would be useful, let's dig into what had to change to make this work.

### TaskConvertible

By pulling the `Downloadable`, `Uploadable` and `Streamable` enumerations into their respective subclasses and adding a `Requestable` enumeration to the `DataRequest` subclass, I was able to get rid of the `TaskConvertible` enumeration and make it a protocol. The protocol now creates a `URLSessionTask` from a session, adapter and queue. This allows the subclassed enumerations to construct the tasks directly without having to do this in the `SessionManager`. This greatly simplifies the `SessionManager.retry` method.

### Progress

The progress APIs have been completely rebuilt from scratch. You can now monitor `downloadProgress` on data, download and upload requests. You can also monitor `uploadProgress` on an upload request. There are variants for both the `Progress` object and the broken out `Int64` values. Each variant can also call the progress closure on a specified queue.

Here's a quick example of how you can use the new APIs.

```swift
Alamofire.upload(data, to: urlString, withMethod: .post)
    .uploadProgress { progress in
        // Called on main dispatch queue by default
        print("upload progress: \(progress.fractionCompleted)")
    }
    .uploadProgress(queue: DispatchQueue.utility) { bytesSent, totalBytesSent, totalBytesExpectedToSend in
        // Can customize the dispatch queue for background operations if needed
        print("Sent: \(bytesSent), Total Sent: \(totalBytesSent), Total Expected: \(totalBytesExpectedToSend)")
    }
    .downloadProgress { progress in
        // Called on main dispatch queue by default
        print("download progress: \(progress.fractionCompleted)")
    }
    .downloadProgress { bytesRead, totalBytesRead, totalBytesExpectedToRead in
        // Called on main dispatch queue by default
        print("Read: \(bytesRead), Total Read: \(totalBytesRead), Total Expected: \(totalBytesExpectedToRead)")
    }
    .response { response in
        debugPrint(response)
    }
```

#### Download Progress

The download progress is now being tracked in the `DataTaskDelegate`. This allows the data, download and upload requests to leverage it. I debated for a long time as to whether to call this `progress` or `downloadProgress`. I would rather error on the side of verbosity here I think. I ultimately decided upon `downloadProgress` because `progress` is REALLY ambiguous when it comes to an upload request.

> If anyone strongly disagrees here, please voice your concerns. I'm still on the fence on this one, but right now lean towards `downloadProgress` and `uploadProgress` APIs to completely avoid ambiguity.

#### Upload Progress

The upload progress is all managed by the `UploadTaskDelegate` and is only exposed publicly through the upload request subclass. The stream request cannot chain `downloadProgress` or `uploadProgress` APIs because it no longer inherits them.

### Top-Level Stream APIs

I added the top-level stream APIs to `Alamofire.swift` because they were missing. Just wanted to callout that this random change made its way into this PR as well.

---

## Summary

These changes vastly improve the `progress` APIs and open up the ability to greatly specialize the available chained methods for each subclass. This will help avoid confusion and will allow for better customization for each request type.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -178,18 +178,18 @@ open class Request {
         return self
     }
 
-    /// Returns a base64 encoded basic authentication credential as an authorization header dictionary.
+    /// Returns a base64 encoded basic authentication credential as an authorization header tuple.
     ///
     /// - parameter user:     The user.
     /// - parameter password: The password.
     ///
-    /// - returns: A dictionary with Authorization key and credential value or empty dictionary if encoding fails.
-    open static func authorizationHeaderFrom(user: String, password: String) -> [String: String] {
-        guard let data = "\(user):\(password)".data(using: String.Encoding.utf8) else { return [:] }
+    /// - returns: A tuple with Authorization header and credential value if encoding succeeds, `nil` otherwise.
+    open static func authorizationHeaderFrom(user: String, password: String) -> (key: String, value: String)? {
+        guard let data = "\(user):\(password)".data(using: .utf8) else { return nil }
 
         let credential = data.base64EncodedString(options: [])
 
-        return ["Authorization": "Basic \(credential)"]
+        return (key: "Authorization", value: "Basic \(credential)")
     }
 
     // MARK: Progress
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -66,6 +66,14 @@ protocol TaskConvertible {
 /// Responsible for sending a request and receiving the response and associated data from the server, as well as
 /// managing its underlying `URLSessionTask`.
 open class Request {
+    /// A closure executed when monitoring upload or download progress of a request.
+    public typealias ProgressHandler = (Progress) -> Void
+
+    /// A closure executed when monitoring the download progress of a request.
+    public typealias DownloadProgressHandler = (_ bytesReceived: Int64, _ totalBytesReceived: Int64, _ totalBytesExpectedToReceive: Int64) -> Void
+
+    /// A closure executed when monitoring the upload progress of a request.
+    public typealias UploadProgressHandler = (_ bytesSent: Int64, _ totalBytesSent: Int64, _ totalBytesExpectedToSend: Int64) -> Void
 
     // MARK: Properties
 
@@ -92,9 +100,6 @@ open class Request {
 
     /// The response received from the server, if any.
     open var response: HTTPURLResponse? { return task.response as? HTTPURLResponse }
-
-    /// The progress of the request lifecycle.
-    open var progress: Progress { return delegate.progress }
 
     let originalTask: TaskConvertible?
 
@@ -169,32 +174,6 @@ open class Request {
         let credential = data.base64EncodedString(options: [])
 
         return (key: "Authorization", value: "Basic \(credential)")
-    }
-
-    // MARK: Progress
-
-    /// Sets a closure to be called periodically during the lifecycle of the request as data is written to or read
-    /// from the server.
-    ///
-    /// - For uploads, the progress closure returns the bytes written, total bytes written, and total bytes expected
-    ///   to write.
-    /// - For downloads and data tasks, the progress closure returns the bytes read, total bytes read, and total bytes
-    ///   expected to read.
-    ///
-    /// - parameter closure: The code to be executed periodically during the lifecycle of the request.
-    ///
-    /// - returns: The request.
-    @discardableResult
-    open func progress(closure: ((Int64, Int64, Int64) -> Void)? = nil) -> Self {
-        if let uploadDelegate = delegate as? UploadTaskDelegate {
-            uploadDelegate.uploadProgress = closure
-        } else if let dataDelegate = delegate as? DataTaskDelegate {
-            dataDelegate.dataProgress = closure
-        } else if let downloadDelegate = delegate as? DownloadTaskDelegate {
-            downloadDelegate.downloadProgress = closure
-        }
-
-        return self
     }
 
     // MARK: State
@@ -349,6 +328,9 @@ extension Request: CustomDebugStringConvertible {
 
 /// Specific type of `Request` that manages an underlying `URLSessionDataTask`.
 open class DataRequest: Request {
+
+    // MARK: Helper Types
+
     enum Requestable: TaskConvertible {
         case request(URLRequest)
 
@@ -365,7 +347,14 @@ open class DataRequest: Request {
         }
     }
 
+    // MARK: Properties
+
+    /// The progress of fetching the response data from the server for the request.
+    open var progress: Progress { return dataDelegate.progress }
+
     var dataDelegate: DataTaskDelegate { return delegate as! DataTaskDelegate }
+
+    // MARK: Stream
 
     /// Sets a closure to be called periodically during the lifecycle of the request as data is read from the server.
     ///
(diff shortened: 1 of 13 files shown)


cnoon:

Source/Request.swift
```diff
@@ -178,62 +162,18 @@ open class Request {
...
+    /// Returns a base64 encoded basic authentication credential as an authorization header tuple.
     ///
     /// - parameter user:     The user.
     /// - parameter password: The password.
     ///
-    /// - returns: A dictionary with Authorization key and credential value or empty dictionary if encoding fails.
-    open static func authorizationHeaderFrom(user: String, password: String) -> [String: String] {
-        guard let data = "\(user):\(password)".data(using: String.Encoding.utf8) else { return [:] }
+    /// - returns: A tuple with Authorization header and credential value if encoding succeeds, `nil` otherwise.
+    open static func authorizationHeaderFrom(user: String, password: String) -> (key: String, value: String)? {
```
jshier: Not really a part of this PR, but I think the naming here could be better. Perhaps `authorizationHeader(withUser:password)`.

Source/Request.swift
```diff
@@ -414,3 +323,283 @@ extension Request: CustomDebugStringConvertible {
...
+    enum Requestable: TaskConvertible {
+        case request(URLRequest)
+
+        func task(session: URLSession, adapter: RequestAdapter?, queue: DispatchQueue) -> URLSessionTask {
+            var task: URLSessionTask!
+
+            switch self {
+            case let .request(urlRequest):
+                let urlRequest = urlRequest.adapt(using: adapter)
+                queue.sync { task = session.dataTask(with: urlRequest) }
```
jshier: Not necessarily for this PR, but to avoid things like this in the future, it might be nice to have a convenience value synchronization wrapper in the future:
```swift
extension DispatchQueue {
    
    func syncAround<T>(_ closure: () -> T) -> T {
        var out: T!
        sync { out = closure() }
        return out
    }
    
}
```

Just an example, but would avoid the IUO dance here and could be returned directly if the `switch` was a `guard`.

Source/Request.swift
```diff
@@ -414,3 +323,283 @@ extension Request: CustomDebugStringConvertible {
...
 }
+
+// MARK: -
+
+/// Specific type of `Request` that manages an underlying `URLSessionDataTask`.
+open class DataRequest: Request {
+
+    // MARK: Helper Types
+
+    enum Requestable: TaskConvertible {
```
jshier: I'm not sure an `enum` with one case is the best here. I think a `struct` would be simpler. 
```swift
struct Requestable: TaskConvertible {
    let request: URLRequest
    
    func task(session: URLSession, adapter: RequestAdapter?, queue: DispatchQueue) -> URLSessionTask {
        var task: URLSessionTask!
        queue.sync { task = session.dataTask(with: request) }
        return task
    }
}
```

Source/Request.swift
```diff
@@ -57,25 +57,23 @@ public protocol RequestRetrier {
 
 // MARK: -
 
+protocol TaskConvertible {
```
jshier: Just double checking to see if it's necessary for this to be public? Perhaps if user's subclass everything they'd need this?

Source/Request.swift
```diff
@@ -414,3 +323,283 @@ extension Request: CustomDebugStringConvertible {
...
+}
+
+// MARK: -
+
+#if !os(watchOS)
+
+/// Specific type of `Request` that manages an underlying `URLSessionStreamTask`.
+open class StreamRequest: Request {
+    enum Streamable: TaskConvertible {
+        case stream(String, Int)
```
jshier: Do we want labels on these cases? I know they're internal but it may be good even for us.

Source/SessionManager.swift
```diff
@@ -179,7 +139,7 @@ open class SessionManager {
     /// `nil` by default.
     open var backgroundCompletionHandler: (() -> Void)?
 
-    let queue = DispatchQueue(label: "Alamofire Session Manager Queue")
+    let queue = DispatchQueue(label: "org.alamofire.session-manager." + NSUUID().uuidString)
```
jshier: `NSUUID` -> `UUID`.

Source/TaskDelegate.swift
```diff
@@ -182,8 +180,12 @@ class DataTaskDelegate: TaskDelegate, URLSessionDataDelegate {
...
     }
 
-    var dataProgress: ((_ bytesReceived: Int64, _ totalBytesReceived: Int64, _ totalBytesExpectedToReceive: Int64) -> Void)?
-    var dataStream: ((_ data: Data) -> Void)?
+    var progress: Progress
+
+    var progressHandler: (closure: Request.ProgressHandler, queue: DispatchQueue)?
+    var progressDebugHandler: (closure: Request.DownloadProgressHandler, queue: DispatchQueue)?
+
+    var dataStream: ((_ data: Data) -> Void)? // this should use handler pattern also
```
jshier: Do we want this comment?

jshier:
@cnoon Review complete. Not much feedback here, looks great!

Source/Request.swift
cnoon: I really want to get rid of the `with` altogether here. Any factor type methods seem like they don't need it. For example, the `expectation(withDescription:)` API in XCTest is now `expectation(description:)`. Thoughts?

Source/Request.swift
cnoon: What would the usage look like here?

Source/Request.swift
cnoon: That seems so logical! Will fix.

Source/SessionManager.swift
cnoon: 🙌🏼 will fix.

Source/Request.swift
cnoon: Great question. Let me investigate a bit and get back to you.

Source/Request.swift
cnoon: Definitely, will add. 👍🏼

Source/Request.swift
jshier: The guidelines don't offer much guidance in this case, except to say:
```
When the first argument forms part of a prepositional phrase, give it an argument label. The argument label should normally begin at the preposition, e.g. x.removeBoxes(havingLength: 12).
```
I believe `with` is the prepostion here. That said, I'm okay with dropping it in most cases.

Source/TaskDelegate.swift
cnoon: Haha, I'm sure not. Great catch. Will probably just remove.

Source/Request.swift
jshier: You'd be able to use it here like:
```swift
task = queue.syncAround { session.dataTask(with: urlRequest) }
```
Naming can be changed, but I saw the IUO -> sync -> return dance a few times in these changes.

cnoon: I dig it. Will change.

Source/Request.swift
cnoon: Addressed in 004b3156.

Source/Request.swift
cnoon: Added `syncResult` in b56bd7d8. I changed it in most places throughout the codebase except in places that I know were changed in other PRs that will generate merge conflicts.

Source/Request.swift
cnoon: Switched to struct in 7e570726.

Source/Request.swift
cnoon: Okay, so after a long investigation, I think we're good to keep this internal. There's nothing about the `TaskConvertible` that needs to be exposed publicly. We don't have any issues subclassing `SessionManager` or `Request` or any of the subclasses. The `Request` initializer is also internal so I think we're good.

In the worst case scenario, we can always bump the ACL to public with only a MINOR version bump.

Source/Request.swift
cnoon: Addressed in 08c0bec9.

Source/SessionManager.swift
cnoon: Fixed in eca9afa9.

Source/TaskDelegate.swift
cnoon: Removed in eca9afa9. I don't think we need a custom dispatch queue for the stream closure unless someone really wants to see that. IMO it's probably best to call that directly from the session delegate queue.

cnoon:
All changes have been addressed @jshier. I'm going to merge this bad boy.

davewang:
I download  file url is  http://10.7.7.100:8080/auth0919.tar.gz progress always 0.0,BTW Other url(not .tar.gz ) is ok

gnik2036:
https://stackoverflow.com/questions/54621441/uploading-image-to-server-using-swift-4

any one can help me

PR #1456, opened 2016-09-05
Refactor User-Agent

### LINKED ISSUE #1452: User agent string causes certain web servers to fail request
The 3.4.2 update of Alamofire creates a user agent string with an extra close paren in Manager.swift line 95. This causes certain web servers to fail when validating the user agent string.

`return "\(executable)/\(bundle) (\(appVersion)/\(appBuild)); \(osNameVersion))"`

should be:

`return "\(executable)/\(bundle) (\(appVersion)/\(appBuild); \(osNameVersion))"`

jshier:
Fixes #1452 

This PR updates our `User-Agent` string generation to not only fix the issue reported but bring it more in line with [RFC7231](https://tools.ietf.org/html/rfc7231#section-5.5.3) in regards to where the various bits of information live. It now looks something like this: `iOS Example/1.0 (com.alamofire.iOS-Example; build:1; iOS 9.3.0) Alamofire/3.4.2`. This more closely matches the `User-Agent` strings returned by browsers and now includes the Alamofire version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,14 @@ Alamofire.request(.GET, "https://httpbin.org/get", headers: headers)
          }
 ```
 
+Alamofire provides a common set of headers for every request. These include:
+
+* `Accept-Encoding`, which defaults to `gzip;q=1.0, compress;q=0.5`, per [RFC7230](https://tools.ietf.org/html/rfc7230#section-4.2.3).
+* `Accept-Language`, which defaults to up to the top 6 preferred languages on the system, formatted like `en;q=1.0`, per [RFC7231](https://tools.ietf.org/html/rfc7231#section-5.3.5).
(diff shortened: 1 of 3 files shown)


cnoon:
Thanks for putting this together @jshier! I made some small modifications (formatting and doc clarification) and pushed commit 68ddac4d into `master` while keeping attribution. I'll also take care of cherry picking this into the `swift2.3` and `swift3` branches right now.

weifans:
Is it really solved?
What I get now is `MyAppName/3 CFNetwork/1121.2.1 Darwin/19.2.0`

jshier:
@weifans That would indicate that you're not using Alamofire's `User-Agent` header but getting the system's default instead. If you're customizing your `URLSessionConfiguration`, be sure to add Alamofire's default headers.

weifans:
Thanks for your reply, I solved him.

casperson:
How do you add Alamofire's default headers?

Issue #1458, opened 2016-09-05
Xcode 8 Beta 6, Swift 3: ParameterEncoding QueryComponents Type Issue


dhanushram:
In `ParameterEncoding.queryComponents()` method, `value` is checked against `Bool` (see below pasted code). But in Swift 3 if `value` is of type `NSNumber`, this check will always succeed. So even though `value` might contain `NSNumber` of value, say 25, queryComponent will always return 1. 

Proposed Fix: ` if let bool = value as? Bool, !(value is NSNumber)`

```Swift
        } else if let bool = value as? Bool {
            components.append((escape(key), escape((bool ? "1" : "0"))))
        } else {
            components.append((escape(key), escape("\(value)")))
        }

cnoon:
Issue has been fixed in 72958d0e.

Issue #1056, opened 2016-02-09
Bool in URL encoded parameters encodes as 0 and 1


pheuberger:
Invoking the request method with a `[String, AnyObject]` dictionary passed as argument yields query parameters where Bools are encoded as 0s and 1s. I would like them to be encoded as `true` and `false` respectively. The parameter encoding in this request was `.URL`.

```swift
Alamofire.request(endpoint.method(), url, parameters: parameters, encoding: encoding, headers: headers)
         .validate(statusCode: 200...299)
         .responseJSON { response in self.handleRequestResponse(response, observer: observer) }
```
I'm using Swift 2 in XCode 7.2.1 with a mixed codebase.

cnoon:
This isn't true. What happens is the `Bool` bridges to an `NSNumber` when converted to an `AnyObject`. When you print out an `NSNumber` created from a `Bool`, it prints out `0` or `1`. However, once you run it through `NSJSONSerialization`, it encodes it as either `false` or `true`.

Cheers. 🍻

pheuberger:
But you wouldn't run URL query parameters through `NSJSONSerialization` would you?
Anyways, I'm monitoring the traffic with TCPMon and it shows the URLs with `Bool`s encoded as 0s and 1s. Right now I'm working around this by converting `Bool` values to `String`s before making the request. Maybe this is intended behavior, if so let me know.

Cheers

cnoon:
Oh my apologies, I missed the `.URL` encoding parameter type. I'm going to re-open this and investigate further.

pheuberger:
No worries, thanks for looking into this!

cnoon:
Okay @pheuberger, I managed to dig into this a bit more. The behavior you are seeing is intended. The reason has to do with the fact that all the parameters bridge over to Objective-C. The parameter encoding logic for URL encoding relies on the `CustomStringConvertible` protocol to covert the `AnyObject` to a string representation. In the case of an `NSNumber`, this will always default to `0` or `1` in the case of a `Bool`.

```swift
let boolValue = true
print(boolValue) // outputs `true`

let bridgedBool: AnyObject = true
print(bridgedBool) // outputs `1`
```

[...]

PR #1460, opened 2016-09-05
Fixing NSNumber vs Bool type issue.

dhanushram:
Proposed fix for https://github.com/Alamofire/Alamofire/issues/1458

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -193,7 +193,7 @@ public enum ParameterEncoding {
             for value in array {
                 components += queryComponents(fromKey: "\(key)[]", value: value)
             }
-        } else if let bool = value as? Bool {
+        } else if let bool = value as? Bool, !(value is NSNumber) {
             components.append((escape(key), escape((bool ? "1" : "0"))))
         } else {
             components.append((escape(key), escape("\(value)")))


cnoon:
Thanks for the PR @dhanushram, but we won't be accepting this change for the same reasons we rejected it in #1056.

Thanks anyways. Cheers. 🍻

cnoon:
My apologies here @dhanushram...I completely misunderstood what you were doing. Too many late nights. I'm going to re-open and investigate further.

jshier:
This seems like what Argo had to deal with in properly detecting `Bool` from `JSONSerialization`. Their solution was to [drop down to CoreFoundation](https://github.com/thoughtbot/Argo/blob/841979f8b8341b7818a15e67662157c4580151f1/Argo/Extensions/NSNumber.swift) and use those APIs to properly detect that an `NSNumber` contained a `Bool`. I think that's the solution we need here. If we can cast to a a `Bool` then check if it's an `NSNumber`. If it isn't we're good, it's a `Bool`. Otherwise check to see if the `NSNumber` contains a `Bool`.

cnoon:
Thanks for putting this together @dhanushram...much appreciated! @jshier is certainly right and it required a more robust check. We had to do something similar in Elevate. I've implemented a `fileprivate` NSNumber extension in the `ParameterEncoding.swift` file and leveraged it in the `queryComponents` API. I've also added failing (now passing) unit tests. I pushed up the fix to `master` in 72958d0e.

Thanks again! 🍻

PR #1462, opened 2016-09-06
Feature - Destinations Closures, Download Options and Top-Level APIs

cnoon:
This PR refactors the download destination system as well as the top-level APIs for constructing requests.

## Motivation

Up until now, the `DownloadFileDestination` closures have been too limiting. They "force" you to move the file from the temporary URL to some other URL. Even if you don't want to move it, you still have to implement a destination closure and return the temp URL.

A bigger issue is that you have to do all your preparation of the file system up front. You cannot modify the file system at the time the destination closure is called. For example, let's say I have an avatar image at `$Sandbox/Documents/avatar.jpg` and I want to update it. Right now, you have to either delete the `avatar.jpg` file before making the request or move it to a different location. Otherwise you'll get a file system error when trying to move the file because a file already exists at the destination url. But you don't want to delete it before you make the request because you don't know that the request will actually succeed.

On a different note, we've wanted to remove the `with` portion of the `withMethod` external parameter name in the top-level APIs. We've also wanted to figure out how add default values for the method without introducing ambiguity.

### Download File Destination Closure

The first major change in this PR is to make the `destination` closure optional in all the top-level APIs. You no longer have to move the file if you don't want to. It's now perfectly acceptable to do the following:

```swift
Alamofire.download("https://httpbin.org/get").responseJSON { response in
    debugPrint(response)
}
```

If you do this, the file will stay in the temporary location. If you want to move it somewhere else, you'll need to handle that on your own.

> It's important to note that you should always move the file out of the temporary location if you are going to need access to it at a later time.

Now if you do want to move the file to a different destination, you'll need to use the `DownloadFileDestination` closure which is a bit different.

```swift
public typealias DownloadFileDestination = (
    _ temporaryURL: URL,
    _ response: HTTPURLResponse)
    -> (destinationURL: URL, options: DownloadOptions)
```

You'll notice that the typealias is the same as it was before with the exception of the `DownloadOptions` in the return type.

### Download Options

The `DownloadOptions` struct is an `OptionSet` that comes with a couple cool new features.

* `.createIntermediateDirectories` - creates all intermediate directories to the destination url if specified
* `.removePreviousFile` - removes a previous file at the destination url if it exists right before moving the temp file to the destination url

These options allow you to customize the file system behavior right before the temp file is moved to the final destination url. Here's an example of what it looks like to use the new destination closure system.

```swift
let destination: DownloadRequest.DownloadFileDestination = { _, _ in 
    return (fileURL, [.removePreviousFile, .createIntermediateDirectories]) 
}

Alamofire.download(urlString, to: destination).response { response in
    debugPrint(response)
}
```

### DownloadResponse

The `DownloadResponse` has been modified to support both the `temporaryURL` and `destinationURL` properties. The `temporaryURL` will always be set if the file was successfully downloaded. The `destinationURL` will be set if the file was successfully downloaded and the `destination` closure was set. Make sure to check whether the `Result` was a success though to know the file was able to be copied to the `destinationURL`. If the move operation fails, an error is generated, but `destinationURL` is not `nil`. I designed it this way so that when an error occurs, you have as much information as possible to debug what happened.

### Download Request Validation

The `Validation` typealias for a `DownloadRequest` has been modified to support both the `temporaryURL` and `destinationURL`.

```swift
public typealias Validation = (URLRequest?, HTTPURLResponse, _ temporary: URL?, _ destination: URL?) -> ValidationResult
```

### Top-Level APIs

The goal of the top-level API changes was to set `method` defaults and get rid of the `with` portion of the `withMethod` external parameter name. The results are pretty awesome. You can now do the following:

```swift
Alamofire.request("https://httpbin.org/get") // defaults to `.get`
Alamofire.download("https://httpbin.org/get") // defaults to `.get`
Alamofire.upload("https://httpbin.org/post") // defaults to `.post`
```

Since this is certainly the majority of the use cases, this makes the call sights much more concise. If you do need to specify the `method`, it's now just `method` instead of `withMethod`.

```swift
Alamofire.request("https://httpbin.org/delete", method: .delete)
```

In order to disambiguate the `request` APIs, we had to add the external `resource` parameter name to the `URLRequestConvertible` variant.

```swift
Alamofire.request(resource: urlRequest)
Alamofire.download(resource: urlRequest, to: destination)
```

The reason for this is that `URLRequest` conforms to both `URLStringConvertible` and `URLRequestConvertible`. Rather than going down the route of removing conforms, I decided to just remove the potential of ambiguity altogether. Otherwise users could add their own extensions to types that would re-introduce ambiguity. And as well all know, ambiguous errors suck to debug.

One thing that I think is still open is whether we add the `resource` name to the upload APIs.

```swift
// Current
Alamofire.upload(fileURL, to: urlString)
Alamofire.upload(fileURL, with: urlRequest)

// Possible Improvements
Alamofire.upload(fileURL, to: urlRequest)
Alamofire.upload(fileURL, toResource: urlRequest)
```

Thoughts here?

---

## Summary

These changes greatly improve the functionality and usefulness of download requests by allowing them to support all the potential use cases. In the event that the `DownloadOptions` don't offer you enough flexibility, you can alway fall back to moving the file on your own in the response serializer. The top-level API changes are a huge improvement and make the common usage cases very concise.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -106,7 +106,7 @@ extension URLRequest {
 /// specified `urlString`, `method`, `parameters`, `encoding` and `headers`.
 ///
 /// - parameter urlString:  The URL string.
-/// - parameter method:     The HTTP method.
+/// - parameter method:     The HTTP method. `.get` by default.
 /// - parameter parameters: The parameters. `nil` by default.
 /// - parameter encoding:   The parameter encoding. `.url` by default.
 /// - parameter headers:    The HTTP headers. `nil` by default.
@@ -115,7 +115,7 @@ extension URLRequest {
 @discardableResult
 public func request(
     _ urlString: URLStringConvertible,
-    withMethod method: HTTPMethod,
+    method: HTTPMethod = .get,
     parameters: [String: Any]? = nil,
     encoding: ParameterEncoding = .url,
     headers: [String: String]? = nil)
@@ -123,7 +123,7 @@ public func request(
 {
     return SessionManager.default.request(
         urlString,
-        withMethod: method,
+        method: method,
         parameters: parameters,
         encoding: encoding,
         headers: headers
@@ -137,8 +137,8 @@ public func request(
 ///
 /// - returns: The created `DataRequest`.
 @discardableResult
-public func request(_ urlRequest: URLRequestConvertible) -> DataRequest {
-    return SessionManager.default.request(urlRequest.urlRequest)
+public func request(resource urlRequest: URLRequestConvertible) -> DataRequest {
+    return SessionManager.default.request(resource: urlRequest)
 }
 
 // MARK: - Download Request
@@ -229,8 +229,8 @@ public func download(
 /// and `headers` for uploading the `file`.
 ///
 /// - parameter file:      The file to upload.
-/// - parameter method:    The HTTP method.
 /// - parameter urlString: The URL string.
+/// - parameter method:    The HTTP method. `.post` by default.
 /// - parameter headers:   The HTTP headers. `nil` by default.
 ///
 /// - returns: The created `UploadRequest`.
@@ -238,11 +238,11 @@ public func download(
 public func upload(
     _ fileURL: URL,
     to urlString: URLStringConvertible,
-    withMethod method: HTTPMethod,
+    method: HTTPMethod = .post,
     headers: [String: String]? = nil)
     -> UploadRequest
 {
-    return SessionManager.default.upload(fileURL, to: urlString, withMethod: method, headers: headers)
+    return SessionManager.default.upload(fileURL, to: urlString, method: method, headers: headers)
 }
 
 /// Creates a `UploadRequest` using the default `SessionManager` from the specified `urlRequest` for
@@ -264,7 +264,7 @@ public func upload(_ fileURL: URL, with urlRequest: URLRequestConvertible) -> Up
 ///
 /// - parameter data:      The data to upload.
 /// - parameter urlString: The URL string.
-/// - parameter method:    The HTTP method.
+/// - parameter method:    The HTTP method. `.post` by default.
 /// - parameter headers:   The HTTP headers. `nil` by default.
 ///
 /// - returns: The created `UploadRequest`.
@@ -272,11 +272,11 @@ public func upload(_ fileURL: URL, with urlRequest: URLRequestConvertible) -> Up
 public func upload(
     _ data: Data,
     to urlString: URLStringConvertible,
-    withMethod method: HTTPMethod,
+    method: HTTPMethod = .post,
     headers: [String: String]? = nil)
     -> UploadRequest
 {
-    return SessionManager.default.upload(data, to: urlString, withMethod: method, headers: headers)
+    return SessionManager.default.upload(data, to: urlString, method: method, headers: headers)
 }
 
 /// Creates an `UploadRequest` using the default `SessionManager` from the specified `urlRequest` for
@@ -298,7 +298,7 @@ public func upload(_ data: Data, with urlRequest: URLRequestConvertible) -> Uplo
 ///
 /// - parameter stream:    The stream to upload.
 /// - parameter urlString: The URL string.
-/// - parameter method:    The HTTP method.
+/// - parameter method:    The HTTP method. `.post` by default.
 /// - parameter headers:   The HTTP headers. `nil` by default.
 ///
 /// - returns: The created `UploadRequest`.
@@ -306,11 +306,11 @@ public func upload(_ data: Data, with urlRequest: URLRequestConvertible) -> Uplo
 public func upload(
     _ stream: InputStream,
     to urlString: URLStringConvertible,
-    withMethod method: HTTPMethod,
+    method: HTTPMethod = .post,
     headers: [String: String]? = nil)
     -> UploadRequest
 {
-    return SessionManager.default.upload(stream, to: urlString, withMethod: method, headers: headers)
+    return SessionManager.default.upload(stream, to: urlString, method: method, headers: headers)
 }
 
 /// Creates an `UploadRequest` using the default `SessionManager` from the specified `urlRequest` for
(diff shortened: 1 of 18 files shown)


cnoon:

Source/Alamofire.swift
```diff
@@ -148,65 +148,74 @@ public func request(_ urlRequest: URLRequestConvertible) -> DataRequest {
 /// Creates a `DownloadRequest` using the default `SessionManager` to retrieve the contents of a URL based on the
 /// specified `urlString`, `method`, `parameters`, `encoding`, `headers` and save them to the `destination`.
 ///
+/// If `destination` is not specified, the contents will remain in the temporary location determined by the
+/// URL session.
```
jshier: Should this be more specific and say `by the underlying URLSession`?

Source/Request.swift
```diff
@@ -405,10 +405,42 @@ open class DownloadRequest: Request {
...
-    /// response, and returns a single argument: the file URL where the temporary file should be moved.
-    public typealias DownloadFileDestination = (URL, HTTPURLResponse) -> URL
+    /// A collection of options to be executed prior to moving a downloaded file from the temporary URL to the
+    /// destination URL.
+    public struct DownloadOptions: OptionSet {
+        /// Defines the `RawValue` type as `UInt` to satisfy the `RawRepresentable` protocol.
+        public typealias RawValue = UInt
+
+        /// Returns the raw bitmask value of the option and satisfies the `RawRepresentable` protocol.
+        public let rawValue: RawValue
```
jshier: You can just define `rawValue` with a type and it will be inferred. This can be simplified to:
```swift
public struct DownloadOptions: OptionSet {
    /// Returns the raw bitmask value of the option and satisfies the `RawRepresentable` protocol.
    public let rawValue: UInt
    
    /// A `DownloadOptions` flag that creates intermediate directories for the destination URL if specified.
    public static let createIntermediateDirectories = DownloadOptions(rawValue: 1 << 0)
    
    /// A `DownloadOptions` flag that removes a previous file from the destination URL if specified.
    public static let removePreviousFile = DownloadOptions(rawValue: 1 << 1)
    
    // MARK: Initialization Methods
    
    /// Creates a `DownloadFileDestinationOptions` instance with the specified raw value.
    ///
    /// - parameter rawValue: The raw bitmask value for the option.
    ///
    /// - returns: A new log level instance.
    public init(rawValue: UInt) { self.rawValue = rawValue }
}
```
Weirdly the init can't be generated automatically for some reason. But the intermediate static values seem unnecessary, unless there's something I'm missing?

Source/Response.swift
```diff
@@ -92,7 +92,7 @@ extension DataResponse: CustomStringConvertible, CustomDebugStringConvertible {
     }
 
     /// The debug textual representation used when written to an output stream, which includes the URL request, the URL
-    /// response, the server data and the response serialization result.
+    /// response, the server data, the response serialization result and the timeline.
```
jshier: Oxford comma or not?

Source/Validation.swift
```diff
@@ -218,9 +218,9 @@ extension DataRequest {
 // MARK: -
 
 extension DownloadRequest {
-    /// A closure used to validate a request that takes a URL request, a URL response and destination URL, and returns
-    /// whether the request was valid.
-    public typealias Validation = (URLRequest?, HTTPURLResponse, URL?) -> ValidationResult
+    /// A closure used to validate a request that takes a URL request, a URL response, a temporary URL and a 
+    /// destination URL, and returns whether the request was valid.
+    public typealias Validation = (URLRequest?, HTTPURLResponse, _ temporary: URL?, _ destination: URL?) -> ValidationResult
```
jshier: You've already broken the column limit, why not go all the way and label all of the parameters? Labeling only two seems odd.

Tests/DownloadTests.swift
```diff
@@ -226,13 +204,13 @@ class DownloadResponseTestCase: BaseTestCase {
...
         let headers = ["Authorization": "123456"]
-        let destination: DownloadRequest.DownloadFileDestination = { _, _ in fileURL }
+        let destination: DownloadRequest.DownloadFileDestination = { _, _ in (fileURL, []) }
 
         let expectation = self.expectation(description: "Download request should download data to file: \(fileURL)")
         var response: DefaultDownloadResponse?
 
         // When
-        Alamofire.download(urlString, to: destination, withMethod: .get, headers: headers)
+        Alamofire.download(urlString, method: .get, headers: headers, to: destination)
```
jshier: Shouldn't need `method` here.

Tests/UploadTests.swift
```diff
@@ -282,7 +282,7 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
                 formData = multipartFormData
             },
             to: urlString,
-            withMethod: .post,
+            method: .post,
```
jshier: Now unnecessary?

Source/Alamofire.swift
```diff
@@ -148,65 +148,74 @@ public func request(_ urlRequest: URLRequestConvertible) -> DataRequest {
...
 /// - parameter urlRequest:  The URL request.
-/// - parameter destination: The closure used to determine the destination of the downloaded file.
+/// - parameter destination: The closure used to determine the destination of the downloaded file. `nil` by default.
 ///
 /// - returns: The created `DownloadRequest`.
 @discardableResult
 public func download(
-    _ urlRequest: URLRequestConvertible,
-    to destination: DownloadRequest.DownloadFileDestination)
+    resource urlRequest: URLRequestConvertible,
```
jshier: What about just using `request`? It may look a little odd `Alamofire.request(request: request)` but it's more accurate about what the parameter actually is.

jshier:
@cnoon Reviews finished, just some minor comments. Looks great!

Source/Alamofire.swift
cnoon: Sure! Will change.

Source/Request.swift
cnoon: The private static properties are merely to group the raw values together to make it virtually impossible to screw up adding a new option. I can remove it if you think it's overboard. I'll definitely remove the typealias declaration!

Source/Validation.swift
cnoon: Yep, it's odd. Will add names across the board.

Tests/DownloadTests.swift
cnoon: Will remove. At first I kept some around just to switch things up a bit to make sure everything was working correctly.

Tests/UploadTests.swift
cnoon: Will remove. 👍🏼

Source/Alamofire.swift
cnoon: Ummm...I would prefer we stick with `resource`. We've used it for the `resumeData` APIs and I think it works well there.

> Discussed on Slack and we'll keep as is.

Source/Request.swift
jshier: If you think they're important, keep them, I just don't think they are.

Source/Alamofire.swift
cnoon: Addressed in 66fc1a87.

Source/Request.swift
cnoon: Simplified in 7a245519.

Source/Validation.swift
cnoon: Addressed in 2e08bb73.

Tests/UploadTests.swift
cnoon: Removed in bc2f0aa9.

Tests/DownloadTests.swift
cnoon: Removed in bc2f0aa9.

cnoon:
Thanks for the review @jshier...much appreciated! 🍻

All comments have been addressed on this PR. I'm going to go ahead and move it through.

PR #1464, opened 2016-09-06
Feature - Parameter Encoding Safety

cnoon:
This PR is a work in progress (WIP). I switched the `ParameterEncoding` `encode` API over to properly throwing an AFError if encoding fails. I'll write up more details tomorrow, but long story short this change throws a parameter encoding error that gets pushed all the way down into the `TaskDelegate` which will bubble out the end after the request completes.

@jshier I'm looking for a rough nod from you that this approach seems viable before I finish it up. Right now I think all that's missing are some updated docstrings and the AFError extensions in the test suite.


cnoon:
Declining this PR since we're going to go with the approach in #1465 instead.

PR #1465, opened 2016-09-07
Feature - Parameter Encoding Protocol

cnoon:
This PR is an alternative to #1464 that goes WAY further in refactoring the `ParameterEncoding` logic into a protocol instead of an enumeration.

## Motivation

The `ParameterEncoding` enum hasn't been flexible enough to allow us to customize the behavior of a particular encoding type. For example, we had to introduce the new `.urlEncodedInURL` case just to allow users to be able to force the parameters into the query string when their backends required them to do so on a `POST` or `PUT`.

The enum also doesn't allow us to open up different writing options for the JSON and plist serializers. We can't customize plist formats, etc. The list goes on and on since we're locked to the enum that can't add extra cases or associated values in a backwards compatible manner.

## Solution

Get rid of the enumeration altogether! Now the `ParameterEncoding` type of a protocol backed by concrete `URL`, `JSON` and `PropertyList` conforming structs.

```swift
public typealias Parameters = [String: Any]

public protocol ParameterEncoding {
    func encode(_ urlRequest: URLRequestConvertible, with parameters: Parameters?) throws -> URLRequest
}
```

It was amazing after I split everything up how much sense it really makes. `ParameterEncoding` implementations are now free of the enumeration bounds and are free to customize to their own needs. 

### URLEncoding

The `.url` and `.urlEncodedInURL` cases have now been combined in the `URLEncoding` struct with some new enumerations and properties of it's own.

```swift
public struct URLEncoding: ParameterEncoding {
    public enum Destination {
        case methodDependent, queryString, httpBody
    }

    public static var `default`: URLEncoding { return URLEncoding() }
    public static var methodDependent: URLEncoding { return URLEncoding() }
    public static var queryString: URLEncoding { return URLEncoding(destination: .queryString) }
    public static var httpBody: URLEncoding { return URLEncoding(destination: .httpBody) }

    public let destination: Destination

    public init(destination: Destination = .methodDependent) {
        self.destination = destination
    }

    public func encode(_ urlRequest: URLRequestConvertible, with parameters: Parameters?) throws -> URLRequest {
        // left out for simplicity
    }
}
```

I added convenience properties for accessing instances without having to go through the initializer. This was a nice touch. We could actually hide the initializer if we wanted, but I figured that was a step too far.

### JSONEncoding

The `JSONEncoding` struct is much less complicated than the `URLEncoding` struct for obvious reasons.

> It's simpler!

```swift
public struct JSONEncoding: ParameterEncoding {
    public static var `default`: JSONEncoding { return JSONEncoding() }
    public static var prettyPrinted: JSONEncoding { return JSONEncoding(options: .prettyPrinted) }

    public let options: JSONSerialization.WritingOptions

    public init(options: JSONSerialization.WritingOptions = []) {
        self.options = options
    }

    public func encode(_ urlRequest: URLRequestConvertible, with parameters: Parameters?) throws -> URLRequest {
        // left out for simplicity
    }
}
```

### PropertyListEncoding

The `PropertyListEncoding` struct is very similar to the `JSONEncoding` struct, just with a few more properties.

```swift
public struct PropertyListEncoding: ParameterEncoding {
    public static var `default`: PropertyListEncoding { return PropertyListEncoding() }
    public static var xml: PropertyListEncoding { return PropertyListEncoding(format: .xml) }
    public static var binary: PropertyListEncoding { return PropertyListEncoding(format: .binary) }

    public let format: PropertyListSerialization.PropertyListFormat
    public let options: PropertyListSerialization.WriteOptions

    public init(
        format: PropertyListSerialization.PropertyListFormat = .xml,
        options: PropertyListSerialization.WriteOptions = 0)
    {
        self.format = format
        self.options = options
    }

    public func encode(_ urlRequest: URLRequestConvertible, with parameters: Parameters?) throws -> URLRequest {
        // left out for simplicity
    }
}
```

> If you noticed `default` and ` xml` are exact duplicate properties, bonus points for you. I wanted to follow the same `.default` property convention across all the `ParameterEncoding` types for consistency. That way at the call sight you can always use `<encoding_struct>.default` and know it will be there. _I wonder if that should be enforced in the protocol_ 🤔

### Wrapping Up Parameters with Encoding

So I tried for probably 2 hours to work the parameters and the encoding into a single type. I had `Parameters` as the protocol and `URLParameters`, `JSONParameters`, etc. and you'd create the object like `URLParameters(parameters)` and call it like so:

```swift
Alamofire.request(urlString, parameters: URLParameters(parameters)
```

Part of this was great because it got rid of the `encoding` parameter altogether which I was all for. However, it greatly increased the majority of our calls because it can't rely on the default parameter value of `.url` anymore. So we end up having to write `URLParameters(parameters)` everywhere instead of just `parameters`. Not to mention it just feels a little off.

Because of all of this, I finally decided enough was enough and it was just a square peg, round hole scenario. Once I backed away, I was able to finally land on the solution you see here.

### Throwing vs. Double Tuple

This also includes all the changes to the `encode` method itself which was part of the previous PR. These changes should definitely be adopted in AF 4. It's time to eliminate the last odd double optional case we have left! 🙌🏼🏆

---

## Summary

Overall, I think this is an awesome set of changes for AF 4. It makes it so much easier to create your own custom `ParameterEncoding` type. It also rids the world of the `.urlEncodedInURL` case that I sadly had to introduce a year ago.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -41,132 +41,105 @@ public enum HTTPMethod: String {
 
 // MARK: -
 
-/// Used to specify the way in which a set of parameters are applied to a URL request.
+/// A dictionary of parameters to apply to a `URLRequest`.
+public typealias Parameters = [String: Any]
+
+/// A type used to define how a set of parameters are applied to a `URLRequest`.
+public protocol ParameterEncoding {
+    /// Creates a URL request by encoding parameters and applying them onto an existing request.
+    ///
+    /// - parameter urlRequest: The request to have parameters applied.
+    /// - parameter parameters: The parameters to apply.
+    ///
+    /// - throws: An `AFError.parameterEncodingFailed` error if encoding fails.
+    ///
+    /// - returns: The encoded request.
+    func encode(_ urlRequest: URLRequestConvertible, with parameters: Parameters?) throws -> URLRequest
+}
+
+// MARK: -
+
+/// Creates a url-encoded query string to be set as or appended to any existing URL query string or set as the HTTP
+/// body of the URL request. Whether the query string is set or appended to any existing URL query string or set as
+/// the HTTP body depends on the destination of the encoding.
 ///
-/// - url:             Creates a query string to be set as or appended to any existing URL query for `GET`, `HEAD`,
-///                    and `DELETE` requests, or set as the body for requests with any other HTTP method. The
-///                    `Content-Type` HTTP header field of an encoded request with HTTP body is set to
-///                    `application/x-www-form-urlencoded; charset=utf-8`. Since there is no published specification
-///                    for how to encode collection types, the convention of appending `[]` to the key for array
-///                    values (`foo[]=1&foo[]=2`), and appending the key surrounded by square brackets for nested
-///                    dictionary values (`foo[bar]=baz`).
-///
-/// - urlEncodedInURL: Creates query string to be set as or appended to any existing URL query. Uses the same
-///                    implementation as the `.url` case, but always applies the encoded result to the URL.
-///
-/// - json:            Uses `JSONSerialization` to create a JSON representation of the parameters object, which is
-///                    set as the body of the request. The `Content-Type` HTTP header field of an encoded request is
-///                    set to `application/json`.
-///
-/// - propertyList:    Uses `PropertyListSerialization` to create a plist representation of the parameters object,
-///                    according to the associated format and write options values, which is set as the body of the
-///                    request. The `Content-Type` HTTP header field of an encoded request is set to
-///                    `application/x-plist`.
-///
-/// - custom:          Uses the associated closure value to construct a new request given an existing request and
-///                    parameters.
-public enum ParameterEncoding {
-    case url
-    case urlEncodedInURL
-    case json
-    case propertyList(PropertyListSerialization.PropertyListFormat, PropertyListSerialization.WriteOptions)
-    case custom((URLRequestConvertible, [String: Any]?) throws -> URLRequest)
+/// The `Content-Type` HTTP header field of an encoded request with HTTP body is set to
+/// `application/x-www-form-urlencoded; charset=utf-8`. Since there is no published specification for how to encode 
+/// collection types, the convention of appending `[]` to the key for array values (`foo[]=1&foo[]=2`), and appending
+/// the key surrounded by square brackets for nested dictionary values (`foo[bar]=baz`).
+public struct URLEncoding: ParameterEncoding {
+
+    // MARK: Helper Types
+
+    /// Defines whether the url-encoded query string is applied to the existing query string or HTTP body of the 
+    /// resulting URL request.
+    ///
+    /// - methodDependent: Applies encoded query string result to existing query string for `GET`, `HEAD` and `DELETE` 
+    ///                    requests and sets as the HTTP body for requests with any other HTTP method.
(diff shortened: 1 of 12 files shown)


Source/ParameterEncoding.swift
```diff
@@ -41,139 +41,108 @@ public enum HTTPMethod: String {
...
-                if urlRequest.value(forHTTPHeaderField: "Content-Type") == nil {
-                    urlRequest.setValue("application/json", forHTTPHeaderField: "Content-Type")
-                }
+        guard let parameters = parameters else { return urlRequest }
 
-                urlRequest.httpBody = data
-            } catch {
-                encodingError = error
+        if let method = HTTPMethod(rawValue: urlRequest.httpMethod!), encodesParametersInURL(with: method) {
+            if var URLComponents = URLComponents(url: urlRequest.url!, resolvingAgainstBaseURL: false), !parameters.isEmpty {
```
jshier: I didn't think `if var` was still in Swift 3. Are we sure all of the IUOs here are safe? Or should it be part of the error handling?

Source/ParameterEncoding.swift
```diff
@@ -41,139 +41,108 @@ public enum HTTPMethod: String {
...
-                encodingError = error
+        } else {
+            if urlRequest.value(forHTTPHeaderField: "Content-Type") == nil {
+                urlRequest.setValue("application/x-www-form-urlencoded; charset=utf-8", forHTTPHeaderField: "Content-Type")
             }
-        case .custom(let closure):
-            (urlRequest, encodingError) = closure(urlRequest, parameters)
+
+            urlRequest.httpBody = query(parameters).data(
+                using: String.Encoding.utf8,
```
jshier: You can probably use `.utf8` here.

jshier:
@cnoon Looks great, just a few comments. Also, how would this interact with the `URLRequestConvertible` protocol and the router pattern? It would be great to use these encoders or a custom one in a router to be able to generate an error if necessary and pass it through the AF request cycle.

Source/ParameterEncoding.swift
cnoon: Definitely, will change.

Source/ParameterEncoding.swift
cnoon: Not sure offhand. I didn't change any of this logic directly. Will take another deep look. 👍🏼

cnoon:
I agree @jshier, but unfortunately it's not possible to pass it down through the request cycle. I posted a big description of the limitations in our Slack channel. We should probably have that discussion over a call. 

Long story short is that we can't because the only way to push the error down is to actually make (and execute) some request. That causes all sorts of issues.

jshier:
I remember, just didn't realize the same thing applied here as well. I'll give it more thought and we can make the improvement later.

Source/ParameterEncoding.swift
cnoon: Addressed in 8ce90881.

Source/ParameterEncoding.swift
cnoon: Removed in b468f74c.

Source/ParameterEncoding.swift
cnoon: Oh, and `if var thing` is still allowed in `if` statements. I think it's only been removed from function signatures.

cnoon:
All comments have been addressed here @jshier...thanks for the good catches! I'm going to go ahead and merge this.

PR #1488, opened 2016-09-08
better update for request

tremblay:
using `_` causes any existing code that called `request(urlRequest)` to
compile but potentially fail. because it needs to be updated to
`request(resource: request)` but this is not caught by the compiler
because `URLRequest` apparently conforms to `URLStringConvertible`.

This issue is evident in some of the tests, which compiled and actually
passed, but were in fact calling the wrong `request`. 

I added a test that would fail if the wrong `request` is called to
illustrate this point.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -122,7 +122,7 @@ public func request(
     -> DataRequest
 {
     return SessionManager.default.request(
-        urlString,
+        url: urlString,
         method: method,
         parameters: parameters,
         encoding: encoding,
--- a/Source/SessionManager.swift
+++ b/Source/SessionManager.swift
@@ -224,7 +224,7 @@ open class SessionManager {
     /// - returns: The created `DataRequest`.
     @discardableResult
     open func request(
-        _ urlString: URLStringConvertible,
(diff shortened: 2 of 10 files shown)


cnoon:
This is a great catch @tremblay! I don't really like modifying the public API that way though. The core issue is that `URLRequest` conforms to `URLStringConvertible` and I'm really leaning towards removing that functionality for the exact reason you posted. I'm going to try this out and see how it feels. Not sure of all the implications of removing the conformance there.

Stay tuned.

cnoon:
Man, great catch @tremblay! I just put together #1491 that takes a different approach to solving the same problem. I'd love to get your thoughts there. We spent forever (literally hours) trying to get the public APIs just right and to be consistent across different request types. I think #1491 accomplishes this nicely by simply removing conformance where it isn't necessary.

cnoon:
Thanks again for finding the issue @tremblay! 

We've decided to go with the approach in #1491 which has been merged into the `master` branch. I'm going to go ahead and close this PR out. If you have any feedback on the approach, please direct all future comments to #1491. We're getting really close to release, so please get your feedback in asap if you have any.

Cheers. 🍻

PR #1489, opened 2016-09-08
Make completion of RequestRetrier escaping

rastersize:
Allows the `RequestRetrier` to perform async work before executing the `completion` closure.

I’m in need of this since I’m using this a point to handle 401 errors and kick off re-authentication (refreshing the access token). Having those requests re-tried automatically once the access token has arrived. As such the `completion` closure needs to be “stored away” for a bit.


cnoon:
Absolutely! Great call @rastersize. This was always the intent, but we missed this change when the GM dropped.

Cheers. 🍻

PR #1491, opened 2016-09-09
Removed URLStringConvertible conformance on URLRequest

cnoon:
This PR is an alternate approach to solving the issue identified in #1488. Overall, I think this is a better approach to solving the problem. This allows us to keep the same public APIs, but allow the compiler to catch all the places where the problem wasn't previously identified.

Overall, I don't think it is necessary to keep the `URLRequest` API conforming to `URLStringConvertible`. We don't actually use it for anything anymore and it's misleading and problematic. I actually almost removed it before but didn't have a strong reason to because I hadn't realized what @tremblay did.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -52,11 +52,6 @@ extension URLComponents: URLStringConvertible {
     public var urlString: String { return url!.urlString }
 }
 
-extension URLRequest: URLStringConvertible {
-    /// The URL string.
-    public var urlString: String { return url!.urlString }
-}
-
 // MARK: -
 
 /// Types adopting the `URLRequestConvertible` protocol can be used to construct URL requests.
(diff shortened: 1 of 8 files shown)


cnoon:

jshier:
ð Looks good to me.

cnoon:
Awesome, thanks @jshier.

adamwaite:
This has broken AlamofireImage 

screen shot 2016-09-09 at 12 49 52

See https://github.com/Alamofire/AlamofireImage/issues/174

tremblay:
For what it's worth, I definitely agree that this is a better solution than my approach in #1488

cnoon:
Thanks for the heads up @adamwaite...we'll get AlamofireImage fixed up as well.

esetnik:
@cnoon @tremblay I just noticed the same issue, strangely enough the Carthage dependency compiles and the app crashes at runtime with a `dyld: lazy symbol binding failed: Symbol not found` error. I would have expected that Carthage would have failed to build AlamofireImage. Any idea why?

cnoon:
AFI is a bit out-of-date with AF at the moment. We're wrapping the AF 4 release right now. Once that is out, AFI will be next.

PR #1492, opened 2016-09-09
Feature - Task Metrics

cnoon:
This PR adds support for the awesome new `URLSessionTaskMetric` statistical tracking. This is essentially what we always wanted the `Timeline` to be, but we didn't have access to the same internal APIs to compute all these data points. Thanks Apple!

> For the time being, it makes sense to keep the `Timeline` around for watchOS and earlier versions of the platforms that don't support the new APIs.

This shouldn't be a difficult feature to build, but it most definitely was due to the complexity of compiling against 4 platforms with only three of the latest actually supporting new APIs. Therefore, I had to get pretty creative with the new `Response` protocol and extension in combination with `AnyObject` abstraction while moving the metrics through the various objects.

Overall, I'm VERY happy with how this turned out. I wish it wasn't so complicated to support so many platforms, but unfortunately, it is. I abstracted the complexity away into the single `Response` protocol and implementation, so that certainly helps isolate all the platform complexity to a single area.

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -37,6 +37,15 @@ public struct DefaultDataResponse {
 
     /// The error encountered while executing or validating the request.
     public let error: Error?
+
+    var _metrics: AnyObject?
+
+    init(request: URLRequest?, response: HTTPURLResponse?, data: Data?, error: Error?) {
+        self.request = request
+        self.response = response
+        self.data = data
+        self.error = error
+    }
 }
 
 // MARK: -
@@ -57,6 +66,8 @@ public struct DataResponse<Value> {
 
     /// The timeline of the complete lifecycle of the `Request`.
     public let timeline: Timeline
+
+    var _metrics: AnyObject?
 
     /// Creates a `DataResponse` instance with the specified parameters derived from response serialization.
     ///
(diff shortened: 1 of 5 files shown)


Source/Response.swift
```diff
@@ -154,6 +183,8 @@ public struct DownloadResponse<Value> {
     /// The timeline of the complete lifecycle of the request.
     public let timeline: Timeline
 
+    var _metrics: AnyObject?
```
jshier: Why is this `AnyObject` and not `URLSessionTaskMetrics`?

jshier: Just so the internal variable didn't need be behind an availability check?

cnoon: Yep...you can't store a `URLSessionTaskMetrics` as a stored property b/c it's not available on the older OS's. This is the only valid workaround I've found to the problem yet. It's nasty, but it does work.

jshier: You can't put the whole stored property behind an availability check?

cnoon: Nope.

jshier:
Looks good.

cnoon:
Awesome, thanks @jshier. I'm going to go ahead and move this through.

PR #1501, opened 2016-09-10
Fix filename comment in TaskDelegate header

auswahlaxiom:
The file was likely renamed without the comment being updated. Comment used to report:
```
//  Error.swift
```


cnoon:
Good catch! Thanks for the fix. 🍻

PR #1505, opened 2016-09-11
Feature - URLRequest Creation Safety

cnoon:
This PR simply put makes Alamofire safe. Invalid URL strings, parameter encoding errors, adaptation errors, etc. are now all allowed to be used and thrown and Alamofire will gracefully handle it and expose the error in the response handlers.

## Motivation

A VERY common problem in Alamofire is that users forget to percent escape their URL strings and Alamofire will crash. Up until now, we (the Alamofire team) have taken the stance that this is how Alamofire is designed and your URLs need to conform to [RFC 2396](https://tools.ietf.org/html/rfc2396). This is certainly not ideal for the community because we all would rather have Alamofire tell us that our URL was invalid rather than having it crash.

## Solution

In order to make this safe, there were several things that needed to change including:

- Making the `task` optional on a `Request`
- Allowing the `URLStringConvertible` protocol to throw
- Allowing the `URLRequestConvertible` protocol to throw
- Allowing the `RequestAdapter` protocol to throw
- Modifying all the `SessionManager` request creation methods to handle all error propagation

The core change here was making the `task` optional. By not requiring a `URLSessionTask` to actually be created and added to the underlying `URLSession`, we freed ourselves from the burden of trying to figure out what to do instead...which was always CRASH, or possibly even worse, send the `URLRequest` off even though we know it was wrong.

Once we made the task optional, I started allowing the protocols involved to be able to throw errors as well. We'd already made the change a few days ago to allow `ParameterEncoding` to throw which was a great change. But previously if it threw, we would record the error on the `Request`, but we would still have to send off the original `URLRequest` (missing the parameter encoding) because Alamofire required us to. You would receive the error in the response handler (good), but the request you intended to send to the server is not actually what got sent (very very bad).

Now, all these problem have been alleviated. If any of the steps encounter an error, Alamofire stops the task creation process, builds the `Request` subclass without the `task`, then starts up the delegate queue when `resume` is called. The response handlers will have access to the original error, and no task was ever sent out since we knew we couldn't create the one you wanted.

### Remaining Tasks

There are still some documentation updates that need to be added to this PR before we move it through. They include:

- [x] Update README with all API changes
- [x] Update `URLStringConvertible` section of migration guide
- [x] Add `URLRequestConvertible` section to migration guide
- [x] Update `RequestAdapter` section of migration guide

---

## Summary

Of all the great changes going into Alamofire 4, I think this one is probably the best. Alamofire 3.x is simply not safe if you don't make sure every request going through it is set up properly. Now most of the community has adjusted to this stance, but it's all to simple to miss a single case and have it end up crashing your app in production.

No more with Alamofire 4. Invalid requests simply go through Alamofire and come out on the other side with an error describing exactly what went wrong. 🎉🎉🎉


cnoon:

cnoon:
All open tasks have been completed.

Source/Alamofire.swift
```diff
@@ -27,42 +27,83 @@ import Foundation
...
-    /// A URL that conforms to RFC 2396.
+    /// A URL string that conforms to RFC 2396.
     ///
     /// Methods accepting a `URLStringConvertible` type parameter parse it according to RFCs 1738 and 1808.
     ///
     /// See https://tools.ietf.org/html/rfc2396
     /// See https://tools.ietf.org/html/rfc1738
     /// See https://tools.ietf.org/html/rfc1808
-    var urlString: String { get }
+    var urlString: String? { get }
```
jshier: I don't think it's necessary to include the property in the protocol definition anymore. Just the protocol extension is enough to provide the property.

Source/Alamofire.swift
```diff
@@ -27,42 +27,83 @@ import Foundation
...
+
+extension URLStringConvertible {
+    public var urlString: String? { return try? asURLString() }
 }
 
 extension String: URLStringConvertible {
-    /// The URL string.
-    public var urlString: String { return self }
+    /// Returns a URL string that conforms to RFC 2396.
+    public func asURLString() throws -> String { return self }
```
jshier: Seems like we want to do something here to ensure the string can be converted to a URL.

Source/Alamofire.swift
```diff
@@ -27,42 +27,83 @@ import Foundation
...
+    }
 }
 
 // MARK: -
 
 /// Types adopting the `URLRequestConvertible` protocol can be used to construct URL requests.
 public protocol URLRequestConvertible {
     /// The URL request.
-    var urlRequest: URLRequest { get }
+    var urlRequest: URLRequest? { get }
```
jshier: Same as with `URLStringConvertible`, I don't think we need the property in the protocol anymore.

Source/Alamofire.swift
cnoon: Removed in e48a78e6.

Source/Alamofire.swift
cnoon: Switched to `URLConvertible` protocol in e48a78e6.

Source/Alamofire.swift
cnoon: Removed in e48a78e6.

README.md
```diff
@@ -1158,7 +1151,7 @@ enum Router: URLRequestConvertible {
 ```
 
 ```swift
-Alamofire.request(resource: Router.ReadUser("mattt")) // GET /users/mattt
+Alamofire.request(Router.ReadUser("mattt")) // GET /users/mattt
```
jshier: Should be `Alamofire.request(Router.readUser("mattt")`.

Source/Alamofire.swift
jshier: I think we decided this property can be removed.

Source/SessionManager.swift
```diff
@@ -250,33 +247,45 @@ open class SessionManager {
...
-        let task = originalTask.task(session: session, adapter: adapter, queue: queue)
-        let request = DataRequest(session: session, task: task, originalTask: originalTask)
+            let task = try originalTask.task(session: session, adapter: adapter, queue: queue)
+            let request = DataRequest(session: session, requestTask: .data(originalTask, task))
 
-        delegate[request.delegate.task] = request
+            delegate[task] = request
 
-        if startRequestsImmediately { request.resume() }
+            if startRequestsImmediately { request.resume() }
```
jshier: Perhaps this could be put in a `defer` block before the `do` so it's not called both here and in `request(failedWith:)`?

Source/SessionManager.swift
```diff
@@ -612,28 +660,38 @@ open class SessionManager {
...
-        }
+            if case let .stream(inputStream, _) = uploadable {
+                upload.delegate.taskNeedNewBodyStream = { _, _ in inputStream }
+            }
 
-        delegate[request.delegate.task] = request
+            delegate[task] = upload
 
-        if startRequestsImmediately { request.resume() }
+            if startRequestsImmediately { upload.resume() }
```
jshier: Similar to my suggestion for `request`, you can probably stick this in a `defer`.

Source/SessionManager.swift
```diff
@@ -672,14 +730,24 @@ open class SessionManager {
...
-        let request = StreamRequest(session: session, task: task, originalTask: streamable)
+        do {
+            let task = try streamable.task(session: session, adapter: adapter, queue: queue)
+            let request = StreamRequest(session: session, requestTask: .stream(streamable, task))
 
-        delegate[request.delegate.task] = request
+            delegate[task] = request
 
-        if startRequestsImmediately { request.resume() }
+            if startRequestsImmediately { request.resume() }
```
jshier: Similar to `request` and `upload`, you can likely stick this in a `defer`.

jshier:
@cnoon Review completion, looks awesome! Just some minor comments and then it's good to land.

README.md
cnoon: Addressed in 9296d43f.

Source/SessionManager.swift
cnoon: I gave it a whirl and it ends up being a bit messier than the current version. It's not bad in the `DataRequest`, but it duplicates a bunch more code in the `download` and `upload` cases. I would recommend we leave as is to keep the implementation exactly the same across all four subclasses.

Source/SessionManager.swift
cnoon: Same as previous comment.

Source/SessionManager.swift
cnoon: Same as previous comment.

cnoon:
All comments addressed @jshier...ready to merge this guy?

Source/SessionManager.swift
```diff
@@ -672,14 +730,24 @@ open class SessionManager {
     // MARK: Private - Stream Implementation
 
     private func stream(_ streamable: StreamRequest.Streamable) -> StreamRequest {
-        let task = streamable.task(session: session, adapter: adapter, queue: queue)
-        let request = StreamRequest(session: session, task: task, originalTask: streamable)
+        do {
+            let task = try streamable.task(session: session, adapter: adapter, queue: queue)
+            let request = StreamRequest(session: session, requestTask: .stream(streamable, task))
```
muescha: why not `request` renamed to `stream` like in the `upload` func?

PR #1523, opened 2016-09-14
Specific Swift Version to 2.3 to support XCode 8

Wasith:


Wasith:
This not fix issue

cnoon:
@Wasith, you need to use Alamofire 3.5.0 if you need to use Swift 2.3. The `master` branch and Alamofire 4.0.0 now support Swift 3.0 only.

Issue #1425, opened 2016-08-25
Request calling twice after 401 server error


filippovdev:
Hello dear Alamofire developers, I have a problem now that after server returns "401 Authorization Required" your library calling  same request second time, receive the same error and only after that stopping and make callback with error.

My code is super standard:

```
let aReq = Alamofire.request(mutableURLRequest) //request with some parameters
aReq.responseJSON { response in
    //do something with response
}
```

After some investigation I find that in file Request.swift in line 304:

```
if challenge.previousFailureCount > 0 {
                    disposition = .RejectProtectionSpace
                } else {
                    credential = self.credential ?? session.configuration.URLCredentialStorage?.defaultCredentialForProtectionSpace(challenge.protectionSpace)

                    if credential != nil {
                        disposition = .UseCredential
                    }
                }
```
challenge.previousFailureCount is 0 after first call and first error, why it behaves like this ??? Only after second call it equal 1 and Alamofire make callback with error. Can I setup library somehow to stop request and return with error after first call ?

iOS 9.3.1, Alamofire 3.4.1

cnoon:
Hi @filippovdev,

The only way the request will be made a second time is if you're either providing a credential through something like the `authenticate` method, or there's a credential being picked up from the `URLCredentialStorage`. If there's no credential, then the request will immediately fail and won't be retried.

If you absolutely want to disable this functionality, you can, but I'd advise you not to. How you would turn this off would be:

```swift
let delegate = Alamofire.SessionManager.shared.delegate

delegate.taskDidReceiveChallengeWithCompletion = { session, task, challenge, completion in
    completion(.performDefaultHandling, nil)
}
```

Again, I wouldn't recommend you do this as this will disable other functionality in Alamofire such as the ability to do certificate or public key pinning.

filippovdev:
@cnoon Hi, sorry for late answer. I didn't use .authenticate method, only in parameters I add login and password, when I made call to server it returns Error 401 Authorization Required and after that Alamofire (or core iOS api) again doing the same request and again receive 401, but where is the sense if for the first time login and password wrong, why it call it again.

I call only:
```
let aReq = Alamofire.request(mutableURLRequest) //request with some parameters
aReq.responseJSON { response in
    //do something with response
}
```
and in parameters I have

filippovdev:
@cnoon after deeper investigation, everything working for me if in Request.swift in line 293 I add:
```
if credential != nil {
       disposition = .UseCredential
  } else {
[...]

PR #1601, opened 2016-09-23
Fixed calling second request after server return 401

filippovdev:
Regarding this issue - https://github.com/Alamofire/Alamofire/issues/1425
After server return 401, request calling second time and only after second call it stops. It behaves the same with previous Alamofire version.

After investigation I find that challenge.previousFailureCount still 0 after first 401 error. 
I try code suggected by @cnoon but it doesn't change anything, the code is:
```
let delegate = Alamofire.SessionManager.shared.delegate

delegate.taskDidReceiveChallengeWithCompletion = { session, task, challenge, completion in
    completion(.performDefaultHandling, nil)
}
```
Only this little fix suggested in this PR helps me. After first 401 error if credential is nil I cancel authentication challenge. If this is doesn't brake anything for you please apply this PR.

I call my request like this:
```
let aReq = Alamofire.request(myRequest)
aReq.responseJSON { response in
    //do something with response
}
```

Login and password inside myRequest parameters. If they wrong server return 401. I didn't store they in URLCredentialStorage and I didn't use .authenticate Alamofire closure(because my server require sending login and password in parameters). 

screen shot 2016-09-23 at 10 22 22

### DIFF
--- a/Source/TaskDelegate.swift
+++ b/Source/TaskDelegate.swift
@@ -122,6 +122,8 @@ open class TaskDelegate: NSObject {
 
                 if credential != nil {
                     disposition = .useCredential
+                } else {
+                    disposition = .cancelAuthenticationChallenge
                 }
             }
         }


cnoon:
Apologies for the response in #1425 @filippovdev. After further investigation, I found that the challenge is actually called a second time. This is the default behavior of iOS. It will always try to continue without providing a response to the challenge. This is clearly documented in the [Authentication Challenges](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/URLLoadingSystem/Articles/AuthenticationChallenges.html) documentation from Apple.

> I'd highly recommend you checkout that article. It walks you through all the reasoning for why challenges work the way they do.

Alamofire follows the exact same logic as Apple when deciding how to handle challenges. If you insist on modifying the behavior (we would recommend you don't), then you will need to override the `taskDidReceiveChallenge` closure with your own custom logic, or subclass `SessionDelegate` and provide your own implementation for the task challenge API.

Again, I would strongly recommend against this since Apple has built a fairly sophisticated fallback system to attempt to fulfill challenges across multiple protection spaces. However, if you feel that the second call MUST be eliminated, you can certainly do so. This just isn't something we're interested in having directly in Alamofire.

Cheers. 🍻

PR #1615, opened 2016-09-25
Fix response validation sample code

vishrutshah:
Alamofire 4.0.1 with Swift 3.0 
one more minor bug in the [Response Validation](https://github.com/Alamofire/Alamofire#manual-validation) section.

**Error**: Value of type 'DefaultDataResponse' has no member 'result'.

Looks like code section should be using responseData / responseJSON / other response handlers which provides access to the `result` member on `response` object like below:

```
Alamofire.request("https://httpbin.org/get")
    .validate(contentType: ["application/json"])
    .responseData { response in
        switch response.result {
        case .success:
            print("Validation Successful")
        case .failure(let error):
            print(error)
        }
    }
```

screen shot 2016-09-25 at 12 20 17 am

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ By default, Alamofire treats any completed request to be successful, regardless 
 Alamofire.request("https://httpbin.org/get")
     .validate(statusCode: 200..<300)
     .validate(contentType: ["application/json"])
-    .response { response in
+    .responseData { response in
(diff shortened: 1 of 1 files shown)


cnoon:
Thanks @vishrutshah! 🍻

PR #1646, opened 2016-09-30
Updated podspec targets to use SWIFT_VERSION 3

moh0sw:
This fixes the problem where Xcode pops a warning that it needs to convert the project source into Swift 3, since it isn't sure that the code's been converted already.


cnoon:
This is no longer needed with the `.swift-version` file and CocoaPods 1.1.0-rc.2. Thanks anyways! 🍻

PR #1670, opened 2016-10-05
Update Alamofire 4.0 Migration Guide.md

Lutzifer:
I guess this needs to be DownloadRequest


cnoon:
Yep...thanks for the fix @Lutzifer! 🍻

Issue #1672, opened 2016-10-06
Request Retrier - limit number of retries


Footjy:
Hi,

I've been working with the request retrier to reauthenticate when the request fails due to an expired token.

Is there a way to limit the number of retries ? This would be needed in case of a server bug that would send the expired token error when that's not the problem. There we have an infinite loop.

There may be other cases where this is needed.

I've seen this question unanswered on SO so I guess/hope I'm not the only one needing this. If you have some alternative way of doing this, please let me know.

Thanks,

cnoon:
Hi @Footjy,

There are all sorts of ways you could handle this. I'd start simple and write some check in your `RequestRetrier` that doesn't allow you to run more than say 3 refreshes within a 10 second timeframe. If you do, then stop retrying the request.

Once you get this check in place, you can slowly modify it over time to be more and more robust. If you post a link to the SO question I'd be happy to answer it there as well.

I also just pushed up PR #1681 which adds a `retryCount` property to a `Request` which would allow you to stop retrying a `Request` after it had been tried once or twice. These options should be more than enough to get you going.

If you have other ideas, please don't hesitate to post them. This system is brand new so we don't know all the ways everyone will be using it.

Cheers. 🍻

Footjy:
Hi @cnoon,

Thanks for your answer! The recent pull request was exactly what I was hoping, this is the simplest way to achieve that. And I could indeed have thought about a timer to do that.

Here is the [http://stackoverflow.com/questions/39713280/alamofire-session-manager-limit-retry-count](SO question). I'll let you post the answer :)

Cheers.

Eash:
Hi cnoon, 
    I am working with request retrier for reauthentication of expired token. When token get expired from retrier delegate "should" method, I am getting new token. 

But after getting new token it get exit in delegate, request is not continuing and I am not getting any response. So please let me know, where I am wrong. 

Thanks

PR #1681, opened 2016-10-08
Added Request retryCount property to support RequestRetrier

cnoon:
This PR adds a `retryCount` property to a `Request`, updates the `SessionManager` to increment the count and also adds the appropriate tests.

By adding the `retryCount` to the `Request`, the `RequestRetrier` protocol can use the `retryCount` to determine whether to continue to retry the `Request`.

> More details can be found in #1672.


cnoon:

jshier:
👍 Looks good to me! I am having visions of a future where we need to add another layer of abstraction to deal with all of this state, but we're not quite there yet.

Issue #1674, opened 2016-10-06
Catch error in RequestRetrier.should that was thrown in RequestAdapter.adapt


hanneshallen:
Hi,

We're building an auth handler using RequestAdapter and RequestRetrier to handle token refresh. Following your OAuth2 example in the documentation will work, but it will also trigger a bunch of 401:s every time our app starts (where triggering multiple requests on wakeup). 

Our token refresh response contains an 'expires_in' parameter, which means the app will know if the access token has expired before a new request is made. It would be nice if we could check if the access token has expired in the RequestAdapter.adapt, and if so throw an error that we could catch in RequestRetrier.should.

We would like to avoid the 401:s, and also, this mechanism wouldn't require you to run a full request to know if a token is expired.

Thanks

cnoon:
Hi @hanneshallen,

These are all AWESOME points and you can do "exactly" what you proposed without any changes to Alamofire. It's already all setup to handle this, you just need to implement it. In your `RequestAdapter`, just check if the `expires_in` parameter has actually expired. If it has, throw an `Error`. It can be some custom error you create just for this case such as `CustomError.expiredAccessToken` or something like that.

Then in your `RequestRetrier`, you can check the `error` to see if it's an `.expiredAccessToken` and trigger a refresh. This will make sure the original request is NOT sent and goes immediately to your `RequestRetrier` where you can refresh the access token, then retry the request.

Cheers. 🍻

tholo:
So it seems that if an error is thrown in the RequestAdapter (`adopt`) method, the RequestRetrier (`should`) method is *not* called...

cnoon:
Interesting...thanks for pointing that out @tholo. Re-opening to investigate further.

cnoon:
Yep, you're certainly correct @tholo...it's going to take some creativity to make that work. Let me see what I can come up with.

cnoon:
Okay @tholo and @hanneshallen, please check out #1682. I'd love to get your thoughts.

tholo:
After a couple of false starts in my code (now that it actually does short-circuit) I have verified that this works for me with the changes in #1682

cnoon:
I'm going to close this issue out. Please redirect all further comments to #1682.

Cheers. 🍻

PR #1682, opened 2016-10-09
Feature - Throwing Adapt Error Calls Retrier

cnoon:
This PR was created to allow errors thrown in the adapter to call the retrier before starting up the request's internal dispatch queue. 
## Problem

The reason this is necessary is if you want to be able to throw an `ExpiredAccessToken` type of `Error` in your `RequestAdapter`. This would allow you to do things like preemptively refresh you token before sending any requests to the server even though the public APIs had already started up the request. The adapter would be able to intervene.

> See #1674 for the a more in-depth explanation of the issue.
## Solution

Okay everyone (especially @jshier), I need a sanity check here. I don't think this is too hackish, but I want to make sure everyone else agrees. This was really tricky to implement since there are so many different things going on.

The biggest decision was whether or not to wrap the errors in the `AdaptError` or not. It all boils down to whether we should call the retrier for things like invalid URLs and parameter encoding failures. IMO, we should not. I feel like calling the retrier in those scenarios could cause all sorts of problems where people are accidentally getting themselves into endless retry scenarios without realizing it.

Instead, I created an `AdaptError` to wrap any `Error` that's thrown by the `RequestAdapter`. This is only used internally by Alamofire to figure out whether the encountered error was actually thrown by the `adapter` or not. If it is an `AdaptError`, extract the underlying `Error` and call the `retrier`. If it isn't an `AdaptError`, then the request isn't going to succeed and just start up the request queue.

We could instead actually expose an `AFError.adaptError` that people are required to throw to get this behavior, but that seems a bit odd and easy to misuse. Instead, I think people are going to want to throw all types of errors from the adapter and we should just wrap them up internally to figure out whether or not we should call the retrier. Once we make that decision, we extract the underlying error and continue on.

I haven't actually built any tests around this yet because I wanted to get everyone's thoughts first. I've dug into the download and upload requests a bit and this is definitely possible to support in a similar manner.

Thoughts?

### DIFF
--- a/Source/SessionManager.swift
+++ b/Source/SessionManager.swift
@@ -231,12 +231,14 @@ open class SessionManager {
         headers: HTTPHeaders? = nil)
         -> DataRequest
     {
+        var originalRequest: URLRequest?
+
         do {
-            let urlRequest = try URLRequest(url: url, method: method, headers: headers)
-            let encodedURLRequest = try encoding.encode(urlRequest, with: parameters)
+            originalRequest = try URLRequest(url: url, method: method, headers: headers)
+            let encodedURLRequest = try encoding.encode(originalRequest!, with: parameters)
             return request(encodedURLRequest)
         } catch {
-            return request(failedWith: error)
+            return request(originalRequest, failedWith: error)
         }
     }
 
@@ -248,9 +250,11 @@ open class SessionManager {
     ///
     /// - returns: The created `DataRequest`.
     open func request(_ urlRequest: URLRequestConvertible) -> DataRequest {
+        var originalRequest: URLRequest?
+
         do {
-            let originalRequest = try urlRequest.asURLRequest()
-            let originalTask = DataRequest.Requestable(urlRequest: originalRequest)
+            originalRequest = try urlRequest.asURLRequest()
+            let originalTask = DataRequest.Requestable(urlRequest: originalRequest!)
 
             let task = try originalTask.task(session: session, adapter: adapter, queue: queue)
             let request = DataRequest(session: session, requestTask: .data(originalTask, task))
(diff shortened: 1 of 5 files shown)


cnoon:
cc @jshier and @kcharwood for a sanity check 😉

Source/SessionManager.swift
```diff
@@ -261,15 +265,31 @@ open class SessionManager {
...
+    private func request(_ urlRequest: URLRequest?, failedWith error: Error) -> DataRequest {
+        var requestTask: Request.RequestTask = .data(nil, nil)
+
+        if let urlRequest = urlRequest {
+            let originalTask = DataRequest.Requestable(urlRequest: urlRequest)
+            requestTask = .data(originalTask, nil)
+        }
+
+        let isAdaptError = error is AdaptError
+        let error = error.extractedAdaptError
```
jshier: It seems simpler to just treat `extractedAdaptError` as an optional, rather than having to check type and extract the value separately. Was there a reason for this pattern over just making the `extractedAdaptError` optional? Also, the `extractedAdaptError`, as it is now, seems misleading, because it's either the `extractedAdaptError` or the original error.

jshier:
Looks pretty good, though I'm still worried about the growing complexity. I did have some questions about the error and the unwrapping, and how it was being used. I agree with the `AdaptError` solution, it's just the handling I'm concerned with.

hanneshallen:
This looks good to me. Agree with the `AdaptError` solution to avoid accidental retry loops, and it solves the promblem I posted in https://github.com/Alamofire/Alamofire/issues/1674.

tholo:
This seems to properly address issue in #1674

jshier:
👍 Looks good.

Issue #1530, opened 2016-09-14
GET Request with JSONEncoding doesn't work


gcjg:
I could have this wrong, but the following doesn't work in master:

manager.request("....", method: .get, parameters: parameters, encoding: JSONEncoding.default)

As far as I can tell Alamofire eats the request, doesn't send anything out on the wire, and just times-out after a while.

Whereas this works perfectly:

manager.request("....", method: .post, parameters: parameters, encoding: JSONEncoding.default)

The only difference being the get versus post method. Furthermore, if I do the get but use URLEncoding it works fine.

Is there a bug?

lawgimenez:
https://github.com/Alamofire/Alamofire/issues/1528 could this issue be somehow related, I'm using headers. Also, only POST are working right now except GET.

jshier:
@gcjg Please provide a sample project that replicates what you're seeing. We have many tests, examples, and demo projects that demonstrate that the code you posted does work. If you do, feel free to open another issue with a more in depth analysis and said sample project and we'll be happy to investigate further.

One thing to note, though, is that `JSONEncoding` encodes your parameters as JSON and put it in the `httpBody` of the `URLRequest` created. When used with a `.get` method, the server may not expect this, as `GET` requests usually encode parameters in the URL itself.

gcjg:
Whoa. Did you just close my bug report because the legitimate call I was trying to make was not as common as some others? I'm aware I'm trying to JSON encode the body for a HTTP GET but that is what the web service that I'm using requires. It's still a perfectly normal operation though.

I know your library has many tests and examples, I had to use them to work out how to do certain things because your documentation was lacking in some areas. I would note that you don't have a test for this particular operation.

It still could be user error on my part, but I'm beginning to feel more confident that it's a bug in your library. Maybe you could reopen this issue for me, and I'll see if I can add a bit more, although you only need to do the one line I gave to demonstrate the problem.

jshier:
If you can provide a test project that can replicate the behavior you're seeing, feel free to add it to this issue and I'll reopen or open a new issue. Until then, I've verified the scenario you've described works fine, both in our tests and in a demo app I set up fresh. My only point in bringing up the details of `JSONEncoding` was to ensure you knew about this behavior, as some servers don't parse body details from GET requests (like `httpbin.org`).

PR #1691, opened 2016-10-12
Empty bodies for GET, HEAD and DELETE methods.

taflanidi:
`JSONEncoding` and `PropertyListEncoding` now consider HTTP method. 
In case of GET, HEAD or DELETE parameters are not serialised into request body.

Addressing https://github.com/Alamofire/Alamofire/issues/1530

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -125,16 +125,9 @@ public struct URLEncoding: ParameterEncoding {
 
         guard let parameters = parameters else { return urlRequest }
 
-        if let method = HTTPMethod(rawValue: urlRequest.httpMethod ?? "GET"), encodesParametersInURL(with: method) {
-            guard let url = urlRequest.url else {
-                throw AFError.parameterEncodingFailed(reason: .missingURL)
-            }
-
-            if var urlComponents = URLComponents(url: url, resolvingAgainstBaseURL: false), !parameters.isEmpty {
-                let percentEncodedQuery = (urlComponents.percentEncodedQuery.map { $0 + "&" } ?? "") + query(parameters)
-                urlComponents.percentEncodedQuery = percentEncodedQuery
-                urlRequest.url = urlComponents.url
-            }
+        if let method = HTTPMethod(rawValue: urlRequest.httpMethod ?? "GET"),
+            encodesParametersInURL(with: method), (destination == Destination.methodDependent || destination == Destination.queryString) {
+            urlRequest.url = try encode(parameters: parameters, inRequestUrl: urlRequest.url)
         } else {
             if urlRequest.value(forHTTPHeaderField: "Content-Type") == nil {
                 urlRequest.setValue("application/x-www-form-urlencoded; charset=utf-8", forHTTPHeaderField: "Content-Type")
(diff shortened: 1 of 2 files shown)


skyline75489:
The `GET` part I get it. But the `HEAD` and `DELETE` part seems to be violating HTTP semantics. AFAIK, there is not defined semantics for body in `HEAD` and `DELETE` request.

taflanidi:
@skyline75489, `HEAD` method is actually a `GET`, which does not expect a body in response.
It serves to fetch HEADers without obtaining much data. For instance, you may send a `HEAD` request to check `Last-Modified` date and send subsequent equal `GET` request in case if update is needed.

This means `HEAD` requests cannot contain body.

Body in `DELETE` requests violate HTTP semantics.
According to [section 4.3](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3):

> if the request method does not include defined semantics for an entity-body, then the message-body SHOULD be ignored when handling the request.

And `DELETE` method specs [(section 9.7)](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7) do not define semantics for the body. 

Moreover, ADO.NET Data Services Framework clearly [says](https://msdn.microsoft.com/en-us/library/dd541340.aspx):

> If a Delete request type (section 2.2.7.4) includes a request payload, it MUST be ignored and the request MUST be treated as if no entity body was provided.

— and I'm tending to trust Microsoft in this case. They're really competent when it comes to define integration architecture.

taflanidi:
Hey @cnoon 
Thanks for your detailed answer.

> In most cases, using JSON parameter encoding for a GET is going to be incorrect, but not always. There could be some crazy set of services that decided to deviate from the norm (we've all been there sadly). We want Alamofire to clearly do what it is you're instructing it to do. If a user is saying "I want a GET request with some JSON encoded in the HTTP body", well then that's what we want to them.

I hope you guys know your onions. Still, I would be glad to hear your answer on my comment below.

You see, whether your webservice is crazy or not, your iOS and tvOS won't be able to interact with it in case the server waits for the `GET` request to have a body.

On the SDK level, the `URLSession` framework won't give you any chance to send a `GET` request with a body — your app will simply freeze until the request is timed out; needless to say, the app won't receive any response.

You may send a `GET` request with a non-empty body when your app operates on macOS, that's true. But for the mobile development your framework will simply cause a non-debugable behaviour, detected only by sniffing the traffic with third party instruments like Wireshark or Charles. 

The question is, why `AFNetworking`, the predecessor of the `Alamofire`, [do care](https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFURLRequestSerialization.m#L1240) about its users, but your library cannot afford it? Should we fall back to the more reliable transport?

jshier:
I'm not sure what you're talking about. `GET` requests with body data work just fine, it's just that the backend server may not expect requests to pass parameters that way. We've never observed or had reports of freezing because of this issue, just servers not receiving parameters. Also, contrary to the AFNetworking serialization architecture, Alamofire's serializers are intended to be simple, single purpose, and for users to switch serialization type based on request, rather than having a single serializer that serve all requests. And like @cnoon said, baking this fallback behavior into the serializers themselves limits the ultimate flexibility of the framework.

taflanidi:
@jshier, gotcha.

So, you [assume](https://github.com/Alamofire/Alamofire/blob/master/Source/ParameterEncoding.swift#L128) the method type (`GET`, `POST` etc.) to define where to put the request parameters (`body` or `URL`) for your URL encoder, but refuse to do the same thing for other encoders?

Such that, when I use an URL encoder with all its default settings, I [can be sure](https://github.com/Alamofire/Alamofire/blob/master/Source/ParameterEncoding.swift#L226) that my parameters are serialised correctly **automatically**, but the JSON encoder doesn't deserve the same thing. To behave well "out of the box" without additional tuning.

Okay, my pull request didn't contain the `destination` enumeration property for the JSON encoder, so we could have a consistent flexible behaviour between all your encoders, but I didn't hear such an argument from you. All I see is that you are trying to achieve some dubious simplicity.

You say that
> Alamofire's serializers are intended to be simple ... for users to switch serialization type based on request

but man, that's a lie. You don't have any "request-based" builders, like [the one](https://square.github.io/okhttp/2.x/okhttp/com/squareup/okhttp/Request.Builder.html) that `OkHttp` library offers. You've got a wildcard `request` method, which doesn't tell me a word if I call it like this:

```swift
request(
    URL(string: "https://google.com")!,
    method: HTTPMethod.get,  // Hi, I'm an HTTP GET method
    parameters: [:],         // Hi, I'm an empty dictionary
    encoding: JSONEncoding.default
)
```
Empty brackets `{}` inside the request body make Google to respond with a `400 Bad Request`, by the way.

Your design doesn't give any hints to the end user about the best practices. Instead, you give the green light to the **bad code**.

I simply don't understand your trade-offs, guys. You are talking about flexibility. But hey, someone can invent a completely new HTTP method, say, `UPDATE`, and he [can actually use](https://developer.apple.com/reference/foundation/urlrequest/2011415-httpmethod) this method in scope of his own API. Because method names are basically strings and the SDK doesn't prohibit this. 

But no, you don't go so far, your list of methods is quite limited:
```swift
public enum HTTPMethod: String {
    case options = "OPTIONS"
    case get     = "GET"
    case head    = "HEAD"
    case post    = "POST"
    case put     = "PUT"
    case patch   = "PATCH"
    case delete  = "DELETE"
    case trace   = "TRACE"
    case connect = "CONNECT"
}
```
So where is the line between your ultimate flexibility and solid RFC standards?

It happened that your library inherited most of the `AFNetworking` users, which is a half of the iOS developers out there, but it looks like you refuse to accept any responsibility for your bad design choices, which is really-really sad.

Please, excuse me for disturbing you. I understood your position, hope you will understand mine.

cnoon:
@taflanidi you can create your own custom JSON encoding type that behaves exactly as you would like. This is why we modified the AF4 design to be more flexible when it comes to parameter encoding. While we completely understand your reasoning here, we simply disagree. 

You are also making several contradictory arguments to your own statements here around the design. If you'd like to see things improve in Alamofire, then let's work together on doing so. Simply ranting about bad design and bad code doesn't help you or anyone else. In the future, if you have an idea on how to improve things, please open an issue and let's discuss it before you modify major parts of the framework in something that we ultimately wouldn't agree belongs in Alamofire. We don't want to waste your time and we'd certainly appreciate that you don't waste ours.

PR #1723, opened 2016-10-23
Fix crash in TaskDelegate

### LINKED ISSUE #1708: Crash in TaskDelegate
Randomly have crashes in TaskDelegate from download task on background session on line 335: 

```
        let result = destination(location, downloadTask.response as! HTTPURLResponse)
```

The download task response is nil that's why it's crashing. Is it normal that there is no safe unwrapping ?

jshier:
This PR fixes #1708 by refactoring logic in `TaskDelegate` to properly check for a nil response.

### DIFF
--- a/Source/TaskDelegate.swift
+++ b/Source/TaskDelegate.swift
@@ -331,29 +331,30 @@ class DownloadTaskDelegate: TaskDelegate, URLSessionDownloadDelegate {
     {
         temporaryURL = location
 
-        if let destination = destination {
-            let result = destination(location, downloadTask.response as! HTTPURLResponse)
-            let destination = result.destinationURL
-            let options = result.options
+        guard
+            let destination = destination,
+            let response = downloadTask.response as? HTTPURLResponse
+        else { return }
 
-            do {
-                destinationURL = destination
+        let result = destination(location, response)
+        let destinationURL = result.destinationURL
+        let options = result.options
 
-                if options.contains(.removePreviousFile) {
-                    if FileManager.default.fileExists(atPath: destination.path) {
-                        try FileManager.default.removeItem(at: destination)
-                    }
-                }
+        self.destinationURL = destinationURL
 
-                if options.contains(.createIntermediateDirectories) {
-                    let directory = destination.deletingLastPathComponent()
-                    try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true, attributes: nil)
-                }
-
-                try FileManager.default.moveItem(at: location, to: destination)
-            } catch {
-                self.error = error
+        do {
+            if options.contains(.removePreviousFile), FileManager.default.fileExists(atPath: destinationURL.path) {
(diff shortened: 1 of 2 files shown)


Source/TaskDelegate.swift
```diff
@@ -330,30 +330,29 @@ class DownloadTaskDelegate: TaskDelegate, URLSessionDownloadDelegate {
...
-                    try FileManager.default.createDirectory(at: directory, withIntermediateDirectories: true, attributes: nil)
-                }
-
-                try FileManager.default.moveItem(at: location, to: destination)
-            } catch {
-                self.error = error
+        
+        guard
+            let destination = destination,
+            let response = downloadTask.response as? HTTPURLResponse else { return }
```
cnoon: Could we pull the `else` down a line?

```swift
guard
    let destination = destination,
    let response = downloadTask.response as? HTTPURLResponse 
else { return }
```

Tests/DownloadTests.swift
```diff
@@ -101,6 +101,34 @@ class DownloadResponseTestCase: BaseTestCase {
...
+        
+        waitForExpectations(timeout: timeout, handler: nil)
+        
+        // Then
+        XCTAssertNotNil(response?.request)
+        XCTAssertNil(response?.response)
+        XCTAssertNil(response?.destinationURL)
+        XCTAssertNil(response?.resumeData)
+        XCTAssertNotNil(response?.error)
+    }
```
cnoon: Does this reproduce  the crash reported in #1708? Assuming yes, but thought I'd ask just to make sure.

Source/TaskDelegate.swift
```diff
@@ -330,30 +330,29 @@ class DownloadTaskDelegate: TaskDelegate, URLSessionDownloadDelegate {
...
+        
+        let result = destination(location, response)
+        let destinationURL = result.destinationURL
+        self.destinationURL = destinationURL
+        let options = result.options
```
cnoon: The `destinationURL` assignment is a bit weird tucked in the middle of the other assignments. How about something like this?

```swift
let result = destination(location, response)
let destinationURL = result.destinationURL
let options = result.options

self.destinationURL = destinationURL

do {
```

Tests/DownloadTests.swift
jshier: Turns out it doesn't, as this delegate method doesn't get called with a cancelled download request. I can't find a way to replicate this cause (nil response) in that method. Perhaps it's related to the background session? I can just remove the test and we can land this as a speculative fix if the logic looks good.

cnoon: I couldn't find a way to repro the crash either. The test is useful though so we just as well keep it. 👍🏼

Source/TaskDelegate.swift
cnoon: I made this change and force pushed.

PR #1744, opened 2016-11-02
Use HTTP header's Host to validate if existed

xspyhack:
In a https connection, when the host in requested URL is an *IP address*, the host can’t match with domain during certificate validation, which leads to failure of SSL/TLS hand shake. it’s better to allow user define Host in HTTPS header and this would be helpful for httpdns.

```
IP: 192.168.0.1
Certificate domain: *.example.com

request: https://192.168.0.1/

let serverTrustPolicies: [String: ServerTrustPolicy] = [
    "test.example.com": .performDefaultEvaluation(validateHost: true)
]
```

### DIFF
--- a/Source/SessionDelegate.swift
+++ b/Source/SessionDelegate.swift
@@ -229,7 +229,7 @@ extension SessionDelegate: URLSessionDelegate {
         if let sessionDidReceiveChallenge = sessionDidReceiveChallenge {
             (disposition, credential) = sessionDidReceiveChallenge(session, challenge)
         } else if challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust {
-            let host = challenge.protectionSpace.host
+            let host = (session.configuration.httpAdditionalHeaders?["Host"] as? String) ?? challenge.protectionSpace.host
 
             if
                 let serverTrustPolicy = session.serverTrustPolicyManager?.serverTrustPolicy(forHost: host),
--- a/Source/TaskDelegate.swift
+++ b/Source/TaskDelegate.swift
@@ -102,8 +102,8 @@ open class TaskDelegate: NSObject {
         if let taskDidReceiveChallenge = taskDidReceiveChallenge {
             (disposition, credential) = taskDidReceiveChallenge(session, task, challenge)
         } else if challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust {
-            let host = challenge.protectionSpace.host
-
+            let host = (session.configuration.httpAdditionalHeaders?["Host"] as? String) ?? challenge.protectionSpace.host
+            
             if
                 let serverTrustPolicy = session.serverTrustPolicyManager?.serverTrustPolicy(forHost: host),
                 let serverTrust = challenge.protectionSpace.serverTrust


jshier:
Can you point to any sort of standard the defines the behavior we should see with this?

xspyhack:
**Problem: DNS Pollution**
In some country(like China), domain name hijacking is serious, in some areas even over 20% of the user domain name resolution was hijacked. Carrier's LocalDNS expires NAT according to the authoritative DNS destination IP address or forwards the resolution request to another DNS server, which causes the authoritative DNS can not correctly identify the local IP address of the carrier, and it will cause the domain name resolution error and the traffic inter-network.

**Solution: Connected via IP Address**
All anti-DNS Pollution solutions are Connected via IP Address.
Instead of using the existing UDP-based DNS protocol, the domain name resolution request is sent directly to the server that provides DNS queries (such as [Google Public DNS](https://dns.google.com)), bypassing the local DNS of the carrier and avoiding the domain name caused by the local DNS Hijacking problems and scheduling inaccuracies.

In the HTTPS service scenario, during a handshake the client needs to verify the certificate issued by the server, check the domain and extension fields of the certificate to confirm if the URL's hostname matched the peer's identification hostname. When connected via IP address, the domain mismatch occurs. The SSL/TLS handshake won't success.

So we need to replace the IP address in the verification process of the original domain name.

```swift
let configuration = URLSessionConfiguration.default
var httpHeaders = SessionManager.defaultHTTPHeaders
httpHeaders["Host"] = "test.example.com"
configuration.httpAdditionalHeaders = httpHeaders

let serverTrustPolicies: [String: ServerTrustPolicy] = [
    "test.example.com": .performDefaultEvaluation(validateHost: true)
]

let sessionManager = Alamofire.SessionManager(configuration: configuration, serverTrustPolicyManager: ServerTrustPolicyManager(policies: serverTrustPolicies))

sessionManager.request("https://192.168.0.1/get")
```

[AFNetworking #2954](https://github.com/AFNetworking/AFNetworking/issues/2954)
[Developer Docs](https://developer.apple.com/library/content/technotes/tn2232/_index.html)
[Aliyun HttpDns](https://help.aliyun.com/document_detail/30143.html?spm=5176.doc30102.6.147.o5RLgH)

cnoon:
Hi @xspyhack,

Thank you for all the information here. I didn't realize this was such a problem in China. While I completely understand your approach, this is a bit too specific for us to put directly into Alamofire. What I'd recommend instead is subclassing the `ServerTrustPolicyManager` and creating a mapping between your certificate hostnames and ip address it maps to. Then use the ip address to define the policy to use for evaluation. Here's a quick example of how this could work:

```swift
class IPServerTrustPolicyManager: ServerTrustPolicyManager {
    let hostToIPMap: [String: String]

    init(policies: [String: ServerTrustPolicy], hostToIPMap: [String: String]) {
        self.hostToIPMap = hostToIPMap
        super.init(policies: policies)
    }

    override func serverTrustPolicy(forHost host: String) -> ServerTrustPolicy? {
        guard let ip = hostToIPMap[host] else { return nil }
        return policies[ip]
    }
}
```

Then you can use it like so:

```swift
class BaseTestCase: XCTestCase {
    func testPolicyManager() {
        let manager = IPServerTrustPolicyManager(
            policies: [
                "192.168.0.1": .performDefaultEvaluation(validateHost: true)
            ],
            hostToIPMap: [
                "*.example.com": "192.168.0.1",
                "test.example.com": "192.168.0.1",
            ]
        )

        let sessionManager = SessionManager(serverTrustPolicyManager: manager)

        // TODO: make some requests...
    }
}
```

Hopefully that all makes sense. Please let me know if that won't actually work for your use case, but I'm pretty confident that it will.

Cheers. 🍻

xspyhack:
Hi @cnoon 

Thanks for your example. But unfortunately, it doesn't work.

In my use case the biggest problem is that the domain name in the certification didn’t match  `challenge.protectionSpace.host`, which leads to SSL handshake fail, but not to find the specific  ServerTrustPolicy.

So in your example, `trustIsValid(_:)` method and `evaluate(_:forHost:)` method in `ServerTrustPolicy` will return false.

And in DNS Pollution, we get IP address from remote server. it maybe change sometime.

🙏🍻

cnoon:
Okay @xspyhack I think you want to read between the lines here a bit. My example was simply showing you how you could start to provide a more robust lookup system. These use cases are way outside of what Alamofire has committed to supporting directly, but Alamofire does not stop you from being able to handle these cases.

You'll need to create a custom `ServerTrustPolicyManager` to allow you to map the `host` to the `ServerTrustPolicy` that you want to use. This might be pretty complicated in your use case. From there, you need to customize the evaluation as necessary by toggling certain evaluations. For example, `.performDefaultEvaluation(validateHost: false)` would probably be what you're looking for. Although you obviously have to be REALLY careful about these types of things since you can really be opening yourself up.

You could also switch over to using your own custom evaluation based on your use case. Overall, you have all the tools in front of you. Alamofire will not support such a complex use case directly, but it absolutely supports the ability for your to extend the base functionality to accomplish this.

Best of luck! 🍻

PR #1774, opened 2016-11-13
Feature - Backing up Deployment Targets to iOS 8.0 and macOS 10.10

cnoon:
This PR reverts back to deployment targets of iOS 8.0 and macOS 10.10 due to Apple dropping Swift 2.3 support post Xcode 8.2. We still need to update some of the documentation to support this change, but all the test suites are passing on all supported platforms.


jjulian:
The podspec needs the ios and macos targets to be updated to 8.0 and 10.10

cnoon:
Great catch @jjulian! Fixed in a352f1fb.

PR #1775, opened 2016-11-13
Properly handle nil Data from HEAD requests

jshier:
This PR adds support for treating `HEAD` requests as successful when they have `nil` response data. It also refactors our existing treatment of `nil` data response codes and adds tests for this new behavior.

### DIFF
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -109,7 +109,7 @@ public enum AFError: Error {
     /// The underlying reason the response serialization error occurred.
     ///
     /// - inputDataNil:                    The server response contained no data.
-    /// - inputDataNilOrZeroLength:        The server response contained no data or the data was zero length.
+    /// - inputDataZeroLength:             The server response data was zero length.
     /// - inputFileNil:                    The file containing the server response did not exist.
     /// - inputFileReadFailed:             The file containing the server response could not be read.
     /// - stringSerializationFailed:       String serialization failed using the provided `String.Encoding`.
@@ -117,7 +117,7 @@ public enum AFError: Error {
     /// - propertyListSerializationFailed: Property list serialization failed with an underlying system error.
     public enum ResponseSerializationFailureReason {
         case inputDataNil
-        case inputDataNilOrZeroLength
+        case inputDataZeroLength
         case inputFileNil
         case inputFileReadFailed(at: URL)
         case stringSerializationFailed(encoding: String.Encoding)
@@ -420,8 +420,8 @@ extension AFError.ResponseSerializationFailureReason {
         switch self {
         case .inputDataNil:
             return "Response could not be serialized, input data was nil."
-        case .inputDataNilOrZeroLength:
-            return "Response could not be serialized, input data was nil or zero length."
+        case .inputDataZeroLength:
+            return "Response could not be serialized, input data was zero length."
         case .inputFileNil:
             return "Response could not be serialized, input file was nil."
         case .inputFileReadFailed(let url):
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -253,18 +253,19 @@ extension DownloadRequest {
 extension Request {
     /// Returns a result data type that contains the response data as-is.
     ///
+    /// - parameter request:  The request performed.
     /// - parameter response: The response from the server.
     /// - parameter data:     The data returned from the server.
     /// - parameter error:    The error already encountered if it exists.
     ///
     /// - returns: The result data type.
-    public static func serializeResponseData(response: HTTPURLResponse?, data: Data?, error: Error?) -> Result<Data> {
+    public static func serializeResponseData(request: URLRequest?, response: HTTPURLResponse?, data: Data?, error: Error?) -> Result<Data> {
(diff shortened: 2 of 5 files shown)


Source/AFError.swift
```diff
@@ -109,15 +109,15 @@ public enum AFError: Error {
...
+    /// - inputDataZeroLength:             The server response data was zero length.
     /// - inputFileNil:                    The file containing the server response did not exist.
     /// - inputFileReadFailed:             The file containing the server response could not be read.
     /// - stringSerializationFailed:       String serialization failed using the provided `String.Encoding`.
     /// - jsonSerializationFailed:         JSON serialization failed with an underlying system error.
     /// - propertyListSerializationFailed: Property list serialization failed with an underlying system error.
     public enum ResponseSerializationFailureReason {
         case inputDataNil
-        case inputDataNilOrZeroLength
+        case inputDataZeroLength
```
cnoon: This isn't a backwards compatible change @jshier...we can't make this type of change until AF5.

Source/ResponseSerialization.swift
```diff
@@ -253,18 +253,19 @@ extension DownloadRequest {
...
     /// Returns a result data type that contains the response data as-is.
     ///
+    /// - parameter request:  The request performed.
     /// - parameter response: The response from the server.
     /// - parameter data:     The data returned from the server.
     /// - parameter error:    The error already encountered if it exists.
     ///
     /// - returns: The result data type.
-    public static func serializeResponseData(response: HTTPURLResponse?, data: Data?, error: Error?) -> Result<Data> {
+    public static func serializeResponseData(request: URLRequest?, response: HTTPURLResponse?, data: Data?, error: Error?) -> Result<Data> {
```
cnoon: This isn't a backwards compatible change @jshier. We'll have to find another way to do this that doesn't change a method that can be subclassed.

Source/ResponseSerialization.swift
```diff
@@ -712,5 +727,32 @@ extension DownloadRequest {
...
 
-/// A set of HTTP response status code that do not contain response data.
-private let emptyDataStatusCodes: Set<Int> = [204, 205]
+// MARK: - Empty Data Handling
+
+extension Request {
+    /// A set of HTTP response status code that do not contain response data.
+    fileprivate static let emptyDataStatusCodes: Set<Int> = [204, 205]
+
+
```
cnoon: Extra newline.

Source/ResponseSerialization.swift
```diff
@@ -712,5 +727,32 @@ extension DownloadRequest {
...
+    /// A set of `HTTPMethod`s that are not required to contained response data.
+    fileprivate static let emptyDataHTTPMethods: Set<HTTPMethod> = [.head]
+
+
```
cnoon: Extra newline.

Tests/TLSEvaluationTests.swift
```diff
@@ -414,7 +422,15 @@ class TLSEvaluationExpiredLeafCertificateTestCase: BaseTestCase {
...
+    #if os(iOS) || os(macOS)
+        if #available(iOS 10.1, macOS 10.12.0, *) {
+            XCTAssertNotNil(error, "error should not be nil")
+        } else {
+            XCTAssertNil(error, "error should be nil")
+        }
+    #else
         XCTAssertNil(error, "error should be nil")
+    #endif
     }
```
cnoon: I don't really understand the diffs here. This is all already in the `master` branch. The only thing that isn't is that you've changed `iOS 10.0` to `10.1` from what I can tell. Do you have any insight here?

jshier: I branched from that commit, but for some reason my branch now includes it. Removing the commit should get rid of the spurious changes.

Source/ResponseSerialization.swift
jshier: Unfortunately I don't see any other way to check the request type without passing the `URLRequest` into the serializer.

Source/AFError.swift
jshier: Well, I thought I had mentioned it. But I can revert this part of the changes, it just means that reason is a bit misleading.

Tests/TLSEvaluationTests.swift
jshier: I also fixed the `macOS` check, as it was just `(macOS)` before, which doesn't work but for some reason compiles.

cnoon: Cool.

Source/AFError.swift
cnoon: That's my bad if you did. I'd say let's punt this to post 4.1.0. Maybe we can figure out a backwards compatible way to handle this feature before AF5.

cnoon:
Hey @jshier, I had a chance to review this more in-depth and I'm afraid we'll have to wait until AF5 to make these changes. Are there any changes in here that are backwards compatible that should still make it into AF 4.x? I remember you saying you found a couple small issues that you thought were incorrect. Maybe with the tests?

I'm going to close this PR out for now. We can revisit in AF 5. I've added a card to our Trello board.

PR #1792, opened 2016-11-18
Fix URLRequest not always being available publicly

### LINKED ISSUE #1785: Failing URLRequest not available to RequestRetrier
I’m in the process of updating to 4.1.0 to adopt the new behavior of the request retried and adapter introduced with #1682. Unfortunately I ran into some troubles. The intention was to make sure we refresh the access token before making any requests if it has expired. So I made the our `RequestAdapter` throw a `missingAccessToken` if the token is expired. That’s when the troubles started.

For context; we have a setup which allows different parts of the app to provide authorizers to the network layer for specific hosts. As such I need to check which host a request was for in order to tell the associated authorizer to refresh its access token.

When the adapter throws the error for a request that hasn’t touched the actual network we can’t actually refresh the token. Because the `Request` object given the `RequestRetrier` returns `nil` for both the `request` and `task` properties. As such I can’t check which host the request is going to try and reach. However the `Request` object does in fact have a reference to the `URLRequest` via the original data task. That property is internal to Alamofire.

To work around this I decided to extend our `missingAccessToken` error case with the host, ala:
```swift
public enum AuthorizerError: Error {
    case missingAccessToken(host: String)
}
```

This doesn’t work either though since `AdaptError` is also internal 😄 As such I can’t check if the `Error` object given to the retrier is of that type and then extract the underlying error. The same is true for the `underlyingAdaptError` category on `Error` so I’m unable to use that.

As such I’m stuck with not knowing anything about the request that failed. Or am I missing something obvious?

---

To remedy this I’d like to propose a few solutions:
- Change so that the `request` property “always<sup>1</sup>” returns the `URLRequest` associated with it. This could be accomplished
  - Change so that `Request.task` returns `originalTask` as a `URLSessionTask` if the `delegate.task` property returns `nil`.
  - Change so that `Request.urlRequest` returns `originalTask`’s `originalRequest` value if `task?.originalRequest` is `nil`.
- Change `AdaptError` and its `error` property to be public.
- Change `Error.underlyingAdaptError` to be public.

Obviously several of these changes could made at the same time to support different use cases. Which one do you prefer? Or is there some other solution you’d rather have? I’m open to submitting a PR.

---

1. I.e. when there is a `URLRequest`, obviously not for certain types of requests which doesn’t have one.

rastersize:
This PR aims to resolve #1785 by first trying to return the `URLRequest` from the task. If it’s `nil` we’ll attempt to return `URLRequest` associated with the `originalTask`, if there is one. As per the https://github.com/Alamofire/Alamofire/issues/1785#issuecomment-261284279.

I’ve also made `Error.underlyingAdaptError` public so that request re-triers can check any underlying adapt errors. I opted to not make the error type public as it might add confusion of what type of error `RequestAdapter`s should throw.

Hopefully this is in line with Alamofire’s style and with what you where thinking @cnoon. Let me know if you want me to make any changes.

🧀🍷

### DIFF
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -139,7 +139,7 @@ struct AdaptError: Error {
 }
 
 extension Error {
-    var underlyingAdaptError: Error? { return (self as? AdaptError)?.error }
+    public var underlyingAdaptError: Error? { return (self as? AdaptError)?.error }
 }
 
 // MARK: - Error Booleans
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -365,6 +365,13 @@ open class DataRequest: Request {
 
     // MARK: Properties
 
+    open override var request: URLRequest? {
+        if let request = super.request { return request }
+        if let requestable = originalTask as? Requestable { return requestable.urlRequest }
+
+        return nil
+    }
+
     /// The progress of fetching the response data from the server for the request.
     open var progress: Progress { return dataDelegate.progress }
 
@@ -465,6 +472,16 @@ open class DownloadRequest: Request {
 
     // MARK: Properties
 
+    open override var request: URLRequest? {
+        if let request = super.request { return request }
+
+        if let downloadable = originalTask as? Downloadable, case .request(let urlRequest) = downloadable {
+            return urlRequest
+        }
+
+        return nil
+    }
+
     /// The resume data of the underlying download task if available after a failure.
     open var resumeData: Data? { return downloadDelegate.resumeData }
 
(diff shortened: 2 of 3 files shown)


rastersize:
The tests seems to have failed due to some test flakiness and Xcode/simulator flakiness. Unfortunately I can’t re-trigger them. Ran the tests locally for all platforms (except watchOS) and they all passed ¯\_(ツ)_/¯

cnoon:
Okay @rastersize, thanks for making the time to put this together! 

I squashed your changes, made a few minor tweaks and pushed them into `master` in commit 19e5012a. I also removed your change to make the `AdaptError` public. That error should definitely not need to be public which led me to investigate some more as to why you thought you needed to do that.

Once I dug in more, I realized that throwing an `Error` in the adapter wasn't working quite right. The retrier was being called correctly, but was being sent the `AdaptError` instead of the underlying error like it was supposed to. The idea is that if you throw an `AuthenticationError.expiredAccessToken` in your adapter, then the retrier error will also be an `AuthenticationError.expiredAccessToken`. You should NEVER see an `AdaptError` in any of the public Alamofire APIs. If you do, it is a bug. I've fixed the underlying issue and added tests to ensure everything is working properly in d2826a11. This should resolve the need for you to ever need the `AdaptError` to be public.

If there's still something I'm missing here, please don't hesitate to reach out. This logic is brand new and it's possible there's still a kink or two. I "think" at this point the logic should be pretty solid. Regardless, if you see anything that seems a bit odd, just assume it is and please file an issue. On a different note, these changes should be released within the next few days.

Thanks again! 🍻

rastersize:
@cnoon Sounds great and what I expected from the API to begin with. Good change 👍

PR #1794, opened 2016-11-19
Updated README for RequestRetrier example

pontus-andersson:
This PR fixes improves the README for the `RequestRetrier`. Some OAuth services returns a result that is not convertable to [String:String]. E.g. some services returns an `expire` timestamp as an integer:

```
{
	"token_type": "Bearer",
	"refresh_token": "******************************",
	"access_token": "******************************",
	"expires_in": 36000,
	"scope": "write read"
}
```

Copying the old example from the README would result in an unsuccessful result when refreshing the access token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1291,8 +1291,8 @@ class OAuth2Handler: RequestAdapter, RequestRetrier {
             .responseJSON { [weak self] response in
                 guard let strongSelf = self else { return }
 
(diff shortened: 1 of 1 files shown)


cnoon:
Good call @pontus-andersson! I made some small formatting tweaks to your PR and pushed them into `master` in ab0c7679 while keeping your attribution.

Thanks again! 🍻

PR #1798, opened 2016-11-22
Changes default in SessionManager to allow it to be overridden.

ricardolpd:
Current default variable in SessionManager access is set to open which made me think that i [could override its behaviour](http://stackoverflow.com/q/40725354/999817), while some @robertmryan at [stackoverflow](http://stackoverflow.com/a/40747649/999817) was quick to point out, that default implementation of class shouldnt be overriden. I do think that part of Alamofire helper methods revolve around the default property. Therefore i think it would be appropriate to be able to override this property in either a subclass or an extension to set TrustPolicy objects, custom delegates, or even just a custom configuration object.

After playing around a little longer with alamofire, i created this small extension which allows to either subclass SessionManager or extend it in a category and continue to use helper methods in Alamofire.swift or even use SessionManager.default, as for me as user default should be something that 'suits the needs of my my app'.

Therefore i wonder if this extension is something that could be of use to comunity and something we could integrate in Alamofire. Please let me know.

### DIFF
--- a/Source/SessionManager.swift
+++ b/Source/SessionManager.swift
@@ -24,11 +24,43 @@
 
 import Foundation
 
+public protocol SessionManagerProtocol {
+
+    associatedtype DefaultType
+    
+    static var `default`:DefaultType {get}
+
+    init(configuration: URLSessionConfiguration,// = URLSessionConfiguration.default,
+        delegate: SessionDelegate,// = SessionDelegate(),
+        serverTrustPolicyManager: ServerTrustPolicyManager?)
+    
+}
+
+public extension SessionManagerProtocol {
+
+    /// A default instance of `SessionManager`, used by top-level Alamofire request methods, and suitable for use
+    /// directly for any ad hoc requests.
+    public static var `default`: SessionManager {
+        let configuration = URLSessionConfiguration.default
+        configuration.httpAdditionalHeaders = SessionManager.defaultHTTPHeaders
+        
+        return SessionManager(configuration: configuration,
+                              delegate: SessionDelegate(),
+                              serverTrustPolicyManager: nil)
+    }
+
+}
+
+
+
+
 /// Responsible for creating and managing `Request` objects, as well as their underlying `NSURLSession`.
-open class SessionManager {
+open class SessionManager: SessionManagerProtocol {
 
-    // MARK: - Helper Types
-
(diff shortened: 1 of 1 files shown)


moonytoes29:
I like it. This would be really useful. I don't like the fact the Almofire doesn't provide enough easy extension points. SOLID design says should be open for extension and closed for mutation. Nice

Graham-Bryan:
I agree, this approach would allow extending the functionality whilst keeping the code under control. +1

jshier:
We're most likely going to decline this PR, as allowing the swapping of `SessionManager`'s could break the entirety of Alamofire, but we are interested in exactly what extension points are desired here. That way we can evaluate where such extension should be happening.

ricardolpd:
@jshier I am not sure i understand what you mean. Alamofire offers different Inits in the SessionManager to override the default behaviour, if you what are saying is such an important behaviour why do these init exist, as far as I can see these inits give the oportunity to "break the entirely of Alamofire".

It seems to me that when a user requires a specific behaviour from alamofire, the user is then responsible for its outcome and for not breaking the framework.

All this pull-request is meant to do, is to redefine as per user requirements the definition of default, and perhaps (i haven't added this) the defaultHeaders property as per the app requirements.
I tend to do work in several corporate places, as per endpoints i connect to, i can see no use of default in its current state. Also i do think that having to subclass SessionManager and create another singleton property just because of this enforcement is a negative experience, as I then loose all Helper request methods in Alamofire.

At least for me and for the type of work i tend to do, i feel that it is important to be able to override the:
- behaviour of SessionConfiguration, 
- add a TrustPolicyManager, 
- or potentially to some just change the default headers.
- sometimes i might even require to change SessionDelegate to stop redirects from occuring.

All these are perfect case scenarios, which could be easily handled by allowing the redefinition of default, and have a default meaning that suits my app needs.

I hope my explanation makes sense. Also if you still prefer to reject this PR, what would be your suggested alternative approach.

ricardolpd:
@jshier I have made some changes to the original implementation where i make sure that the singleton pattern is kept. while allowing to redefine it. It doesnt seem to have updated the pull request.
My question is shall i open a new one, or is this something you guys are not interested at all?? And if not why, as URLSession allows similar behaviour as the one i propose.

PR #1810, opened 2016-11-30
Update README.md

tiagomartinho:
adds missing casting as AnyObject to read me


cnoon:
I just verified that this change is not necessary in Xcode 8.2. If there is something that I'm missing here, please let me know. For now though I'm going to close this out.

Thanks anyways! 🍻

Issue #1839, opened 2016-12-12
Potential Bug for Array serialization with URLEncoding.Default


julien-sarazin:
### Implementation sample: 
``` swift
let parameters = [
			"uid": "12DE3BEB-C088F467E5F1-D1AF-44C8-AD01",
			"memories": [
				["foo": "bar"],
				["foo": "baz"]
			]
		] as [String : Any]
		
		Alamofire.request("http://api/endpoint",
			method: .post,
			parameters: parameters,
			encoding: URLEncoding.default).responseJSON { (response) in
				if let JSON = response.result.value as? [String: Any] {
					print("JSON: \(JSON)")
					let response = NPDResponse(with: JSON)
					return completion(response)
				}
				
				completion(nil)
		}
```

### Actual formatted query :
```
memories[][foo] bar
memories[][foo] baz
uid	12DE3BEB-C088F467E5F1-D1AF-44C8-AD01
```

### Expected formatted query : 
```
memories[0][foo] bar
memories[1][foo] baz
uid	12DE3BEB-C088F467E5F1-D1AF-44C8-AD01
```

cnoon:
There's no rfc spec around collection URL encoding. What you're seeing in Alamofire is our default implementation. If you need the indices, you'll need to create your own custom `ParameterEncoding` to add the indices. This can be accomplished by using the `URLEncoding` implementation as a guide.

For more info on creating a custom `ParameterEncoding`, please refer to our [example](https://github.com/Alamofire/Alamofire#custom-encoding) in the README. For more info on the current `URLEncoding` implementation and reasoning, please refer to the [docstrings](https://github.com/Alamofire/Alamofire/blob/master/Source/ParameterEncoding.swift#L62-L69).

Cheers. 🍻

PR #1840, opened 2016-12-12
added extra parameter to maintain index when parameters are Array.

julien-sarazin:
Potential fix for issue #1839.


cnoon:
Closing with the same response as #1839.

PR #1844, opened 2016-12-14
update README.md to fix URLConvertible example issue

zhongwuzw:
update README.md to fix URLConvertible example issue


cnoon:
Thanks @zhongwuzw for putting this together. I made a small tweak and pushed it up into `master` in 05c9bcdc. It will go out in the 4.3.0 release most likely. Thanks again! 🍻

Issue #1853, opened 2016-12-15
RequestRetrier timeDelay not honored after encountering AdaptError


ejensen:
In `SessionManager.allowRetrier()` the [timeDelay supplied by the request retrier](https://github.com/Alamofire/Alamofire/blob/master/Source/SessionManager.swift#L869) is not honored. 
This code path is hit when a request fails due to an `AdaptError`.
Alamofire v4.2.0

cnoon:
I've verified this is in fact a bug. Should be a simple fix.

cnoon:
I just put up PR #1864 to address this issue. I'm going to close this issue out. Please direct all future comments to #1864. Thanks @ejensen for filing! 🍻

PR #1864, opened 2016-12-19
Fix - SessionManager Retry Delay

cnoon:
This PR addresses the issue pointed out in #1853 where the `SessionManager` internal retry logic for `AdaptError` instances was not respecting the time delay. This small change now respects the delay returned by the retrier.

### DIFF
--- a/Source/SessionDelegate.swift
+++ b/Source/SessionDelegate.swift
@@ -471,10 +471,10 @@ extension SessionDelegate: URLSessionTaskDelegate {
         /// If an error occurred and the retrier is set, asynchronously ask the retrier if the request
         /// should be retried. Otherwise, complete the task by notifying the task delegate.
         if let retrier = retrier, let error = error {
-            retrier.should(sessionManager, retry: request, with: error) { [weak self] shouldRetry, delay in
+            retrier.should(sessionManager, retry: request, with: error) { [weak self] shouldRetry, timeDelay in
                 guard shouldRetry else { completeTask(session, task, error) ; return }
 
-                DispatchQueue.utility.after(delay) { [weak self] in
+                DispatchQueue.utility.after(timeDelay) { [weak self] in
                     guard let strongSelf = self else { return }
 
                     let retrySucceeded = strongSelf.sessionManager?.retry(request) ?? false
--- a/Source/SessionManager.swift
+++ b/Source/SessionManager.swift
@@ -874,12 +874,16 @@ open class SessionManager {
                     return
                 }
 
-                let retrySucceeded = strongSelf.retry(request)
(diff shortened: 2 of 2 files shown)


Source/SessionManager.swift
```diff
@@ -874,12 +874,16 @@ open class SessionManager {
                     return
                 }
 
-                let retrySucceeded = strongSelf.retry(request)
+                DispatchQueue.utility.after(timeDelay) {
```
jshier: I'm unclear as to whether we should be executing this code on a separate queue or the same queue that originally executed the request. Wouldn't this logic create a new queue for every retry attempt? 

Perhaps at some point one of us should create a runtime diagram that outlines the various queues used...

Source/SessionDelegate.swift
```diff
@@ -471,10 +471,10 @@ extension SessionDelegate: URLSessionTaskDelegate {
         /// If an error occurred and the retrier is set, asynchronously ask the retrier if the request
         /// should be retried. Otherwise, complete the task by notifying the task delegate.
         if let retrier = retrier, let error = error {
-            retrier.should(sessionManager, retry: request, with: error) { [weak self] shouldRetry, delay in
+            retrier.should(sessionManager, retry: request, with: error) { [weak self] shouldRetry, timeDelay in
```
jshier: This rename doesn't seem necessary, as the previous name wasn't ambiguous (aren't all delays in time?). But it doesn't necessarily need to be changed.

cnoon: The only reason I changed it was to match the "documented" name of the closure parameter. I also thought it would be good to match the same name in the `SessionManager`.

Source/SessionManager.swift
cnoon: Great question. All this logic does is dispatch after a given interval onto the utility queue. We're not creating a new queue. The utility queue is concurrent and is merely a thread pool that will execute once it has resources. Generally this will happen immediately, unless you have multiple higher priority operations running.

None of this retry logic executes on the calling thread. If you create a request that fails to be adapted (for the sake of argument let's say your OAuth token is expired), then the retry logic hops off the calling thread in the `allowRetrier(_:toRetryRequest:withError:)` API. Then all the retry logic is executed on the utility queue. Once the completion is called (we don't know what queue we'll be called on), then we dispatch to utility once more to allow the delay to pass before retrying. The retry logic simply re-adapts the request and hands the task over to the session to execute. The new task will then be executed on the URLSession's internal queue and all callbacks will again execute on the SessionDelegate queue.

I think putting together a queue diagram is a great idea and is something we should add to the Trello backlog.

jshier:
:+1:

PR #1865, opened 2016-12-20
Feature - Opening up Default Response Initializers

cnoon:
This PR modifies the `DefaultDataResponse` and `DefaultDownloadResponse` initializers to be `public` instead of `internal`. This allows you to rebuild the responses if necessary in client code to do things like wrap the `error` in your own custom error or modify the server data or possibly inject a missing header, etc. The main use case is to wrap the `error`.

This PR also adds convenience properties to the `DataResponse` and `DownloadResponse` structs for accessing the `value` and `error` in the result. Both of these computed properties are merely for convenience.

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -87,6 +87,12 @@ public struct DataResponse<Value> {
     /// The timeline of the complete lifecycle of the request.
     public let timeline: Timeline
 
+    /// Returns the associated value of the result if it is a success, `nil` otherwise.
+    public var value: Value? { return result.value }
+
+    /// Returns the associated error value if the result if it is a failure, `nil` otherwise.
+    public var error: Error? { return result.error }
+
     var _metrics: AnyObject?
 
     /// Creates a `DataResponse` instance with the specified parameters derived from response serialization.
@@ -219,6 +225,12 @@ public struct DownloadResponse<Value> {
 
     /// The timeline of the complete lifecycle of the request.
     public let timeline: Timeline
+
+    /// Returns the associated value of the result if it is a success, `nil` otherwise.
+    public var value: Value? { return result.value }
+
+    /// Returns the associated error value if the result if it is a failure, `nil` otherwise.
+    public var error: Error? { return result.error }
 
     var _metrics: AnyObject?
 
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -43,7 +43,23 @@ public struct DefaultDataResponse {
 
     var _metrics: AnyObject?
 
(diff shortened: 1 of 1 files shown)


groue:
> This allows you to rebuild the responses if necessary in client code to do things like wrap the error in your own custom error or modify the server data or possibly inject a missing header, etc. The main use case is to wrap the error.

Side note: https://github.com/Alamofire/Alamofire/pull/1836 also addresses error handling: DataResponse.map and DataResponse.flatMap do recreate new responses that wrap a new result or a new error, but without exposing public initializers:

```swift
Alamofire.request("https://httpbin.org/get").responseJSON { response in
    // Here response is DataResponse<[String:Any]>,
    // and stuffResponse is DataResponse<Stuff>:
    let stuffResponse = response.flatMap { json in try Stuff(json: json) }
}
```

It's not that I want to push my #1836 at all cost. I just want to say that it addresses response transformations in a way that prevents the complex initializers from bubbling up into the public API.

Besides, preserving responses' metrics is only possible with internal APIs (see https://github.com/Alamofire/Alamofire/pull/1836/files#diff-0532e61b339db1c06287f9110afe7dfcR147).

jshier:
@groue I think the functional approach your PR creates, and I'm sure many developers will be comfortable there, but there is an argument for the straightforward use of an initializer. Depending on whatever else there doing, either approach may be more readable than the other.

groue:
@jshier That's perfectly clear. I just saw an opportunity to explain some uses of the functional approach which can be overlooked - since #1836 had no feedback whatsoever. Now I'm *not* a functional guru/evangelist/lobbyist - I'm a regular imperative developer, and certainly don't want to force abstract functional cathedrals down into anyone's throat. #1836 is actually pretty narrow.

Now the use cases exposed by @cnoon are *not* straightforward: injecting headers, modifying server data, etc. That's looks pretty advanced to me.

On the other side, map/flatMap are a concise syntax for a very common use case, which is to turn server data into application model types, while preserving errors and response metadata. The map/flatMap idioms are, I think, simpler than extensions that [custom response serialization](https://github.com/Alamofire/Alamofire#custom-response-serialization) requires today. And they blend naturally in the map/flatMap landscape that Swift already brings with optionals and collections.

Maybe give a look to #1836 some day?

cnoon:
Thanks for the feedback @groue! I agree with @jshier that both are useful. Apologies for your PR being left out there without feedback. We got pretty busy with the holidays and haven't been able to keep up with all the great work coming in.

PR #1880, opened 2016-12-29
Add the request method to debugDescription.

jshier:
This PR adds the request method to the `DataResponse` and `DownloadResponse` `debugDescriptions`.


cnoon:
I made a small tweak to this @jshier to fallback on `GET` rather than force unwrap the `httpMethod`. I pushed it up into `master` in 3f8575ad.

PR #1185, opened 2016-04-14
Add missing parameter for failing test in URLProtocolTests

edopelawi:

PR #1883, opened 2016-12-30
Add QueryURL wrapper class and test for url/body parameter combination

Dschee:
This adds the feature requested in #374 to Alamofire in a backwards-compatible and (hopefully) clean way. I'd love to hear feedback on this approach. The usage would be as simple as this:

```swift
let queryUrl = QueryURL(url: urlString, parameters: ["lorem": "ipsum"])
let bodyParameters = ["foo": "bar"]
Alamofire.request(queryUrl, method: .post, parameters: bodyParameters, encoding: JSONEncoding.default).response // ...
```

Progress:

- [x] Implement support for both query and body parameters
- [x] Write test to ensure the feature survives future changes
- [ ] Add a section to the README explaining this feature

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -64,6 +64,37 @@ extension URLComponents: URLConvertible {
     }
 }
 
+/// Wrapper around URLConvertible to force-include query parameters on body-enabled requests.
+/// Use this if you want to pass body parameters and query paramters in one request.
+public class QueryURL: URLConvertible {
+    let url: URLConvertible
+    let parameters: Parameters
+
+    /// Creates a new QueryURL object with given url and query parameters.
+    ///
+    /// - parameter url:        The URL.
+    /// - parameter parameters: The parameters to be included into the URL as query params.
+    public init(url: URLConvertible, parameters: Parameters) {
+        self.url = url
+        self.parameters = parameters
+    }
+}
+
+extension QueryURL {
+    /// Returns a URL that conforms to RFC 2396 or throws an `Error`.
+    ///
+    /// - throws: An `Error` if the type cannot be converted to a `URL`.
+    ///
+    /// - returns: A URL or throws an `Error`.
+    public func asURL() throws -> URL {
+        let url = try self.url.asURL()
+        let fakeUrlRequest = URLRequest(url: url)
+        let queryFakeUrlRequest = try URLEncoding.default.encode(fakeUrlRequest, with: self.parameters)
+        guard let queryUrl = queryFakeUrlRequest.url else { throw AFError.invalidURL(url: self) }
+        return queryUrl
+    }
+}
+
 // MARK: -
 
 /// Types adopting the `URLRequestConvertible` protocol can be used to construct URL requests.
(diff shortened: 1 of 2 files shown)


jshier:
Thanks, we'll take a look @Dschee. However, you can already accomplish this by using the `URLEncoding` `ParameterEncoding` on the `URLRequest` and then passing that `URLRequest` into `request()` as you're doing, where the body will be `JSON` encoded. So really this is just a slightly more convenient version of that. Perhaps a more general solution would be to create a `ParameterEncoding` that can take sets of encodings and parameters and applies them to a request all at once?

AnthonyMDev:
This is an interesting approach, but I agree that this is similar to what is already achievable.
@jshier That was one of the solutions I was considering actually. The problem is that you can still pass parameters into the `request()` functions. So if you pass an `encoding` of this new multiple encoding type with parameters given to it, and you also pass parameters to the `request()` function, what happens then? The only way this would make sense to me would be if we removed the `parameters` parameter from the `request()` functions and instead made people pass the parameters to the `ParameterEncoding` struct. But this seems to complicate the API a bit.

Another option I have considered is adding a `queryParameters` parameter to all request creation methods. This would make `URLEncoding.Destination.queryString` redundant, so we would likely want to deprecate it. Since we would be using `queryParameters` for encoding in the url, `URLEncoding` would now only be used to setting the query string in the `httpBody` This seems simple to implement and easy to use, however it may make the `parameter`parameter confusing. Maybe we could rename it to `bodyParameters`. This would require a major version release and an API breaking change, which isn't ideal.

Example:

````swift
    open func request(
        _ url: URLConvertible,
        method: HTTPMethod = .get,
        queryParameters: Parameters? = nil,
        encoding: ParameterEncoding = URLEncoding.default,
        headers: HTTPHeaders? = nil)
        -> DataRequest
````

**Comment by Dschee** [comment]
I'm not a fan of the idea of **changing** the request methods parameters to include `queryParameters` and `bodyParameters`, it communicates to the API user that he should split his data into query and body parameters – which may not be needed for many APIs and just adds clutter to the method interface for all others. So I think the old method should be kept.

Instead I suggest adding a new **convenience method**. This would result to something like this:
```swift
// new convenience method, note that I also changed 'encoding' to 'bodyEncoding'
open func request(
        _ url: URLConvertible,
        method: HTTPMethod = .get,
        queryParameters: Parameters? = nil,
        bodyParameters: Parameters? = nil,
        bodyEncoding: ParameterEncoding = URLEncoding.default,
        headers: HTTPHeaders? = nil)
        -> DataRequest

// old method, unchanged
open func request(
        _ url: URLConvertible,
        method: HTTPMethod = .get,
        queryParameters: Parameters? = nil,
        encoding: ParameterEncoding = URLEncoding.default,
        headers: HTTPHeaders? = nil)
        -> DataRequest
```

Dschee:
I've just posted #1185 which implements my recent suggestion. What do you think?

cnoon:
Hi @Dschee,

Thanks for taking the time to put this together. I'm afraid this addition isn't something we'd like to have in Alamofire. It is confusing to parameters going into both your `QueryURL` and into the top level APIs.

With that said, feel free to use this all over in your own projects. There's nothing wrong with doing what you are doing, it could just lead to misuse and confusion among the AF community.

Thanks again. 🍻

PR #1885, opened 2016-12-31
Add request/download methods allowing query/body parameter combinations

Dschee:
This is a follow-up on #1883 and implements my second approach explained in https://github.com/Alamofire/Alamofire/pull/1883#issuecomment-269853832.

Progress:
- [x] Add new request method which accepts both query and body parameters
- [x] Make the existing request method a convenience method for the new one
- [ ] Add new download method which accepts both query and body parameters
- [ ] Make existing download method a convenience method for the new one
- [ ] Add tests to cover the new method use cases
- [ ] Document the new method in the README

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -146,6 +146,37 @@ public func request(
     )
 }
 
+/// Creates a `DataRequest` using the default `SessionManager` to retrieve the contents of the specified `url`,
+/// `method`, `queryParameters`, `bodyParameters`, `bodyEncoding` and `headers`.
+///
+/// - parameter url:                The URL.
+/// - parameter method:             The HTTP method. `.get` by default.
+/// - parameter queryParameters:    The query parameters. `nil` by default.
+/// - parameter bodyParameters:     The body parameters. `nil` by default.
+/// - parameter encoding:           The body parameter encoding. `URLEncoding.default` by default.
+/// - parameter headers:            The HTTP headers. `nil` by default.
+///
+/// - returns: The created `DataRequest`.
+@discardableResult
+public func request(
+    _ url: URLConvertible,
+    method: HTTPMethod = .get,
+    queryParameters: Parameters?,
+    bodyParameters: Parameters? = nil,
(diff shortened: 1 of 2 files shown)


jshier:
@Dschee I appreciate your work on this. However, I'm sure this scenario is common enough to need top level APIs dedicated to it. Once Alamofire usage has reached this level of complexity (covering an entire JSONAPI setup), users should really be adopting a variation of the `Router` pattern. This can fairly easily cover the scenarios you require by using multiple encoders. Ultimately, it seems like there are few options here:

1. Add top level API for encoding both URL and body parameters into a request, as demonstrated in this PR.
2. Add a specialized `ParameterEncoding` conforming type that can take both url and body parameters as well as their encodings and generates a request.
3. Both 1 and 2.
4. Nothing, and let users add their own particular implementations.

I think perhaps 2 is the way to go if we want to support this, as I really don't think this is common enough to need the additional complexity of adding and refactoring the top APIs for it.

@cnoon, what do you think? I'm not particularly sold on any solution (or not) here.

cnoon:
Hi @Dschee,

Thanks again for your efforts here. We greatly appreciate new ideas on how to make improvements. @jshier talked it through while we like some of the ideas here, we don't think they have a place directly in Alamofire for a couple different reasons.

The first reason is that the top level APIs are designed for convenience for the common use cases. We try to have as few as possible to make it easy to decide what you're looking for. As soon as you start adding query parameters, query encoding and http body parameters in the same top level APIs, you're outside of the normal cases for sure. So that's the first problem. We don't want another top level API. Not to mention a top level API that needs two more parameters.

The second problem is that even if we did allow the API with the additional parameters, it's really easy to misuse the encodings. If you mix up either of the encodings or the parameters you're in trouble and it won't be very easy to debug what's going on. For example, if you accidentally use `URLEncoding.default` for your query parameters (which I believe you did you your PR), you're in for a world of hurt with a `.put` or `.post` because the parameters will be encoded in the body in those cases then potentially replaced by the body parameter encoded data.

Overall, we don't really want to add another top level API, and if we did, it would need to be very difficult to misuse.

Thanks again. 🍻

Issue #1544, opened 2016-09-15
Build error with Swift Package Manager


skyline75489:
I'm using Xcode 8 and the official SPM.  I try to use Alamofire through SPM:

```swift
import PackageDescription

let package = Package(
    name: "swift-playground",
    dependencies: [
        .Package(url: "https://github.com/Alamofire/Alamofire.git", majorVersion: 4)
    ]
)
```

However, `swift build` gives me a lot of errors:

```plaintext
Cloning https://github.com/Alamofire/Alamofire.git
HEAD is now at 5fe5b20 Added release notes to the CHANGELOG and bumped the version to 4.0.0.
Resolved version: 4.0.0
Compile Swift Module 'Alamofire' (17 sources)
/Users/skyline/Projects/Github/swift-playground/Packages/Alamofire-4.0.0/Source/DispatchQueue+Alamofire.swift:33:25: error: use of undeclared type 'TimeInterval'
    func after(_ delay: TimeInterval, execute closure: @escaping () -> Void) {
                        ^~~~~~~~~~~~
/Users/skyline/Projects/Github/swift-playground/Packages/Alamofire-4.0.0/Source/SessionDelegate.swift:111:50: error: 'URLSessionStreamTask' is only available on OS X 10.11 or newer
    open var streamTaskReadClosed: ((URLSession, URLSessionStreamTask) -> Void)?
                                                 ^
/Users/skyline/Projects/Github/swift-playground/Packages/Alamofire-4.0.0/Source/SessionDelegate.swift:111:50: note: add @available attribute to enclosing class
    open var streamTaskReadClosed: ((URLSession, URLSessionStreamTask) -> Void)?
                                                 ^
/Users/skyline/Projects/Github/swift-playground/Packages/Alamofire-4.0.0/Source/SessionDelegate.swift:114:51: error: 'URLSessionStreamTask' is only available on OS X 10.11 or newer
```

jshier:
From those errors it appears you aren't using Xcode 8. Alamofire 4 requires Xcode 8. If you need to keep using Xcode 7, please use Alamofire 3.5.0.

skyline75489:
I'm indeed using Xcode 8 and Swift 3. I think there may be some kind of bug in SPM.

Siceberg:
I used Xcode8 too, got the same errors
(https://cloud.githubusercontent.com/assets/8327777/18612992/2221341e-7da0-11e6-9f58-73a57e8451d0.png)

[...]

PR #1905, opened 2017-01-12
Added the dependencies key since it's now needed with SPM

cyberdude:
This is related to this issue #1544   I'm using swift for Ubuntu with Perfect.org


jshier:
@cyberdude This looks fine. Is there someplace where changes like this are tracked? I'm not seeing anything in the SPM docs about it being required now.

PR #1916, opened 2017-01-15
URLString -> URL

fulldecent:


cnoon:
Thanks @fulldecent...that was definitely wrong. I had to tweak your changes a bit and pushed them up into `master` in f3e250ab while keeping your attribution. This should go out shortly as part of the Alamofire 4.3.0 release.

Thanks! 🍻

PR #1930, opened 2017-01-23
fix error of a unescaped logging for symbol "$" in headers

akkuprin:
if request header contains "$" symbol, then debugDescription (cUrlRepresentation)  returns incorrect value: header contains not escaping value for this symbol, and on server we have header with incorrect value (it includes only part before "$" symbol).

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -327,7 +327,8 @@ extension Request: CustomDebugStringConvertible {
         }
 
         for (field, value) in headers {
-            components.append("-H \"\(field): \(value)\"")
+            let escapedValue = "\(value)".replacingOccurrences(of: "$", with: "\\$")
+            components.append("-H \"\(field): \(escapedValue)\"")
         }
 
         if let httpBodyData = request.httpBody, let httpBody = String(data: httpBodyData, encoding: .utf8) {


cnoon:
Hi @akkuprin,

This is not a change that should be part of Alamofire itself. The `$` character is a supported ASCII character which [should be supported](http://stackoverflow.com/a/4410331/1342462) by servers. Here's a quick example of sending headers with `$` embedded in both the header fields and values.

```swift
func testDollarSignsInHeaders() {
    let expectation = self.expectation(description: "request should succeed")
    let headers: HTTPHeaders = [
        "foo": "b$r",
        "currency": "$for$dollars",
        "$momoney$": "$$$cash$$$"
    ]

    Alamofire.request("https://httpbin.org/headers", headers: headers).responseJSON { response in
        debugPrint(response)
        expectation.fulfill()
    }

    waitForExpectations(timeout: 20, handler: nil)
}
```

What you should do instead in this case is pre-encode your headers before adding them to Alamofire. If they need to be escaped, it needs to be you who does this because Alamofire can't know whether or not you need to actually escape the `$` symbol.

Thanks anyways! 🍻

PR #1969, opened 2017-02-16
Return the request's state

vietstone:
As title.


cnoon:
Hi @vietstone,

Thanks for putting this PR together...much appreciated! With that said, I don't really see the use in duplicating the entire task state at the request level. Could you speak to that a bit? I think we'd all need a really good reason to justify adding this logic into Alamofire when you could simply use `request.task.state` instead.

I'm going to decline this PR for now. If you have a good justification for this and we decide to re-open the PR, then we'll need documentation and tests before we merge the PR in.

Thanks again! 🍻

PR #1984, opened 2017-02-25
Fix README example in the "POST Request with Parameters" section

pbartolome:
In the `README.md` the usage example for `POST Request with Parameters` doesn't include the parameter `method: .post`.

If the method is not included in the `request` it will default to `.get`


cnoon:
Good catch here @pbartolome...thanks for putting this together! 🍻

PR #1988, opened 2017-02-27
Added structure to example code block

ejdigby:
Added URL structure to Download File Destination example (lines 632 & 633) as required by Xcode


jshier:
Thanks for the PR! However, we won't be landing it, as the types are unnecessary. However, it did prompt me to fix another issue, which as fixed in #2058. Thanks! 🍻

PR #1994, opened 2017-03-01
Remove custom string extension and use standard library equivalent in its place.

Nirma:
The string extension defined in `Tests/String+AlamofireTests.swift` is redundant since the 
Swift Standard Library provides `String(repeating: count:)` which accomplishes the same thing.


jshier:
Thanks for the PR! A nice little change. Cheers! 🍻

PR #1997, opened 2017-03-01
Fixes bugs in examples

reitzig:
`URL.appendPathComponent` is mutating; the example is set up to use the generating version `URL.appendingPathComponent`.


jshier:
Thanks for the PR! While this could just be a change to `var`, preserving `let` semantics is better. Cheers! 🍻

PR #2021, opened 2017-03-17
Update README.md

sss989870:
Update README.md

SessionManager() isn't have 'adapter'
Change SessionManager() to Alamofire.SessionManager()


sss989870:
Sorry, I got it wrong.

PR #2035, opened 2017-03-23
Update custom encoding section of Readme.md

jorgifumi:
urlRequest property of URLRequestConvertible is optional, so you have to unwrap it or make non-optional.


jshier:
Thanks for the PR! Good catch! Cheers! :beers:

PR #2047, opened 2017-03-29
Added missing charset to JSONEncoding.encode()

2h4u:
The charset is needed to make sure, that the server uses the correct charset for decoding.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -267,7 +267,7 @@ public struct URLEncoding: ParameterEncoding {
 // MARK: -
 
 /// Uses `JSONSerialization` to create a JSON representation of the parameters object, which is set as the body of the
-/// request. The `Content-Type` HTTP header field of an encoded request is set to `application/json`.
+/// request. The `Content-Type` HTTP header field of an encoded request is set to `application/json; charset=utf-8`.
 public struct JSONEncoding: ParameterEncoding {
 
     // MARK: Properties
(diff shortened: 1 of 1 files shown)


vonox7:
+1, URLEncoding.default already uses `application/x-www-form-urlencoded; charset=utf-8`

jshier:
In my view this is redundant, as JSON is, by definition, already UTF-8. Using it with any other encoding is non-standard and not supported by Alamofire or the underlying `JSONSerialization` API it uses for JSON. @cnoon, what do you think?

cnoon:
Thanks for putting this together @2h4u! 

I agree that this should be added since JSON defaults to UTF-8, but can also be encoded in UTF-16 and UTF-32. There are some good details in this [thread](http://stackoverflow.com/a/9254967/1342462) and you can also check out [RFC 7158 - Section 8.1](https://tools.ietf.org/html/rfc7158#section-8.1) for more info.

I've made some additional changes to your PR @2h4u and pushed them into `master` in c8700ac7 while maintaining your attribution. This change will ship in the next release.

Thanks again! 🍻

jshier:
@cnoon From RFC 7158:
```
Note:  No "charset" parameter is defined for this registration.
           Adding one really has no effect on compliant recipients.
```
As mentioned in the SO thread you linked, servers interested in supporting UTF-16 or -32 can just check the first four bytes of the response to determine both the encoding and endianness (which can be either big or little), or any other standard algorithm for the same result.

So it still seems redundant.

cnoon:
Just as a heads up here, after much deliberation, @jshier and I have decide to revert this addition from the AF 4.5 release. I just pushed up the reverted commit into `master` in fb76d834. The main reasoning is that the `charset` parameter is not defined and there are already ways described in detail in the links posted on this thread on how to determine the encoding.

Thanks everyone for all their help on this particular issue. Cheers. 🍻

cnoon:
Also, please refer to the write-up in #2157 for more info.

PR #2070, opened 2017-04-12
Use curl -v instead of curl -i in cURLRepresentation()

webmaster128:
this prints header information to STDERR instead of STDOUT making it possible to process response body using pipes (e.g. decode gzip or apply JSON pretty print)

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -269,7 +269,7 @@ extension Request: CustomDebugStringConvertible {
     }
 
     func cURLRepresentation() -> String {
-        var components = ["$ curl -i"]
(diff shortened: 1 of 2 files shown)


webmaster128:
Fixed the tests for this update. Looks like the remaining fails are travis configuration issues

cnoon:
Hi @webmaster128, thank you for putting this change together! 🍻

After some debate, we've decided to accept this PR and use `curl -v` moving forward. If anyone finds it to be a bit much, then you can always switch back to `curl -i` when pasting back into terminal.

Thanks again!

PR #2072, opened 2017-04-13
enable setter of the MultipartFormData's contentType

ericdesa:
**Purpose**
The ```contentType``` of a ```MultipartFormData``` is ```multipart/form-data```
We could need to set it to another value (ex. ```multipart/related```), so i propose to set it writable

**Usage**
```
Alamofire.upload(multipartFormData: { mulitFormData -> Void in
        mulitFormData.contentType = "multipart/related; type=text/xml; boundary=\(mulitFormData.boundary)"
}, to: "https://example.com",
        method: .post,
        encodingCompletion: { encodingResult -> Void in
})
```

### DIFF
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -92,7 +92,7 @@ open class MultipartFormData {
     // MARK: - Properties
 
     /// The `Content-Type` header value containing the boundary used to generate the `multipart/form-data`.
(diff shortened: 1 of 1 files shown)


cnoon:
Hi @ericdesa,

Thanks for putting this change together...much appreciated! This will be part of the AF 4.5 release shortly.

Thanks again! 🍻

PR #2080, opened 2017-04-18
Delete: Warning

MMizogaki:
String interpolation produces a debug description for an optional value; did you mean to make this explicit?

Samplecode change String(describing: )


cnoon:
Hi @MMizogaki, thank you for taking the time to put this together...much appreciated! 🍻 I made a few modifications to your PR and pushed them up in a356cc24 while maintaining your attribution. These changes will go out as part of the AF 4.5 release.

Thanks again! 🍻

PR #2081, opened 2017-04-20
Fix ranges in readme

buscarini:
I fixed the 2 ranges that were missing one value (400..<499 and 500..<599) and I changed the other one to match the rest in order to simplify the document


jshier:
Thanks for the PR!

PR #2088, opened 2017-04-22
AFError conform to CustomNSError

Alexander-Ignition:
`AFError` converted in the `NSError` contains a `localizedDescription`, but the `userInfo` is empty.

```swift
let error = AFError.parameterEncodingFailed(reason: .missingURL)
let nsError = error as NSError

print(nsError.localizedDescription) // URL request to encode was missing a URL
print(nsError.userInfo) // [:]
```

If `AFError` to adapt to the `CustomNSError`, the `userInfo` will contain the `localizedDescription `.

```swift
let error = AFError.parameterEncodingFailed(reason: .missingURL)
let nsError = error as NSError

print(nsError.localizedDescription) // URL request to encode was missing a URL
print(nsError.userInfo) // [AnyHashable("NSLocalizedDescription"): URL request to encode was missing a URL]
```


jshier:
Thanks for the PR! However, we don't support Objective-C interop in this library, so why would we want this? Secondly, if we were to do this for one error type, it seems like we should do it for all of them.

jshier:
Without a strong case for why this needs to be done, I'm going to decline this PR. Thanks anyway!

Issue #2090, opened 2017-04-25
Add Additional Response Documentation for ResponseTests.swift


agnosticdev:
I was writing a few tests in `ResponseTests.swift` to validate where an issue may have been coming from while using AlamoFire and noticed that `ResponseTests.swift` has test cases grouped by the response data that is returned by no comment to indicate that.  Wanted to add some clarity around this.

cnoon:
Closing this issue out in favor of #2091.

PR #2091, opened 2017-04-25
Update ResponseTests.swift

agnosticdev:
Updated ResponseTests.swift with some enhanced comments for clarity.
This is in reference to the issue opened: https://github.com/Alamofire/Alamofire/issues/2090

### DIFF
--- a/Tests/ResponseTests.swift
+++ b/Tests/ResponseTests.swift
@@ -80,7 +80,7 @@ class ResponseTestCase: BaseTestCase {
     }
 }
 
-// MARK: -
+// MARK: - Testing Data Responses
 
 class ResponseDataTestCase: BaseTestCase {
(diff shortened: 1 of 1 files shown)


cnoon:
Hi @agnosticdev,

IMO, the jumpbar is pretty clear as to the separation by simply reading the names of the test cases. The MARKs are added to make it easier to read the jumpbar quickly. Is that not sufficient? We've used that approach in many places throughout the tests to split them up. The MARKs you've added here really just duplicate what's already in the name of the test case. I'd also argue that if we were to make such an addition, we'd need to make it throughout the test suite for consistency.

On a different note, we're going to be restructuring the entire test suite at some point in the near future. When we do that, we'll certainly take a pass at making the test case names as clear as possible and using MARKs where applicable.

Thanks anyways. 🍻

PR #2095, opened 2017-05-01
Fix super minor language error in README

mtrudel:
There are three previous examples, so the wording should be 'not use any' instead of 'not use either'.


cnoon:
Thanks @mtrudel! 🍻

PR #2102, opened 2017-05-11
Update Alamofire 4.0 Migration Guide.md

jack903652:
correct avaliable target


cnoon:
Thanks for putting this together @jack903652, but it should be `iOS 8.0+`. Alamofire 4.x does officially support iOS 8.0 now even though it didn't with the initial Alamofire 4.0.0 release.

Thanks anyways! 🍻

PR #2124, opened 2017-05-27
Add the execution result of example in README.md

maka-nai:
Swift & Objective-C beginners can not know the validity of execution results.
We are glad that there is no black box at least beginning of READ.ME.


jshier:
Thanks for the PR! However, we won't be merging it, as responses are unnecessary for our README and distract from the actual content.

PR #2128, opened 2017-05-30
Add timeout argument for request function

EJSohn:
I found out that many people use timeout like this:

```swift
let manager = Alamofire.SessionManager.default
manager.session.configuration.timeoutIntervalForRequest = 120
manager.request("yourUrl", method: .post, parameters: ["parameterKey": "value"])
```

Or rebuild configuration and set it to object’s session. 

I think it is not a good way to do it because if people work in teams these weird checks on exact "under-water" codes will make code very unreadable. Also not suit for Alamofire’s structure. 

I added this code because I thought it would be nice if Alamofire had this function as a basic feature.

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -126,6 +126,7 @@ extension URLRequest {
 /// - parameter parameters: The parameters. `nil` by default.
 /// - parameter encoding:   The parameter encoding. `URLEncoding.default` by default.
 /// - parameter headers:    The HTTP headers. `nil` by default.
+/// - parameter timeout:    The Request timeout. `nil` by default.
 ///
 /// - returns: The created `DataRequest`.
 @discardableResult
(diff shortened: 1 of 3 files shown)


cnoon:
Hi @EJSohn,

Thank you for putting this change together...much appreciated! 🍻

Unfortunately, we will not be accepting this change into Alamofire for a couple reasons. First off, it doesn't work. You can only modify the `URLSessionConfiguration` before it is applied to a `URLSession`. After it has been set, the mutation of it no longer applies.

Additionally, we want to keep the top-level APIs as lightweight as possible. We don't want to introduce any new parameters into those APIs unless absolutely necessary. Instead of adding a timeout property, you should either set it on the `URLSessionConfiguration` when you are setting it up, or you should set the value directly on the `URLRequest` and use the `Alamofire.request(urlRequest)` API.

Cheers. 🍻

PR #2131, opened 2017-06-01
avoid breaking up character sequences such as 👴🏻👮🏽

Coeur:
encoding fix on @cnoon code by using `rangeOfComposedCharacterSequences` like it was done on https://github.com/AFNetworking/AFNetworking/blob/master/AFNetworking/AFURLRequestSerialization.m#L67 by @kcharwood


Coeur:
Actually, after careful testing, @kcharwood change on AFNetworking was not needed and there is no need to apply it to AlamoFire.

PR #2132, opened 2017-06-01
New URLEncoding without brackets

ifero:
Even if there is no standard most of the web framework will avoid the usage of square brackets while passing an array as parameter.
That is why I've added a new parameter to avoid the usage of brackets during URL parameter encoding.


jshier:
Thanks for the PR! However, we're not going to merge it. As you said, there is no real standard for URL encoded collections, so adding levers to for every possible encoding is unfeasible. Creating your own encoding that does this is straightforward, and we'll look into enhancing `URLEncoding` in the future so that this sort of customization is easier.

PR #2152, opened 2017-06-14
201 HTTP status code added to set emptyDataStatusCodes

iosheadhunter:


jshier:
Thanks for the PR! However, we won't be merging it, as [RFC 7231, 6.3.2](https://tools.ietf.org/html/rfc7231#section-6.3.2) states that `Created` responses should contain the item(s) created, as simple best practice. We strive to support best practices where we can in Alamofire, so adding this code as an expected empty response isn't a good idea. Thanks though! :beers:

PR #2165, opened 2017-06-17
Update .swift-version

hyerra:
It's a really minor fix, but this branch supports Swift 4 so shouldn't we update this file to Swift 4?


jshier:
This is really only useful for CocoaPods, who are still deciding what to do with Xcode 9. We'll update this file if needed, if we need to publish a Swift 4 version to CocoaPods. Thanks anyway! 🍻

PR #2184, opened 2017-06-27
Remove unsafe URLCredential optional fields force unwrapping

achernoprudov:
Issue Link 🔗
No registered issue. 

Goals ⚽️
Alamofire crashes if log is enabled and request use client authentication through .pfx certificate. Log cant fetch user and password fields and crash by force unwrapping.

Implementation Details 🚧
Removed unsafe wrapping for user and password field.

Testing Details 🔍

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -293,11 +293,15 @@ extension Request: CustomDebugStringConvertible {
 
             if let credentials = credentialStorage.credentials(for: protectionSpace)?.values {
                 for credential in credentials {
-                    components.append("-u \(credential.user!):\(credential.password!)")
+                    let user = credential.user ?? ""
+                    let password = credential.password ?? ""
+                    components.append("-u \(user):\(password)")
                 }
             } else {
                 if let credential = delegate.credential {
-                    components.append("-u \(credential.user!):\(credential.password!)")
+                    let user = credential.user ?? ""
+                    let password = credential.password ?? ""
+                    components.append("-u \(user):\(password)")
                 }
             }
         }


jshier:
Thanks for the PR! However, before we merge, it would be a good idea to make this encoding a little more elegant, as empty username or password fields aren't usually the best behavior in some scenarios. Can you look into how cURL handles various types of auth and how `URLCredential` conveys them and see if you can come up with a more general solution?

cnoon:
I agree @jshier. I remember I looked into this a while back and it's much more complicated than this PR makes it. Certain types of auth support passwords and not usernames, etc. We need to make sure the cURL representation works in all the various cases.

As a temporary solution to the crash issue, I'd suggest we add a guard to ensure both the user and password exist on the credential, otherwise skip it. That is essentially all we support correctly today, so that change would just make it safe.

Source/Request.swift
```diff
@@ -293,11 +293,15 @@ extension Request: CustomDebugStringConvertible {
 
             if let credentials = credentialStorage.credentials(for: protectionSpace)?.values {
                 for credential in credentials {
-                    components.append("-u \(credential.user!):\(credential.password!)")
+                    let user = credential.user ?? ""
+                    let password = credential.password ?? ""
+                    components.append("-u \(user):\(password)")
```
cnoon: Could we change both of these to `guard` or `continue`?

```swift
guard let user = credential.user, let password = credential.password else { continue }
components.append("-u \(user):\(password)")
```

Then we'll circle back on this change in a future release to add better support for other types of auth.

cnoon:
Okay @achernoprudov, I ended up just making these changes myself and pushed them up into `master` in 0d2348da while maintaining your attribution. We'll still need to circle back to this and add better support for credentials other than basic auth, but this is certainly an improvement.

Thanks again! 🍻

PR #2186, opened 2017-06-28
refactored DispatchQueue extensions

adamnemecek:
There's no functionality difference, the change is cosmetic.


jshier:
Thanks for the PR! However, this is, in fact, functionally different, as it will return the same queue for each QoS level, not generate a new one like the original code. Since the intent here is to let the system manage what queues it gives us, this would fundamentally alter the behavior of this API. For that reason we won't be merging it. Cheers! :beers:

PR #2192, opened 2017-07-04
Add PathComponents parameter encoding

MrAsterisco:
### Goals :soccer:
Recent REST services exposes GET APIs with parameters passed as subcomponents of the Request URL. 

This means that instead of encoding parameters as

`requestUri?key=value&key2=value2`

their requests are something like

`requestUri/value/value2`

I am using Alamofire to interact with an ASP.NET WebApi 2 service that accepts parameters as described above. I think that Alamofire should be able to treat these kind of requests as normal parameters, even if they actually modify the request URL.

To do this, I have added a new struct that implements the ParameterEncoding interface and encodes parameters as subcomponents of the request URL.

### Testing Details :mag:
I have two tests which verify that parameters are inserted in the requested order.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -442,7 +442,7 @@ public struct PropertyListEncoding: ParameterEncoding {
 ///
 ///     baseUri/value1/value2
 ///
-struct PathComponents: ParameterEncoding {
+public struct PathComponents: ParameterEncoding {
     
     // MARK: Properties
     
@@ -459,7 +459,7 @@ struct PathComponents: ParameterEncoding {
     /// - throws: An `Error` if the encoding process encounters an error.
     ///
     /// - returns: The encoded request.
-    func encode(_ urlRequest: URLRequestConvertible, with parameters: Parameters?) throws -> URLRequest {
+    public func encode(_ urlRequest: URLRequestConvertible, with parameters: Parameters?) throws -> URLRequest {
         var urlRequest = try urlRequest.asURLRequest()
         
         guard let parameters = parameters else { return urlRequest }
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! Before we add another parameter encoding, can you point us to some documentation regarding this standard? To me it seems strange, but if it's already a standard I'd like to learn more.

MrAsterisco:
Hi @jshier ,

Actually, I don't know if this is a real standard yet, but it is widely used in REST services built with Microsoft ASP.NET WebAPI 2.

You can find further information on this page: [Routing in ASP.NET Web API](https://docs.microsoft.com/en-us/aspnet/web-api/overview/web-api-routing-and-actions/routing-in-aspnet-web-api). Also, this page is relevant, if you want to read more about this: [Attribute Routing in ASP.NET Web API 2](https://docs.microsoft.com/en-us/aspnet/web-api/overview/web-api-routing-and-actions/attribute-routing-in-web-api-2)

Just to take a look at other worlds too, I know a few PHP frameworks that work in the same way. For example, there's [FatFree Framework](https://fatfreeframework.com/3.6/routing-engine).

jshier:
Sorry for the long delay. In reading the ASP.NET docs, it seems like this isn't really a parameter encoding, but really a way of combining a few parameters with other path elements. Most of the examples given are something like `/api/authors/{authorId:int}/books`. In a case like that, I don't think a `ParameterEncoding` is a good fit, because, as you saw, many of those components aren't a parameter at all but something that's always part of the path. Really, what you'd want here is something that combines parameters like `["authorId": 1234]` with a path template like the one that's shown above. Alamofire doesn't really have an abstraction that fits that mold, but I think something that takes both parameters and a path template could be made to work with the existing Alamofire request protocols. So thanks anyway! Cheers! 🍻

Issue #1526, opened 2016-09-14
“Use Legacy Swift Language Version” with  pod 'Alamofire', '~> 4.0'


anilios:
I have to convert my project from XCode 7 to 8. But and i have following in my pod file:     pod 'Alamofire', '~> 4.0'

And i get this error on compilation.

“Use Legacy Swift Language Version” (SWIFT_VERSION) is required to be configured correctly for targets which use Swift. Use the [Edit > Convert > To Current Swift Syntax…] menu to choose a Swift version or use the Build Settings editor to configure the build setting directly.

How to fix this.

Jeyhey:
Have the same error

adrimarti:
Same here

ostatnicky:
Have you tried this? http://stackoverflow.com/a/38466703/1054550

Jeyhey:
Thanks! Worked for me. Hope there are no side-effects.

SimonTheEngineer:
I'm having the same issue and the above fix resulted in:

Module file was created by an older version of the compiler; rebuild 'Alamofire' and try again:
.../path/to/Alamofire.framework/Modules/Alamofire.swiftmodule/arm64.swiftmodule

Here's my Podfile if it helps..

```
platform :ios, '9.0'
use_frameworks!

target 'my-project' do

  pod 'Firebase'
  pod 'Firebase/Auth'
  pod 'Firebase/Messaging'
  pod 'ChameleonFramework/Swift', :git => 'https://github.com/ViccAlexander/Chameleon.git', :branch => 'swift3'
  pod 'SwiftyJSON', :git => 'https://github.com/acegreen/SwiftyJSON.git', :branch => 'swift3'
  pod 'Alamofire', '~> 4.0'

end

post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      config.build_settings['SWIFT_VERSION'] = '3.0'
    end
  end
end
```

jshier:
@ostatnicky's workaround should not be necessary if you're using CocoaPods 1.1.0, as it adds that setting to your `Pods` project automatically. As with any upgrade of Xcode and dependencies, clearing your DerivedData folder is always a good idea.

amirpervaiz086:
Hi,
[...]

PR #2197, opened 2017-07-10
Updated podspec targets to use SWIFT_VERSION 3

ooga:
### Issue Link :link:
#1526


jshier:
Thanks for the PR, but Alamofire is compatible with both Swift 3 and 4, so we don't need this setting. Also, in the future, please fill out the whole PR template. Thanks anyway!

PR #2240, opened 2017-08-13
Fixed `substring` API warnings for Swift 3.2 and up

htinlinn:
As of the latest version of Xcode 9 beta, `substring` functions from `String` have been deprecated and show up as warnings when compiling using Swift 4 compiler. 

This pull request fixes those warnings by replacing calls to those functions with the new subscript slicing syntax.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -308,7 +308,12 @@ extension Request: CustomDebugStringConvertible {
                 let cookies = cookieStorage.cookies(for: url), !cookies.isEmpty
             {
                 let string = cookies.reduce("") { $0 + "\($1.name)=\($1.value);" }
-                components.append("-b \"\(string.substring(to: string.characters.index(before: string.endIndex)))\"")
+
+                #if swift(>=3.2)
+                    components.append("-b \"\(string[..<string.index(before: string.endIndex)])\"")
+                #else
+                    components.append("-b \"\(string.substring(to: string.characters.index(before: string.endIndex)))\"")
+                #endif
             }
         }
 
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -308,11 +308,10 @@ extension Request: CustomDebugStringConvertible {
                 let cookies = cookieStorage.cookies(for: url), !cookies.isEmpty
             {
                 let string = cookies.reduce("") { $0 + "\($1.name)=\($1.value);" }
-
                 #if swift(>=3.2)
-                    components.append("-b \"\(string[..<string.index(before: string.endIndex)])\"")
+                components.append("-b \"\(string[..<string.index(before: string.endIndex)])\"")
                 #else
-                    components.append("-b \"\(string.substring(to: string.characters.index(before: string.endIndex)))\"")
+                components.append("-b \"\(string.substring(to: string.characters.index(before: string.endIndex)))\"")
                 #endif
             }
         }
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -223,9 +223,9 @@ public struct URLEncoding: ParameterEncoding {
                 let endIndex = string.index(index, offsetBy: batchSize, limitedBy: string.endIndex) ?? string.endIndex
                 let range = startIndex..<endIndex
 
-                let substring = string.substring(with: range)
+                let substring = string[range]
 
-                escaped += substring.addingPercentEncoding(withAllowedCharacters: allowedCharacterSet) ?? substring
+                escaped += substring.addingPercentEncoding(withAllowedCharacters: allowedCharacterSet) ?? String(substring)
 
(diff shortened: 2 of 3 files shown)


ivanaxel:
thanks

Source/Request.swift
```diff
@@ -308,7 +308,12 @@ extension Request: CustomDebugStringConvertible {
                 let cookies = cookieStorage.cookies(for: url), !cookies.isEmpty
             {
                 let string = cookies.reduce("") { $0 + "\($1.name)=\($1.value);" }
-                components.append("-b \"\(string.substring(to: string.characters.index(before: string.endIndex)))\"")
+
+                #if swift(>=3.2)
+                    components.append("-b \"\(string[..<string.index(before: string.endIndex)])\"")
```
jshier: I don't believe you need the `before:` call anymore, as `..<` does essentially the same thing.

Source/Request.swift
```diff
@@ -308,7 +308,12 @@ extension Request: CustomDebugStringConvertible {
                 let cookies = cookieStorage.cookies(for: url), !cookies.isEmpty
             {
                 let string = cookies.reduce("") { $0 + "\($1.name)=\($1.value);" }
-                components.append("-b \"\(string.substring(to: string.characters.index(before: string.endIndex)))\"")
+
```
jshier: No need for extra whitespace around the `#if` checks. We prefer to keep the whitespace the same as if there was no check in the first place.

Source/Request.swift
jshier: Please unindent the code inside the `#if` check to match out style.

htinlinn: Updated.

htinlinn: I think we will still want the `before:` call. `substring(to:)` gets you every character except for the one specified so in this case it seems to be stripping out the last character in the string.

jshier: `index(before: endIndex` gives you `endIndex - 1`. `..<endIndex` also gives you `endIndex - 1`. `..<index(before: endIndex)` would give you `endIndex - 2`. So to be equivalent to what it was before it would be `..<endIndex`, no?

Source/Request.swift
```diff
@@ -308,7 +308,12 @@ extension Request: CustomDebugStringConvertible {
                 let cookies = cookieStorage.cookies(for: url), !cookies.isEmpty
             {
                 let string = cookies.reduce("") { $0 + "\($1.name)=\($1.value);" }
+
+            #if swift(>=3.2)
```
jshier: `#if` statements and the code should be at the same level.

htinlinn: Fixed

Source/Request.swift
htinlinn: The original version with `substring(to:)` also gives you `endIndex - 2`: `-1` from `index(before:)` and another `-1` from `substring(to:)`.

jshier: Ah, thanks.

Source/Request.swift
```diff
@@ -308,7 +308,12 @@ extension Request: CustomDebugStringConvertible {
                 let cookies = cookieStorage.cookies(for: url), !cookies.isEmpty
             {
                 let string = cookies.reduce("") { $0 + "\($1.name)=\($1.value);" }
+
```
jshier: Sorry, like I said before, please remove the newlines before and after all version checks. Basically, where there are version checks, the whitespace should match what was there before. Same indentation level, same newlines.

cnoon:
Hi @htinlinn,

The compiler warnings you mention here do not appear for me in Xcode 9 beta 5. I'm currently installing beta 6 and will give it a shot there, but I don't anticipate we'll see the warnings there either. Were you seeing these issues in beta 4?

I'm going to close issue out for now since I'm not seeing the warnings. We'll certainly take a pass at updating all the APIs and implementations to the latest Swift 4 syntax in the AF 5 branch.

> If I'm completely mistaken here and the warnings are still present in beta 5 and for some reason I'm not seeing them, then please comment back on this ticket and we'll happily re-open.

Thanks anyways! 🍻

jshier:
@cnoon They only appear in Swift 4 mode.

@htinlinn One last change and this should be good to go.

htinlinn:
@jshier Done. Sorry about all the back and forth with this. I wasn't thinking about newlines when you said whitespace.

cnoon:
Ah thanks @jshier...late night. I figured I was missing something. 😕

jshier:
Merging without full Travis, because it would likely be hours before Travis got to us. All tests pass locally for me.

PGRBryant:
Looks like we've still got three substring deprecated and one 'characters' deprecated warnings hanging around. Unless the version I'm running from cocoapads isn't up to date?

PR #2242, opened 2017-08-15
Service Unavailable support

richardPFisk:
### Issue Link :link:

When an API is responding with Service Unavailable, due to situation like there is site maintenance required, clients can feedback the relevant time and situation to the user. 

In the section, [10.5.4 503 Service Unavailable](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.5.4), it is documented that "If known, the length of the delay MAY be indicated in a Retry-After header". 

See the documentation on the [`Retry-After` header](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.37)

"The Retry-After response-header field can be used with a 503 (Service Unavailable) response to indicate how long the service is expected to be unavailable to the requesting client."

### Goals :soccer:

This commit adds the ability to easily tell that an API is under maintenance and when it will be back online.

```
let afError = error as? AFError

if afError.isServiceUnavailableError {
    let retryAfter = afError.retryAfter
    //blah
}
```
This change adds the finer grained error of `isServiceUnavailableError` and then allows the date in the header to be used easily `afError.retryAfter`

### Testing Details :mag:

Integration tests with `Validation` and unit tests to cover both the happy paths and all variants of response `statusCode` and response `Retry-After` headers.

All other tests are passing after the small change too.

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -24,11 +24,11 @@
 		31ED52E81D73891B00199085 /* AFError+AlamofireTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31ED52E61D73889D00199085 /* AFError+AlamofireTests.swift */; };
 		31ED52E91D73891C00199085 /* AFError+AlamofireTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31ED52E61D73889D00199085 /* AFError+AlamofireTests.swift */; };
 		31ED52EA1D73891C00199085 /* AFError+AlamofireTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31ED52E61D73889D00199085 /* AFError+AlamofireTests.swift */; };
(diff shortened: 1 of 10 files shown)


richardPFisk:
I just wanted to say, all feedback is welcome. Thanks.

richardPFisk:
I have made a mistake. The `Retry-Header` is honoured by network infrastructure, given it is part of the HTTP protocol. Hence, the client will never receive this value. Instead, a request that receives this response is delayed until the `Retry-Header` time has elapsed *before* it reaches the client.

Thanks.

jshier:
Thanks for the PR! However it's unlikely this feature will be merged, for a few reasons. 

1. From my research, support for this version of the 503 response is very rare. Not only that, but even if it wasn't, supporting an arbitrary retry date, which could be hours or days in the future, is not a good fit for Alamofire. 
2. The changes you've made are breaking and invasive, so we wouldn't be able to ship them with Alamofire 4 anyway.
3. We'll be looking at how validation and retry work for Alamofire 5, so we may be able to provide a more flexible interface where this sort of feature is easier to implement. Perhaps a set of compassable retriers built into the library?

We'll review this and see how it could fit into the library.

cnoon:
I agree with @jshier here @richardPFisk. This implementation is best left in your logic on top of Alamofire for the time being. We'll certainly keep these changes in mind when working on AF 5.

Thanks anyways! 🍻

Issue #2245, opened 2017-08-17
No response headers when the response fails `ResponseValidationFailureReason.unacceptableStatusCode`


richardPFisk:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

Coded for client's requests failing.

## What did you expect to happen?

To be able to access the `statusCode` *and* response headers.

## What happened instead?

I can only access the `statusCode`.

## Alamofire Environment

**Alamofire version: 4.5.0**

Would you be willing to accept a well tested pull request which made the simple change adding the response headers to the `ResponseValidationFailureReason.unacceptableStatusCode` enum?

jshier:
We've discussed this feature internally. We created the `AFError` cases very specifically to capture only the content required for the descriptions of the error, not as general values. If you need this information, I suggest capturing the whole `*Response` value and using it to perform your logic. Otherwise a feature like would naturally lead to including every response value with the error, including the `Data` returned and the original `URLRequest`, which is obviously infeasible. If you have some more specific use cases in mind that can't be handled by the `*Response` values, please open a new issue with more specific examples and we'll investigate how they could be handled.

PR #2246, opened 2017-08-18
Added response headers to the Validation failure case of `ResponseValidationFailureReason.unacceptableStatusCode`

richardPFisk:
### Issue Link :link:

https://github.com/Alamofire/Alamofire/issues/2245

### Goals :soccer:

Adding the response headers to the `ResponseValidationFailureReason.unacceptableStatusCode` enum

The response headers help me to determine necessary information about the validation failure.

### Testing Details :mag:

A test has been created that the header is passed all the way through from `validate` in the test `testThatValidationForRequestWithUnacceptableStatusCodeResponseFailsWithResponseHeaders`

### DIFF
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -103,7 +103,7 @@ public enum AFError: Error {
         case dataFileReadFailed(at: URL)
         case missingContentType(acceptableContentTypes: [String])
         case unacceptableContentType(acceptableContentTypes: [String], responseContentType: String)
-        case unacceptableStatusCode(code: Int)
+        case unacceptableStatusCode(code: Int, headers: [AnyHashable : Any])
     }
 
     /// The underlying reason the response serialization error occurred.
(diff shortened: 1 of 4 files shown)


jshier:
Thanks for the PR! However, it is unlikely to be merged for a few reasons:

1. When implementing the new error model for Alamofire 4, we thought long and hard about what to include with the error cases themselves. We decided that including only the most relevant pieces of information would be best, otherwise we'd eventually just end up capturing the entire `Request` with the error, which is unnecessary. 
2. This fundamental change to the error enum would be a breaking change and would have to wait until Alamofire 5 to ship anyway. 
3. Alamofire's validation is designed to be extensible, so you could provide your own validation and error, which will be passed to the response parsers, so any custom payloads that are needed could be handled that way.
4. We'll be looking at how validation and and response parsing interact for Alamofire 5, so there may be significant changes in this area over the next few months.

We'll review how this sort of request can be integrated into the library.

cnoon:
I agree with @jshier here @richardPFisk. You are duplicating the Request itself in the error condition which is something we worked hard to avoid in the AF 4 implementation. You have access to the `Request` object in validation and response closures if you need it. That would be a more applicable way to achieve the same result.

Thanks anyways! 🍻

richardPFisk:
Thanks for the time to consider this PR. Using custom validation is a good solution.

Keep up the great work!

PR #2260, opened 2017-08-29
Fix SCNetworkReachabilitySetCallback block-style

hansemannn:
### Goals :soccer:
- Resolve an old block-style `{ (arg) in }` vs. { arg in }`

I was going through one of my projects and fixed all legacy-styles, so this one crossed my way 🙂.

### Testing Details :mag:
(No test required, just a syntactical change)

### DIFF
--- a/Source/NetworkReachabilityManager.swift
+++ b/Source/NetworkReachabilityManager.swift
@@ -1,7 +1,7 @@
 //
 //  NetworkReachabilityManager.swift
 //
-//  Copyright (c) 2014-2016 Alamofire Software Foundation (http://alamofire.org/)
+//  Copyright (c) 2014-2017 Alamofire Software Foundation (http://alamofire.org/)
 //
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR! However, there really isn't such a thing as old vs. new style parameter capture. In fact, dropping the parentheses has been possible for quite a while now. It's Alamofire project style to use the parentheses when capturing more than one value, so we won't be merging this PR. Thanks anyway! Cheers! 🍻

hansemannn:
Mhh. Thanks for the feedback! But in other sources like [here](https://github.com/Alamofire/Alamofire/blob/master/Source/SessionManager.swift#L61) it never uses parantheses, that's why my linting only complained at this particular point. But it's fine then.

jshier:
We'll look into adding a linter on our side for the future, to catch style inconsistencies like this.

Issue #2180, opened 2017-06-27
Decodable Response Parsing


jshier:
Alamofire should support both generic `Decodable` response parsing and specific `JSONDecoder` and `PropertyListDecoder` based response handlers.

mRs-:
Implementation for `Decodable` response request done with JSON is pretty easy.

My goto snippet for enabling this in `NSURLSession`
```swift
@discardableResult
func responseDecodable<T: Decodable>(request: Request, completion: @escaping (T?, URLResponse?, Error?) -> Void) -> URLSessionDataTask {
    return response(request: request, completion: { (data, response, error) in
        if let data = data {
            DispatchQueue.global(qos: .background).async {
                do {
                    let codable = try JSONDecoder().decode(T.self, from: data)
                    DispatchQueue.main.async {
                        completion(codable, response, error)
                    }
                } catch {
                    DispatchQueue.main.async {
                        completion(nil, response, error)
                    }
                }
            }
        } else if let error = error {
            completion(nil, response, error)
        } else {
            fatalError("Something went terrible wrong :(")
        }
    })
}

protocol NetworkServiceRequestable {
    @discardableResult
    func response(request: Request, completion: @escaping (Data?, URLResponse?, Error?) -> Void) -> URLSessionDataTask
}
```

jshier:
This has been merged into Alamofire 5. 

```swift
Alamofire.request(request).responseJSONDecodable { (response: DataResponse<HTTPBinResponse>) in
    print(response)
}
```

piyushnishanteu:
Hi,
I have installed Alamofire using Pod it installed Alamofire (4.7.2). There is no responseJSONDecodable method in Alamofire version 4.7.2.
It would be better if you suggest me installation instruction alamofire5 using pod .

sovanndyul:
There is no responseJSONDecodable in Alamofire, 4.7.3. any update about responseJSONDecodable  in Alamofire, 4.7.3 ?

jshier:
This feature won't be added to Alamofire 4, as not all users can support it on that version. That said, it's pretty trivial to add yourself, and you can take a look at how it's implemented in Alamofire 5 and almost directly copy it over.

krjw:
I couldn't find `responseJSONDecodable`in Alamofire 5 but this works for me:

```
        AF.request(Blub.url).validate().responseJSON { response in
            guard response.error == nil else {
                print(response.error!)
                return
            }
            guard let data = response.data else {
[...]

PR #2265, opened 2017-09-03
Decodable response serialization and major refactor

jshier:
### Issue Link :link:
#2180

### Goals :soccer:
The primary goal of the PR is to add a response serializer that uses `JSONDecoder` to generically decode responses. However, that lead to a major refactor of the response serialization protocols and their implementations.

Functionally, the serializers are the same, they've just been encapsulated into classes, rather than being generated closures. This allows they to be more easily used, and the awkward extensions on `Request` that generated the closures have been removed. Also, the serializations protocols have had their requirements changed from a serialization closure to a throwing function. This allows for far less awkward usage, as well as providing superior semantics with argument labels. 

### Testing Details :mag:
Tests have been added for the decodable serializers and response methods. Additionally, some of the tests have been refactored.


Tests/ResponseTests.swift
```diff
@@ -323,6 +323,47 @@ class ResponseJSONTestCase: BaseTestCase {
     }
 }
 
+class ResponseJSONDecodableTestCase: BaseTestCase {
```
jshier: I need more tests here.

Tests/ResponseTests.swift
```diff
@@ -323,6 +323,68 @@ class ResponseJSONTestCase: BaseTestCase {
...
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: timeout, handler: nil)
+
+        // Then
+        XCTAssertNotNil(response?.request)
+        XCTAssertNotNil(response?.response)
+        XCTAssertNotNil(response?.data)
+        XCTAssertNotNil(response?.data)
```
mRs-: you are testing two times the data, is this intended or a typo?

jshier: No, I must have copy pasted something.

jshier: Removed in [9e374bf](https://github.com/Alamofire/Alamofire/pull/2265/commits/9e374bfdb2cd81577d89338025ced09efc4608c5).

PR #2267, opened 2017-09-06
Alamofire 4.5.1 Release

cnoon:
This PR wraps up the AF 4.5.1 release. The majority of the changes are code and project cleanup and putting together the CHANGELOG. 

> The only change here that really warrants investigation is the `.swift-version` change. Seems to me that the main supported version here is Swift 3.2 so the file should reflect that.


.swift-version
```diff
@@ -1 +1 @@
-3.0
+3.2
```
jshier: Pretty sure this isn't necessary, as it's only used by CocoaPods to run the local lint build when doing a `pod trunk push`. Since Xcode only supports `3.0` or `4.0` (in Xcode 9) for this setting, only `3.0` should be necessary.

Alamofire.xcodeproj/project.pbxproj
```diff
@@ -1391,8 +1393,6 @@
 		4CF627001BA7CB3E0011A099 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				BITCODE_GENERATION_MODE = marker;
```
jshier: What are all of these removed settings?

cnoon: The `BITCODE_GENERATION_MODE` setting was needed originally for Carthage, but hasn't been needed for some time now (more details [here](https://github.com/Carthage/Carthage/issues/535)). They can simply be removed. 

As for the others, they are almost all duplicates of the project settings. If you go into the project settings for say the `Alamofire iOS` framework, then select `Customized` instead of `All`, you can see all the settings that "override" the project settings. And by project settings I mean the `Alamofire` project target above all the others.

> My apologies if you already know all this. I just want to make sure I provide enough context for the changes to make sense.

If you don't override the settings for each target, then it defaults to those set in the project. Anything that's bold is an "override" from the default. If something is bold, you can hit delete to remove the override and fallback to the default setting which comes from the project.

> Someone was nice enough to teach me this about a year ago. I'd use Xcode for probably 5+ years before I realized this. 😩

So...all the red in this PR in the project is actually just removing overrides in the framework and test targets that were the same as the project settings. Basically, they were overriding the default values with the same values. By removing the overrides, it makes it much easier to see what's been customized in each target. Generally, when working your way through the overrides, you'll find ones that are incorrect. Hence the cleanup.

jshier: Ah, sounds good.

Issue #2234, opened 2017-08-07
Reevaluate Lock Usage


jshier:
Alamofire currently uses a few different locking solutions, including `NSLock` and `DispatchQueue.sync`.  We should use a single pattern, if possible, for these needs. I like the [PThreadMutex / Protector](https://github.com/ProcedureKit/ProcedureKit/blob/81c6c6276aa198edf6a82b7dbabde2ae41e368c5/Sources/ProcedureKit/Support.swift#L67) combo from ProcedureKit but other options can be evaluated.

jshier:
This has been merged into Alamofire 5.

PR #2290, opened 2017-09-24
Refactor Lock Usage

jshier:
### Issue Link :link:
#2234

### Goals :soccer:
This PR refactors the usages of `NSLock` and other mechanisms to use the `Protector` type for values which need to be threadsafe.

This PR adds the `Protector`, `Mutex`, and `UnfairLock` types, as well as a `Lock` protocol, which abstracts the use of one locking mechanism or another. On newer OSes, `UnfairLock` is used, which is a wrapper around `os_unfair_lock`. On older systems `Mutex` is used, which is a wrapper around `pthread_mutex`. `Protector` uses these lock types to ensure thread-safe access to an internal value. `Protector` instances should be used rather than the `_property`, `property` with locks dance. It can also be extended to provide additional functionality in the future.

### Testing Details :mag:
No specific tests were added. Instead the test suite was run with the thread sanitizer enabled with no issues. Test may be added if we want coverage for something in particular.


Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+        return closure()
+    }
+
+    /// Execute a closure while aquiring the mutex.
+    ///
+    /// - Parameter closure: The closure to run.
+    func around(_ closure: () -> Void) {
+        lock(); defer { unlock() }
+        return closure()
+    }
```
cnoon: I think you might be able to implement both `around` functions in a protocol extension that leverages the `lock` and `unlock` APIs.

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+
```
cnoon: Extra newline.

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+        return closure()
+    }
+}
+
+/// An `os_unfair_lock` wrapper.
+@available (iOS 10.0, macOS 10.12, tvOS 10.0, watchOS 3.0, *)
+final class UnfairLock: Lock {
+    private var unfairLock = os_unfair_lock()
+
+    public init() { }
```
cnoon: `public` on an `internal` class?

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+    func around(_ closure: () -> Void) {
+        lock(); defer { unlock() }
+        return closure()
+    }
+}
+
+/// An `os_unfair_lock` wrapper.
+@available (iOS 10.0, macOS 10.12, tvOS 10.0, watchOS 3.0, *)
+final class UnfairLock: Lock {
+    private var unfairLock = os_unfair_lock()
```
cnoon: I need to read up on `os_unfair_lock` a bit.

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+
+    /// Execute a closure while aquiring the lock.
+    ///
+    /// - Parameter closure: The closure to run.
+    func around(_ closure: () -> Void) {
+        lock(); defer { unlock() }
+        return closure()
+    }
+}
+
```
cnoon: Could we get some `// MARK: -` separators between the types?

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+/// A thread-safe wrapper around a value.
+final class Protector<T> {
+    private let lock: Lock = {
+        if #available(iOS 10.0, macOS 10.12, tvOS 10.0, watchOS 3.0, *) {
+            return UnfairLock()
+        } else {
+            return Mutex()
+        }
+    }()
```
cnoon: Newline?

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+    private var ward: T
+
+    init(_ ward: T) {
+        self.ward = ward
+    }
```
cnoon: Would `value` make more sense than `ward`? Especially since you have an `unsafeValue` API to `ward`?

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+    ///
+    /// - Parameter closure: The closure to run.
+    func around(_ closure: () -> Void) {
+        lock(); defer { unlock() }
+        return closure()
+    }
+}
+
+/// A thread-safe wrapper around a value.
+final class Protector<T> {
```
cnoon: I've built similar abstractions before that were named `Locker` instead of `Protector`. Not really sure which one I prefer more. Thoughts?

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+            ward.append(contentsOf: newElements)
+        }
+    }
+
+    func append<C: Collection>(contentsOf newElements: C) where C.Iterator.Element == T.Iterator.Element {
+        write { (ward: inout T) in
+            ward.append(contentsOf: newElements)
+        }
+    }
+}
```
cnoon: These are slick!

Source/Mutex+Protector.swift
```diff
@@ -0,0 +1,176 @@
...
+
+    init(_ ward: T) {
+        self.ward = ward
+    }
+
+    /// The contained value. Unsafe for anything more than direct read or write.
+    var unsafeValue: T {
+        get { return lock.around { ward } }
+        set { lock.around { ward = newValue } }
+    }
```
cnoon: Should we really name this `unsafe` when it's actually safe for direct reads and writes? Seems a bit scary when used throughout the codebase. Makes me think it's unsafe since it's part of the API.

Source/TaskDelegate.swift
```diff
@@ -41,29 +41,23 @@ open class TaskDelegate: NSObject {
...
     }
     var initialResponseTime: CFAbsoluteTime?
     var credential: URLCredential?
     var metrics: AnyObject? // URLSessionTaskMetrics
 
-    private var _task: URLSessionTask? {
-        didSet { reset() }
-    }
-    private let taskLock = NSLock()
+    private let protectedTask: Protector<URLSessionTask?>
```
cnoon: Is the `protected` prefix just meant to make it more clear throughout the code that it's actually wrapped in a `Protector`? Is it overkill?

Source/Mutex+Protector.swift
jshier: Fixed in [6907baf](https://github.com/Alamofire/Alamofire/pull/2290/commits/6907bafde7b8e9648147cd045f0bc102d58d8a0f).

Source/Mutex+Protector.swift
jshier: Good call! Fixed in [6907baf](https://github.com/Alamofire/Alamofire/pull/2290/commits/6907bafde7b8e9648147cd045f0bc102d58d8a0f).

Source/Mutex+Protector.swift
jshier: Fixed in [6907baf](https://github.com/Alamofire/Alamofire/pull/2290/commits/6907bafde7b8e9648147cd045f0bc102d58d8a0f).

Source/Mutex+Protector.swift
jshier: Fixed in [6907baf](https://github.com/Alamofire/Alamofire/pull/2290/commits/6907bafde7b8e9648147cd045f0bc102d58d8a0f).

Source/Mutex+Protector.swift
jshier: Fixed in [6907baf](https://github.com/Alamofire/Alamofire/pull/2290/commits/6907bafde7b8e9648147cd045f0bc102d58d8a0f).

Source/Mutex+Protector.swift
jshier: I think it was just `ward` to match the protector theme, but I like value better. Fixed in [6907baf](https://github.com/Alamofire/Alamofire/pull/2290/commits/6907bafde7b8e9648147cd045f0bc102d58d8a0f).

Source/Mutex+Protector.swift
jshier: After a lot of thought, I've renamed the property to `directValue`. So while it isn't unsafe in the Swift sense of the word (i.e. could lead to undetected runtime memory corruption), I did want a name that made end users take notice. Fixed in [6907baf](https://github.com/Alamofire/Alamofire/pull/2290/commits/6907bafde7b8e9648147cd045f0bc102d58d8a0f).

PR #2315, opened 2017-10-09
Upgrade project to Swift 4.0

DavidPotter:
### Issue Link :link:
Upgrades project to Swift 4.0

### Goals :soccer:
Allow projects that depend on Alamofire to be converted without Xcode trying to convert Alamofire.

By providing a version of Alamofire that supports Swift 4, Xcode won't attempt to convert Alamofire when a project that depends on it is converted, which reduces confusion for the developer.

### Testing Details :mag:

Ran all tests on all platforms..

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -970,11 +970,11 @@
 					};
 					F8111E3219A95C8B0040E7D1 = {
 						CreatedOnToolsVersion = 6.0;
-						LastSwiftMigration = 0900;
+						LastSwiftMigration = "";
 					};
 					F8111E3D19A95C8B0040E7D1 = {
 						CreatedOnToolsVersion = 6.0;
-						LastSwiftMigration = 0900;
+						LastSwiftMigration = "";
 					};
 					F829C6B11A7A94F100A2CD59 = {
 						CreatedOnToolsVersion = 6.1.1;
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! However, we won't be merging it, as we'd like to keep our Swift default at 3 for Alamofire 4, as that maintains maximum out of the box compatibility with Xcode and Swift versions (3.0, 3.1, 3.2). We'll be moving to Swift 4 only for Alamofire 5. Thanks again!

DavidPotter:
Ah, okay.  Do you have a ETA for Alamofire 5?

jshier:
Not specifically, no. You can use Alamofire 4 with Swift 4 right now, or in Swift 3.2 mode with other Swift 4 modules.

DavidPotter:
Right.  What I'm hoping to accomplish is eliminate all warnings so that the only warnings we are presented with are "real" (this one isn't real) so the sooner you can release a version of Alamofire that eliminates this warning, the better.

Thanks,
**_David_**

jshier:
There are no warnings with the current version of Alamofire. How are you integrating it?

DavidPotter:
I use CocoaPods with this line:

`pod 'Alamofire', '~> 4.5.1'

It turns out that the warning was only produced while I was in the middle of converting the projects in my workspace.  Once they were converted, I exited Xcode, then I did a `pod install` again, the problem went away.  I guess you've got one more piece of information to provide people with when they say they get warnings 😀.

Sorry for not finding this out sooner.  Thanks for your help.

PR #2329, opened 2017-10-16
Swift dictionary convention

Dekablade01:
change invalid dictionary convention

remove spacing between `key` and and `:`


jshier:
Thanks for the PR! Cheers! 🍻

PR #2338, opened 2017-10-24
Add missing @discardableResult

rastersize:
### Goals :soccer:
- Align `request(_:)` with the other methods.

All the other request, download, upload and so on methods are annotated with `@discardableResult`. This changes makes sure that method can be used for fire-and-forget calls.

### DIFF
--- a/Source/SessionManager.swift
+++ b/Source/SessionManager.swift
@@ -249,6 +249,7 @@ open class SessionManager {
     /// - parameter urlRequest: The URL request.
     ///
     /// - returns: The created `DataRequest`.
+    @discardableResult
     open func request(_ urlRequest: URLRequestConvertible) -> DataRequest {
         var originalRequest: URLRequest?
 


jshier:
Thanks for the PR! Cheers! 🍻

jpdutoit:
Should this be reverted in the latest version? It looks like the request is being released automatically if no response handler is added...

jshier:
We probably should, but it would need a major version revision. We'll have to see how the next version's APIs shake out and reevaluate.

jpdutoit:
Ok. In our case this caused a bunch of problems when our fire-and-forget API calls stopped firing after updating Alamofire.

jshier:
To be clear, that behavioral change is intentional, as `Request`s no longer start until a response handler has been added or `resume()` has been called. That's true whether or not the instance is retained.

jpdutoit:
Thanks for the clarification. I must have missed it in the release notes

PR #2341, opened 2017-10-27
Add description for SPM API version 4

hayashikun:
### Goals :soccer:
Enable to build project including Alamofire by using Swift Package Manager API Version 4

Recently, Swift Package Manager API Version is updated.
So, I added how to write `Package.swift` in APIv4 format.

https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV3.md
https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ The [Swift Package Manager](https://swift.org/package-manager/) is a tool for au
 
 Once you have your Swift package set up, adding Alamofire as a dependency is as easy as adding it to the `dependencies` value of your `Package.swift`.
 
-#### PackageDescription API Version 3
+#### Swift 3
 
 ```swift
 dependencies: [
@@ -137,7 +137,7 @@ dependencies: [
 ]
 ```
 
-#### PackageDescription API Version 4
+#### Swift 4
 
 ```swift
 dependencies: [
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR! I think it would be less confusing to just mark the two sections as Swift 3.1 and Swift 3.2+ or perhaps Xcode 8 and Xcode 9. Once that change is made it can be merged.

hayashikun:
Actually, API version is not compatible with Swift version or Xcode version. 
So, marking `Swift tools version 3, 4` is correct, I think.

The default tools version is same with swift major version.
So there is a option to mark Swift 3 and Swift 4.

Which is the best?

jshier:
Version 3 shipped with the Swift 3 compiler and version 4 with the Swift 4 compiler, no? That should match exactly to Xcode versions, so that's probably the simplest way to indicate the distinction.

hayashikun:
https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV3.md#swiftlanguageversions

We can use APIv3 from Swift 4, but it is rare case. So I marked `Swift 3` and `Swift 4`

jshier:
How would you invoke the v3 behavior in the Swift 4 compiler? By having a `Package.swift` with a `swiftLanguageVersions` of `[3]`?

hayashikun:
In swift 4 compiler, default tool verion is 4 just after `swift package init`.
I can change swift tool version 3 by typing `swift package tools-version --set 3.0.0`

hayashikun:
or
Top line of `Package.swift` default is `// swift-tools-version:4.0`.
By changing this to `// swift-tools-version:3.0`, swift package tool version is switched.

jshier:
Thanks! Cheers! :beers:

Issue #2249, opened 2017-08-20
Extract and refactor ServerTrustPolicy


jshier:
In order for Alamofire to support Linux, or any platform where Apple doesn't provide a security framework, we need to extract `ServerTrustPolicy` behind a protocol and a parameter which defaults to nil. While this is happening we should refactor `ServerTrustPolicy` to no longer be `enum` based but a more open `struct` (or `class`), and add more evaluation types along with the ability to chain different evaluations.

PR #2344, opened 2017-10-30
Refactor ServerTrustPolicy

jshier:
### Issue Link :link:
#2249

### Goals :soccer:
This PR refactors the previous `ServerTrustPolicy` types into a protocol and several implementing classes, doing away with the previous `enum`. 

- `ServerTrustEvaluating` protocol added, which is intended to allow us to support Linux trust evaluation in the future, once that's possible. A method requirement is provided for Apple platforms, with the Linux requirement to be defined later. 

- The previous `ServerTrustPolicy` `enum` has be refactored into classes which implement `ServerTrustEvaluating`. This will allow us to more easily add support in the future (no breaking enum changes) for new evaluations or when adding Linux support. All logic remains the same.

- `CompositeTrustEvaluator` has been added, allowing users to use multiple `ServerTrustEvaluating` types at the same time.

- `ServerTrustPolicyManager` has been renamed to `ServerTrustManager`, with various relevant usage sites updated.

### Testing Details :mag:
All tests remain, but refactored for the new types. Tests for `CompositeTrustEvaluator` have been added.


jshier:
Merging this for now, so that AF5 development can move forward. We can reevaluate the API before release.

Issue #2230, opened 2017-08-07
Brotli Support


jshier:
In the new OSes, `URLSession` now supports Brotli content encoding. We should include the value in our default `Content-Encoding` header value.

PR #2346, opened 2017-10-30
Brotli Accept-Encoding Support

jshier:
### Issue Link :link:
#2230

### Goals :soccer:
This PR adds the Brotli encoding to the default `Accept-Encoding` header.

The `Accept-Encoding` header was made dynamic based on the available system. Unfortunately it can't be made entirely dynamic, as there's no way to query for supported encodings, but these changes match the header generated by `URLSession` if it isn't already there. It also removes `compress` support since it's not normally advertised and there wasn't a way to test it.

### Testing Details :mag:
Adds a `SessionManager` test against `httpbin.org` endpoints for supported encoding types to ensure we detect regressions.

### DIFF
--- a/Source/SessionManager.swift
+++ b/Source/SessionManager.swift
@@ -55,7 +55,19 @@ open class SessionManager {
     /// Creates default values for the "Accept-Encoding", "Accept-Language" and "User-Agent" headers.
     open static let defaultHTTPHeaders: HTTPHeaders = {
         // Accept-Encoding HTTP Header; see https://tools.ietf.org/html/rfc7230#section-4.2.3
-        let acceptEncoding: String = "br;q=1.0, gzip;q=0.9, compress;q=0.8"
+        let acceptEncoding: String = {
+            let encodings: [String]
+            if #available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, *) {
+                encodings = ["br", "gzip", "deflate"]
+            } else {
+                encodings = ["gzip", "deflate"]
+            }
+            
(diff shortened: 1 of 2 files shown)


SandrikaJanjghava:
Hello, I'm using this great and wonderful library - Alamofire. At first I want to thank you for this great library. I have a question about compression and decompression. I'm using latest beta version, my server supports brotli compression if I send to it Accept-Encdoin "br" in headers and I wanna know how to check if your project supports and receives compressed responce and than decompress, or if it does not do automatically how I must do it. Big thanks and good luck to you and your hole team.

jshier:
Alamofire's compression support is built on `URLSession`'s support, so we support everything it does on any given OS version. For something like Brotli this means that it's only support on 2017 OSes and later (iOS 11, etc.). `URLSession` is also responsible for the decompression itself, so as long as the response contains a `Content-Encoding` header that properly describes the compression used, and the OS supports that compression, then decompression should have automatically, before Alamofire ever sees the data.

SandrikaJanjghava:
big thanks for you quick and best answer
-- 
[image: logo]
*Sandro Janjghava*
CEO, Co-founder iDevelopers
571 99 26 00 <571992600>  | sandro.janjghava@idevelopers.eu |
Innovation St #7
<https://www.google.ge/maps/place/Tech+Park+Georgia/@41.6878006,44.7683255,17z/data=!4m5!3m4!1s0x0:0x62876d14c29bde36!8m2!3d41.6881125!4d44.7693005>
<http://facebook.com/sandruncha>

PR #2359, opened 2017-11-15
Enable injection of FileManager

fwigorcferreira:
### Goals :soccer:
Some enterprise MDM SDKs, like [BlackBerry Dynamics](https://us.blackberry.com/enterprise/blackberry-dynamics) have their own secure virtual file system, and the cache, and download of data needs to be stored using a custom subclass of FileManager, to interact with this database. In the current implementation of Alamofire, it relies on the `.default` singleton, being unable to inject, and so, replace this class.

Basically, all elements on Alamofire that uses a `FileManager` is receiving an instance on the initialiser, having a custom init with `.default` as the default parameter. Also, this reference is propagated to every sub component created.

### Testing Details :mag:
The hole project should behave equally, using `FileManager.default`, and interacting with the system the same way.

### DIFF
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -99,6 +99,9 @@ open class MultipartFormData {
 
     /// The boundary used to separate the body parts in the encoded form data.
     public let boundary: String
+    
+    /// The file manager handling file related operations.
+    public let fileManager: FileManager
 
     private var bodyParts: [BodyPart]
     private var bodyPartError: AFError?
@@ -109,7 +112,7 @@ open class MultipartFormData {
     /// Creates a multipart form data object.
     ///
     /// - returns: The multipart form data object.
-    public init() {
+    public init(fileManager: FileManager = .default) {
         self.boundary = BoundaryGenerator.randomBoundary()
         self.bodyParts = []
 
@@ -120,6 +123,8 @@ open class MultipartFormData {
         ///
 
         self.streamBufferSize = 1024
+        
+        self.fileManager = fileManager
     }
 
     // MARK: - Body Parts
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -262,7 +262,7 @@ open class MultipartFormData {
         var isDirectory: ObjCBool = false
         let path = fileURL.path
 
-        guard FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory) && !isDirectory.boolValue else {
+        guard self.fileManager.fileExists(atPath: path, isDirectory: &isDirectory) && !isDirectory.boolValue else {
             setBodyPartError(withReason: .bodyPartFileIsDirectory(at: fileURL))
             return
         }
@@ -274,7 +274,7 @@ open class MultipartFormData {
         let bodyContentLength: UInt64
 
         do {
-            guard let fileSize = try FileManager.default.attributesOfItem(atPath: path)[.size] as? NSNumber else {
+            guard let fileSize = try self.fileManager.attributesOfItem(atPath: path)[.size] as? NSNumber else {
                 setBodyPartError(withReason: .bodyPartFileSizeNotAvailable(at: fileURL))
                 return
             }
@@ -381,7 +381,7 @@ open class MultipartFormData {
             throw bodyPartError
         }
 
-        if FileManager.default.fileExists(atPath: fileURL.path) {
+        if self.fileManager.fileExists(atPath: fileURL.path) {
             throw AFError.multipartEncodingFailed(reason: .outputStreamFileAlreadyExists(at: fileURL))
         } else if !fileURL.isFileURL {
             throw AFError.multipartEncodingFailed(reason: .outputStreamURLInvalid(url: fileURL))
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -536,11 +536,12 @@ open class DownloadRequest: Request {
     /// - returns: A download file destination closure.
     open class func suggestedDownloadDestination(
         for directory: FileManager.SearchPathDirectory = .documentDirectory,
-        in domain: FileManager.SearchPathDomainMask = .userDomainMask)
+        in domain: FileManager.SearchPathDomainMask = .userDomainMask,
+        with fileManager: FileManager = .default)
         -> DownloadFileDestination
     {
         return { temporaryURL, response in
-            let directoryURLs = FileManager.default.urls(for: directory, in: domain)
+            let directoryURLs = fileManager.urls(for: directory, in: domain)
 
             if !directoryURLs.isEmpty {
                 return (directoryURLs[0].appendingPathComponent(response.suggestedFilename!), [])
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -103,6 +103,9 @@ open class Request {
 
     /// The session belonging to the underlying task.
     open let session: URLSession
+    
+    /// The file manager handling file related operations.
+    open let fileManager: FileManager
 
     /// The request sent or to be sent to the server.
     open var request: URLRequest? { return task?.originalRequest }
@@ -127,6 +130,7 @@ open class Request {
 
     init(session: URLSession, requestTask: RequestTask, error: Error? = nil, fileManager: FileManager) {
         self.session = session
+        self.fileManager = fileManager
 
         switch requestTask {
         case .data(let originalTask, let task):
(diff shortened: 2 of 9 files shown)


cnoon:
Thanks for taking the time to put this together @fwigorcferreira! Could you walk us through the specific reasons as to why you need to inject the file manager? This really REALLY muddies up the APIs, so we want to make sure this is absolutely necessary before considering the PR.

fwigorcferreira:
@cnoon My personal user case: I was creating an app using a MDM sdk that needs to keep the cache, and downloaded files securely encrypted on a virtual file system. When auditing the app, the client complained that cached data was not being properly used. So, I had to refactor Alamofire as stated on this PR to ensure that the data that goes thought network remains inside the virtual file system.

Given a more specific example, in the `Request` class:

```
if options.contains(.removePreviousFile), fileManager.fileExists(atPath: destinationURL.path) {
	try fileManager.removeItem(at: destinationURL)
}
```

This is a file manipulation that needs to happen in the virtual file system. A FileManager.default would try to look/manipulate files out of it.

cnoon:
Could you provide more details as to what your file manager does instead? Is it not possible to first save the file into your app sandbox, then move it to the virtual file system?

fwigorcferreira:
The MDM encrypts the data, splits it on 1, or more files, and create an encrypted mapper of this new virtual file path. The SDK provides a easy way to interact with this logic, by providing a subclass of FileManager to mask the paths, and handle file manipulations.

Given that, and that clients like banks have a strict police on file manipulation, I cannot save a temporary file on the sandbox, and, latter on, move it to an encrypted environment.

jshier:
Any reason you can't just use the built in file protection offered by the OS?

fwigorcferreira:
@jshier Other than the same strict policies that do not allow me to write on the sandbox file system, the MDM tool can control that virtual file system, like log the file installed on the application, securely share file between applications (when using the SDK sharing service), and remotely delete/modify files.

fwigorcferreira:
The possibility to inject the File Manager allows me to use the helper sub class together with the Alamofire library, without having to break read policies.

jshier:
For easier consideration, can you remove the modifications to the response serializers? Those are easy enough to write your own versions of with this capability. It'd be good to evaluate what injecting the `FileManager` looks like for just the core types.

fwigorcferreira:
@jshier I removed it from the response serialisers.

Source/MultipartFormData.swift
```diff
@@ -257,7 +262,7 @@ open class MultipartFormData {
         var isDirectory: ObjCBool = false
         let path = fileURL.path
 
-        guard FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory) && !isDirectory.boolValue else {
+        guard self.fileManager.fileExists(atPath: path, isDirectory: &isDirectory) && !isDirectory.boolValue else {
```
jshier: Unnecessary `self` reference.

Source/MultipartFormData.swift
```diff
@@ -269,7 +274,7 @@ open class MultipartFormData {
         let bodyContentLength: UInt64
 
         do {
-            guard let fileSize = try FileManager.default.attributesOfItem(atPath: path)[.size] as? NSNumber else {
+            guard let fileSize = try self.fileManager.attributesOfItem(atPath: path)[.size] as? NSNumber else {
```
jshier: Unnecessary `self` reference.

Source/MultipartFormData.swift
```diff
@@ -376,7 +381,7 @@ open class MultipartFormData {
             throw bodyPartError
         }
 
-        if FileManager.default.fileExists(atPath: fileURL.path) {
+        if self.fileManager.fileExists(atPath: fileURL.path) {
```
jshier: Unnecessary `self` reference.

Source/Request.swift
```diff
@@ -125,15 +128,16 @@ open class Request {
 
     // MARK: Lifecycle
 
-    init(session: URLSession, requestTask: RequestTask, error: Error? = nil) {
+    init(session: URLSession, requestTask: RequestTask, error: Error? = nil, fileManager: FileManager) {
```
jshier: This should default to `.default`.

Source/ResponseSerialization.swift
```diff
@@ -429,7 +429,10 @@ extension DownloadRequest {
     ///                       response, falling back to the default HTTP default character set, ISO-8859-1.
     ///
     /// - returns: A string response serializer.
-    public static func stringResponseSerializer(encoding: String.Encoding? = nil) -> DownloadResponseSerializer<String> {
+    public static func stringResponseSerializer(
```
jshier: Please don't rearrange lines you aren't changing as part of this PR.

Source/ResponseSerialization.swift
```diff
@@ -559,7 +562,7 @@ extension DownloadRequest {
             guard let fileURL = fileURL else {
                 return .failure(AFError.responseSerializationFailed(reason: .inputFileNil))
             }
-
+
```
jshier: Unnecessary whitespace changes. You can run the `Cleanup Whitespace` target to eliminate these.

Source/ResponseSerialization.swift
```diff
@@ -680,7 +683,7 @@ extension DownloadRequest {
...
```
jshier: Unnecessary whitespace changes.

Source/ServerTrustPolicy.swift
```diff
@@ -151,10 +151,10 @@ public enum ServerTrustPolicy {
     /// - parameter bundle: The bundle to search for all `*.cer` files.
     ///
     /// - returns: All public keys within the given bundle.
-    public static func publicKeys(in bundle: Bundle = Bundle.main) -> [SecKey] {
+    public static func publicKeys(in bundle: Bundle = Bundle.main, with fileManager: FileManager) -> [SecKey] {
```
jshier: Default to `.default`.

Source/SessionDelegate.swift
```diff
@@ -162,6 +162,7 @@ open class SessionDelegate: NSObject {
 
     var retrier: RequestRetrier?
     weak var sessionManager: SessionManager?
+    var fileManager: FileManager
```
jshier: This should be a `let`.

Source/SessionDelegate.swift
```diff
@@ -184,6 +185,7 @@ open class SessionDelegate: NSObject {
     ///
     /// - returns: The new `SessionDelegate` instance.
     public override init() {
+        self.fileManager = .default
```
jshier: Take it as a parameter defaulted to `.default`, otherwise you can't really inject it.

Source/SessionManager.swift
```diff
@@ -707,7 +718,7 @@ open class SessionManager {
                     // Cleanup the temp file once the upload is complete
                     upload.delegate.queue.addOperation {
                         do {
-                            try FileManager.default.removeItem(at: fileURL)
+                            try self.fileManager.removeItem(at: fileURL)
```
jshier: Unnecessary `self` reference.

Source/SessionManager.swift
```diff
@@ -727,7 +738,7 @@ open class SessionManager {
                 // Cleanup the temp file in the event that the multipart form data encoding failed
                 if let tempFileURL = tempFileURL {
                     do {
-                        try FileManager.default.removeItem(at: tempFileURL)
+                        try self.fileManager.removeItem(at: tempFileURL)
```
jshier: Unnecessary `self` reference.

Source/TaskDelegate.swift
```diff
@@ -322,8 +324,9 @@ class DownloadTaskDelegate: TaskDelegate, URLSessionDownloadDelegate {
 
     // MARK: Lifecycle
 
-    override init(task: URLSessionTask?) {
+    init(task: URLSessionTask?, fileManager: FileManager) {
```
jshier: Should be defaulted to `.default`.

jshier:
Looking at this more, this certainly seems like a change that, if we want it at all, should wait for Alamofire 5. We'll be significantly refactoring the internals of Alamofire and landing this change now would lead to additional work in that regard. Once we've refactored, then I think we can consider something like this. However, I'm not sure we'd want it, considering this is literally the first time I've ever seen `FileManager` subclassed. So the use case here is extremely rare. I also question whether we want to support `FileManager` subclasses directly vs. some sort of protocol abstraction. @cnoon Any thoughts here?

fwigorcferreira:
Given the amount of requests, and feedback about the functionality, I'll close this PR, and re-open, if necessary, when I have a better proposition.

Issue #2181, opened 2017-06-27
Encodable Request Creation


jshier:
Alamofire may be able to use `Encodable` types to help create requests. This will require some thought.

NickGaens:
Just being curious ... 

Do you have already figured out a way to approach this Encodable protocol? If yes, please do share any thoughts / ideas you have regarding this feature request so far.

arslan2012:
Here's what I think would work.

ParameterEnconding.swift

```swift
#if swift(>=4.0)
public func encode(_ urlRequest: URLRequestConvertible, with parameters: Encodable) throws -> URLRequest {
        var urlRequest = try urlRequest.asURLRequest()

        if urlRequest.value(forHTTPHeaderField: "Content-Type") == nil {
            urlRequest.setValue("application/x-www-form-urlencoded; charset=utf-8", forHTTPHeaderField: "Content-Type")
        }

        let encoder = JSONEncoder()
        do {
            let encoded = try encoder.encode(parameters)
            urlRequest.httpBody = encoded
        } catch {
            throw AFError.parameterEncodingFailed(reason: .jsonEncodingFailed(error: error))
        }
        return urlRequest
    }
#endif
```
I still don't know if encodable could be converted into querys. Hope this helps.

hshah2811:
@arslan2012  I did try this similar approach for encodable request but facing issue due to "static linking". 

> try encoder.encode(parameters)

here parameter is parameters: Encodable is not concrete type. encoder.encode(parameters) will throw below error at run-time.

> <Encodable> does not conform to Encodable because Encodable does not conform to itself. You must use a concrete type to encode or decode.

may be we need find a way to pass concrete type upto encoding function similar to like we have  for response decoding responsejsondecodable <T:Decodable>

jjatie:
Why can't we just make it a generic function of type `T` where `T: Encodable` i.e.:
```
#if swift(>=4.0)
public func encode<T: Encodable>(_ urlRequest: URLRequestConvertible, parameters: T) throws -> URLRequest {
        var urlRequest = try urlRequest.asURLRequest()

        if urlRequest.value(forHTTPHeaderField: "Content-Type") == nil {
            urlRequest.setValue("application/x-www-form-urlencoded; charset=utf-8", forHTTPHeaderField: "Content-Type")
        }

        let encoder = JSONEncoder()
        do {
            let encoded = try encoder.encode(parameters)
            urlRequest.httpBody = encoded
        } catch {
[...]

PR #2370, opened 2017-11-29
Requests taking Encodable parameters

jjatie:
### Issue Link :link:
#2181

### Goals :soccer:
Allow users to pass in their `Encodable` types instead of having to manually convert them to dictionaries.

- `JSONEncoding` and `PropertyListEncoding` are currently using the old serializations methods. Those implementations have been deprecated in favour of new implementations using `JSONEncoder` and `PropertyListEncoder` respectively.
- New generic Alamofire request method that accepts an `Encodable` type.
- `typealias Parameters = [String: Any]` is now `typealias Parameters = [String: Encodable]`. This should be okay as all JSON and XML types, as well as URL query parameters are `Encodable`. When possible (likely Swift 4.1) this should be redefined as
```
protocol ParameterValue: Encodable {}
extension Int: ParameterValue {}
extension String: ParameterValue {}
extension Array: ParameterValue where Element: ParameterValue {}
// etc..

typealias Parameters = [String: ParameterValue]
```

### DIFF
--- a/Source/Alamofire.swift
+++ b/Source/Alamofire.swift
@@ -159,10 +159,10 @@ public func request(
 ///
 /// - returns: The created `DataRequest`.
 @discardableResult
-public func request(
+public func request<T: Encodable>(
     _ url: URLConvertible,
     method: HTTPMethod = .get,
-    parameters: Encodable? = nil,
+    parameters: T? = nil,
     encoding: ParameterEncoding = JSONEncoding.default,
     headers: HTTPHeaders? = nil)
     -> DataRequest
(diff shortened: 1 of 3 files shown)


jshier:
Sorry it has taken so long to get to this, we've been investigating how we can integrate `Encodable`. Unfortunately this PR isn't the approach we need to take for a feature integrated so deeply into the library. It would be very confusing the users to provide an API that only works for some encoding types and not others. However, it could be a valuable extension on its own, so I encourage you to separate it into it's own library and publish it for others who want an `Encodable` request method. Thanks again! Cheers! :beers:

PR #2372, opened 2017-12-01
Fix 'characters' is deprecated

ykws:
use String or Substring directly count

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,11 @@
 language: objective-c
-osx_image: xcode9.1
 branches:
   only:
     - master
 env:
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! However, we need to keep Alamofire 4, including tests, compatible with Swift 3.1, 3.2, and 4. So if you could add the previous code back with Swift version checks we can get this merged.

ykws:
@jshier Thanks for comment! Try it with Swift 3.1

jshier:
Thanks for this @ykws. I've made these changes in #2441.

PR #2375, opened 2017-12-04
Alamofire 5!

jshier:
### Goals :soccer:
This PR tracks the mergeability of the `alamofire5` branch.


toohotz:
Hey there @jshier love the draft work that you have done so far for the next release of Alamofire (surprising 10 days and no other comments I guess I was a bit too excited to try out your changes).  Was just going over the new additions / changes and had a few questions to asks that may improve the clarity (at least for me) on how some of the new APIs should be used.

In one of my iOS projects I quickly swapped to using your branch as I use `Codable` in quite a few places since it has been introduced in Swift 4.
In regards to the new `JSONDecodableResponseSerializer` class, I assume the usage of this would look like something below:

Imgur1

Which is fine but I only came across this after I tried using `responseJSONDecodable` directly which didn't work of course because the generic type to be decoded could not be inferred so I was wondering if using it via this way, how would I specify what the type should be similarly to how I did ```let hashSerializer = JSONDecodableResponseSerializer<MD5Hash>()```?

Imgur

On another note, one issue I was hoping would get resolved (issue I feel sits within Apple's API itself) and it has to do with how `DecodingError` is handled.  So I noticed that if I don't have to do anything _special_ I can just call my completion handler with the object that I get back and let the caller handle it accordingly. 

Imgur

This worked fine with the exception of when an error was thrown (quickly tested this by changing the `md5` name to  `md6` so that the decode would fail on invalid `CodingKey` name) and noticed that when doing so, the failure case was not being called.  Decided to break up the method call to see what was taking place where inside the API call where the `.failure` was being called with a `DecodingError` as to be expected though I found it odd this was not propogated to the caller.  The other issue (and the one I came across before when first testing out the API) was that the `DecodingError` did not seem very _usable_ as I saw from the variables view what type of error was thrown being a `keyNotFound` which I expected to be thrown but trying to do a switch of anykind to sift through the different `DecodingError` types (`typeMismatch`, `valueNotFound`) gave me nothing so here I was stuck again with an error where at runtime I could see what the Variables View would show me but in terms of usability, I could not find how to parse the error itself.  

Imgur

Pretty longish 2nd comment but I hope it can help out as I love the changes especially the new locking that you have going on 😉.

Here’s the test code in its entirety for reference

```
class MD5Hash: Decodable {
    
    let md5: String
    
    let original: String
    
}

class AFJSONDecodableAPI {
    
    class func md5Hash(forString string: String, requestResult: @escaping (Result<MD5Hash>) -> () ) {
        let hashSerializer = JSONDecodableResponseSerializer<MD5Hash>()
        
        let urlString = "http://md5.jsontest.com"
        let params:[String :Any] = ["text": string]
        
        Alamofire.request(urlString, method: .get, parameters: params, encoding: URLEncoding.default, headers: nil).response(responseSerializer: hashSerializer) { (serializedObject) in
            switch serializedObject.result {
            case .success(let hash):
                print(hash)
            case .failure(let error):
                print(error)
            }
        }
    }
}
```

jshier:
@toohotz Previously, Alamofire's response methods were non-generic, since each serializer could only serialize one type (e.g. `String`, `Data`, `Any`). However, with `responseJSONDecodable`, you need to provide a fully qualified generic type for the serializer to know what to parse the JSON received into. So something like this:
```swift
Alamofire.request("https://httpbin.org/get).responseJSONDecodable { (response: DataResponse<HTTPBinResponse>) in
    print(response)
}
```
That way the serializer knows that you're expecting a specific type (in this case an `HTTPBinResponse`, which can you find in our test cases). So when the compiler complains about not being able to infer a parameter, that's likely what it's referring to.

toohotz:
Thanks for the quick reply, didn't realize that there is a preceding parameter name which has to be there, the autocomplete from Xcode didn't stub it in so thanks for that one, makes the call more succinct.

In regards to the errors I would have ideally like to handle the errors differently if a certain key were missing (mostly in the case for tests I would find it more useful).

Also noticed that you made the changes on your branch, is that to close out PR #2372 I assume?  Just curious.

jshier:
For error handling, like I said, cast it out to the right type and perform your logic. If you find yourself doing that a lot, make an extension on `Error` to do the cast for you.
```swift
extension Error {
    var asDecodingError: DecodingError? {
        return self as? DecodingError
    }
}
```
which makes it pretty easy to use:
```swift
switch response.error?.asDecodingError {
case .dataCorrupted?:
    // do something
...
}
```

toohotz:
That I had done once before but missed the fact that for the parameters on the error itself, those can be stubbed in which achieves what I actually wanted out of the error more descriptively, useful when decoding an array of `Decodable` objects from data coming from a backend for example and one happens to fail, the entire array if it contained anything prior to will be tossed so it's helpful to myself at least to know the cause of it.
Imgur

vfn:
Tests are failing because there is a reference to 1 simulator using `iOS 8.4`. As part of this PR you might want update the [.swift-version file](https://github.com/Alamofire/Alamofire/blob/master/.swift-version) as well to swift 4.0

Source/Session.swift
```diff
@@ -0,0 +1,497 @@
...
+            case let .resumeData(resumeData):
+                self.rootQueue.async { self.didReceiveResumeData(resumeData, for: request) }
+            }
+        }
+    }
+
+    func performSetupOperations(for request: Request, convertible: URLRequestConvertible) {
+        do {
+            let initialRequest = try convertible.asURLRequest()
+            rootQueue.async { request.didCreateURLRequest(initialRequest) }
```
RahulKatariya: As the implementation of this method uses async, the `Session.default.request(urlRequest).request` object doesn't have the same properties as of `urlRequest` when accessed immediately.

I have a broken test at https://github.com/Restofire/Restofire/blob/2abd788ea9e5454f5f5bf90ce1ae386028b2c4b3/Tests/AlamofireRequestable/ARequestableSpec.swift#L41

jshier: Yes. If you need to test such properties, you can either make a request and check the properties when it's complete, or use an `EventMonitor` to tap into the request creation events.

RahulKatariya: Thanks for quick response.

Just thinking, doesn't it make sense at the time of creation of `Request` object, it should give the same `urlRequest` object back immediately but if there is a side-effect then it should listen using the `EventMonitor`.?

Right now the following code will return false
```swift
let urlRequest = URLRequest(...)
let header = HTTPHeaders.authorization(username: "user", password: "password")
header.forEach { urlRequest.setValue($0.value, forHTTPHeaderField: $0.key) }
print(Session.default.request(urlRequest).request.allHTTPHeaderFields == urlRequest.allHTTPHeaderFields)
```

jshier: Is it really unexpected when you realize the request creating process is now asynchronous? I suppose we could update the inline docs to indicate that it won't be available until request creation process is complete, and even then, that property is actually just the last of all `URLRequest`s received by the `Request`, so there's more complexity there than before, where intermediate `URLRequests` were lost. Additionally, testing that the Alamofire property has some value doesn't actually guarantee that the underlying `URLSession` used all of those properties to make a request. Certain headers, especially authorization headers, are specifically documented as not necessarily being used from the `URLRequest` but instead can be changed by the session. In any event, if you really need this test, I think waiting for the event is the right way to go, as it largely replicate the previous behavior.

RahulKatariya: I realize the request creating process is now asynchronous. The urlRequest is something which is already created and I am passing to Request and expecting it to be the same.

The following code returns true
```swift
print(try! Session.default.request(urlRequest).convertible.asURLRequest().allHTTPHeaderFields == urlRequest.allHTTPHeaderFields)
```

RahulKatariya: FYI - Immediately calling `print(request.debugDescription)` also broke as it uses the last request property too.

jshier: Yes, we know. It should print "$ curl command could not be created" until everything is available.

RahulKatariya: 👍🏻

Source/ResponseSerialization.swift
```diff
@@ -463,251 +423,149 @@ extension DownloadRequest {
...
-        options: PropertyListSerialization.ReadOptions = [],
-        completionHandler: @escaping (DownloadResponse<Any>) -> Void)
-        -> Self
-    {
-        return response(
-            queue: queue,
-            responseSerializer: DownloadRequest.propertyListResponseSerializer(options: options),
-            completionHandler: completionHandler
-        )
+    public func responseJSONDecodable<T: Decodable>(queue: DispatchQueue? = nil,
```
philtre: Perhaps it would be useful to add a `type` parameter to this function in order to avoid having to specify `DataResponse<...>` in each completion closure.

```
public func responseJSONDecodable<T: Decodable>(type: T.Type = T.self, queue: DispatchQueue? = nil, decoder: JSONDecoder = JSONDecoder(), completionHandler: @escaping (DataResponse<T>) -> Void) -> Self
```

Calling the `response` method would be somewhat simpler:

```
request.responseJSONDecodable(type: HTTPBinResponse.self) { response in
    ...
}
```

jshier: That doesn't seem to gain you much, if anything, as it's more verbose to pass it as a parameter. Passing it could also prevent proper type inference if you were to pass a closure to this method.  Adding the type to closure parameter is standard Swift usage for any generic closure like this.

philtre: Good point. 
This is probably more useful as a convenience for synchronous methods where 'T' is the return type, eg:
`func doSomething<T>(type: T.Type = T.self) -> T`
and then either inferring the return value
`let val = doSomething(type: Int.self)`
or the function parameter parameter
`let val: Int = doSomething()`

I guess one could add the proposed convenience in an extension to `DataRequest`

PR #2389, opened 2017-12-18
Clarify OS versions affected by download resume bug

getaaron:
The bug referenced in `Usage.md` was fixed in iOS 10.2, etc. This PR updates the documentation which still erroneously refers to these as the "latest" release.


jshier:
Thanks for the PR, this is great info! One minor nit, please add spaces around the hyphens. Once that's done we can merge.

getaaron:
Hi @jshier, to be extra nit picky, those are not hyphens, but en dashes, which typically are not surrounded by spaces when expressing a numeric range. See, for example, http://www.thepunctuationguide.com/en-dash.html

PR #2393, opened 2017-12-21
Set appropriate protection levels on NetworkReachabilityManager members

parski:
### Goals :soccer:
I want to be able to mock `NetworkReachabilityManager` for unit testing in integrating projects. The current protection level of `public` does not allow inheritance forcing me to inconvenient testing methods with wrappers and the like. By changing the protection level to `open` the class and the appropriate members can be subclassed and overridden making mocking it is a breeze.

I've changed protection levels from `public` to `open` where applicable.

### Testing Details :mag:
I used a test project to reproduce the problem with inheritance and subsequent overriding. I made my changes and ran the test suite to make sure everything passes. I was thinking of creating a mock of `NetworkReachabilityManager` and adding it to the test target of the example app but I'm not sure it's a viable means of testing since compilation would be the assertion and the mock would be unused. Let me know if you think any kind of test is appropriate for this PR.


jshier:
Looks good, thanks! This should go out with the 4.7.0 release soon.

PR #2412, opened 2018-01-09
Removes the old request to avoid endless growth.

mrdepth:
### Implementation Details
Remove the old request from delegate.protectedRequests before assigning new one.

### Testing Details :mag:
	
```
class Retrier: RequestRetrier {
	func should(_ manager: SessionManager, retry request: Request, with error: Error, completion: @escaping RequestRetryCompletion) {
		completion(request.retryCount < 5, 0)
	}
}

var session: SessionManager? = SessionManager()
session?.retrier = Retrier()
session?.request("http://google.com")
	.validate(statusCode: 400...500) //simulate error
	.responseString { (response) in
		//session.delegate.protectedRequests contains 5 records
		session = nil
}

```


jshier:
Thanks for the PR! This looks straightforward, but I'd like @cnoon to review as well. 

Also, we need some tests around this to ensure the requests don't grow on retry.

oadk:
Hi,

Do you have a rough timescale for when this PR might be merged? We have unfortunately been hit by this problem, and it is causing unbounded memory growth when there are lots of retried requests (e.g. on poor quality mobile connections). We can patch locally, but it would be great to see the fix in an official release.

(Thank you for making Alamofire - it's awesome!).

cnoon:
@jshier this is absolutely something we need to get resolved ASAP. I'd recommend we move this through on this branch, then cherry pick the change back to `master` along with some tests. 

There's quite a few ways we could test this, but it will require a retrier to keep retrying (maybe up to say 2 or 3 times), and then stop retrying. Then we just need to run a count on the session delegate's number of requests.

> If you have time to take that on awesome! Otherwise I can take a pass tomorrow.

oadk:
Thanks everyone for the patch and 4.7.2 release. This is very helpful.

PR #2416, opened 2018-01-21
ResultTests: Small updates and typos fix

alaphao:
### Goals :soccer:
Update ResultTests by:
* Fixing typos in unit tests failure messages
* Removing String.character.count warnings
* Using XCTAssertNil and XCTAssertNotNil for nil assertion


jshier:
We can probably get this into 4.7.0, as long as it doesn't conflict with my changes.

PR #2427, opened 2018-02-03
extend userInfo-dict of .DidComplete-Notification to contain responseData of DataTasks

svendr:
### Goals :soccer:
Alamofire currently doesn't allow generic logging of DataTask-Responses via Notifications. This would be especially useful if you use a wrapper around Alamofire like SwaggerClient and can't access or configure the logging directly.

The .DidComplete-Notification already provides access to the URLSessionDataTask via the userInfo-dictionary, but there's no way to get the raw responseData associated with such a DataTask.

AFNetworking implements a similar notification mechanism, but instead also provides access to the responseData via userInfo-dictionary.

I've simply extended `open func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?)` to check if the task's delegate is a DataTaskDelegate and if so, put it's data into the userInfo-dictionary of the .DidComplete-notification.

### Testing Details :mag:
I've extended SessionDelegateTestCase with 2 proper test-methods.

### DIFF
--- a/Source/Notifications.swift
+++ b/Source/Notifications.swift
@@ -48,5 +48,8 @@ extension Notification {
     public struct Key {
         /// User info dictionary key representing the `URLSessionTask` associated with the notification.
         public static let Task = "org.alamofire.notification.key.task"
+
+        /// User info dictionary key representing the responseData associated with the notification.
+        public static let ResponseData = "org.alamofire.notification.key.responseData"
     }
 }
--- a/Source/SessionDelegate.swift
+++ b/Source/SessionDelegate.swift
@@ -442,10 +442,16 @@ extension SessionDelegate: URLSessionTaskDelegate {
 
             strongSelf[task]?.delegate.urlSession(session, task: task, didCompleteWithError: error)
 
+            var userInfo: [String: Any] = [Notification.Key.Task: task]
+
+            if let data = (strongSelf[task]?.delegate as? DataTaskDelegate)?.data {
+                userInfo[Notification.Key.ResponseData] = data
+            }
+
             NotificationCenter.default.post(
                 name: Notification.Name.Task.DidComplete,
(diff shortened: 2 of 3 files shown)


jshier:
Thanks for the PR! I'm hesitant to include such a piecemeal change when Alamofire's notifications need to be overhauled anyway and that may be part of Alamofire 5. Introduction and use of the feature now means more breakage in the future. But we'll discuss.

svendr:
Hey jshier, thanks for your feedback. May I ask about the roadmap of Alamofire5? How will the notification mechanism be overhauled?

Did you consider a 4.6.1 or 4.7.0 release, where this small change could be included? It would really help at several projects I’m working on.

I’ll be happy to support you re-implementing and / or testing notifications in AF5 if required ;)

jshier:
I'm inclined to add it for now, since AFNetworking seems to have survived with it. My worry is that we could see memory usage increases. `Data` is copy-on-write I believe, so there may not be an issue. @svendr, do you mind investigating how the notification affects the memory usage of apps with large responses? 

@cnoon, what do you think? Was this considered when adding these notifications initially?

driemecker:
Hey @jshier, I've just used Xcode's Memory Gauge to compare memory consumption of the master vs my ResponseBodyInNotification-branch (with registered observer for .DidComplete-Notif) by using the Example App and downloading a 180MB JSON file (https://raw.githubusercontent.com/zemirco/sf-city-lots-json/master/citylots.json). As expected, there has been no difference in memory usage.

jshier:
@driemecker Thanks for the test! This will be part of the 4.7 release. Thanks again!

svendr:
@jshier thanks for merging this! And 4.7.0 is already released 👍  Much appreciated!

PR #2431, opened 2018-02-08
Add configurable parameter handling to URLEncoding.

heiberg:
### Goals :soccer:

There is no real specification for how to URL encode collection parameters like arrays and
dictionaries.

For arrays, two conventions are in common use:

1. `foo=[1,2]` is encoded as `foo[]=1&foo[]=2`
2. `foo=[1,2]` is encoded as `foo=1&foo=2`

Alamofire follows the first convention, as does libraries such as jQuery and Rails.

The second convention is not currently supported by Alamofire, but is commonly required for communicating with Java-based web services.

This PR introduces optional support for the convention where square brackets are omitted 
from array keys.

- A new `ArrayEncoding` enumeration is added to `URLEncoding`. The cases are `.brackets` and `.noBrackets`.
- An `arrayEncoding` parameter is added to `URLEncoding.init` with a default value of `.brackets` to preserve the existing default behavior and maintain source compatibility.
- The `arrayEncoding` passed to `init` is used to control the array key serialization in `URLEncoding.queryComponents`.

### Testing Details :mag:

The two existing tests for URL encoding of arrays are:

- `testURLParameterEncodeStringKeyArrayValueParameter` (top level array)
- `testURLParameterEncodeStringKeyNestedDictionaryArrayValueParameter` (array nested inside a dictionary)

These have been duplicated and adapted to test an instance of `URLEncoding(arrayEncoding: .noBrackets)` instead of the default `URLEncoding`.

The two new tests are named like the existing tests, with an appended suffix:

- `testURLParameterEncodeStringKeyArrayValueParameterWithoutBrackets`
- `testURLParameterEncodeStringKeyNestedDictionaryArrayValueParameterWithoutBrackets`

All tests pass.

---

### Update

Amended the pull request with configurable boolean parameter encoding, as per @jshier 's comment.

- The bool value configuration follows the pattern of the array key configuration closely for consistency. 
- Source compatibility is maintained and the default bool encoding behavior is preserved.

Added the following test case: 

`testURLParameterLiteralBoolEncodingWorksAndDoesNotAffectNumbers`

The test verifies that 

- Boolean values can be encoded to literals as expected.
- Enabling literal bool encoding has no effect on the encoding of truthy or falsy numeric parameters like `NSNumber(value: 1)` etc.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -64,9 +64,12 @@ public protocol ParameterEncoding {
 /// the HTTP body depends on the destination of the encoding.
 ///
 /// The `Content-Type` HTTP header field of an encoded request with HTTP body is set to
-/// `application/x-www-form-urlencoded; charset=utf-8`. Since there is no published specification for how to encode
-/// collection types, the convention of appending `[]` to the key for array values (`foo[]=1&foo[]=2`), and appending
-/// the key surrounded by square brackets for nested dictionary values (`foo[bar]=baz`).
+/// `application/x-www-form-urlencoded; charset=utf-8`.
+///
+/// There is no published specification for how to encode collection types. By default the convention of appending
+/// `[]` to the key for array values (`foo[]=1&foo[]=2`), and appending the key surrounded by square brackets for
+/// nested dictionary values (`foo[bar]=baz`) is used. Optionally, `ArrayEncoding` can be used to omit the
+/// square brackets appended to array keys.
 public struct URLEncoding: ParameterEncoding {
 
     // MARK: Helper Types
@@ -80,6 +83,15 @@ public struct URLEncoding: ParameterEncoding {
     /// - httpBody:        Sets encoded query string result as the HTTP body of the URL request.
     public enum Destination {
         case methodDependent, queryString, httpBody
+    }
+
+    /// Configures how array parameters are encoded.
+    ///
+    /// - brackets:        An empty set of square brackets is appended to the key for every value.
+    ///                    This is the default behavior.
+    /// - noBrackets:      No brackets are appended. The key is encoded as is.
+    public enum ArrayEncoding {
+        case brackets, noBrackets
     }
 
     // MARK: Properties
@@ -99,15 +111,20 @@ public struct URLEncoding: ParameterEncoding {
     /// The destination defining where the encoded query string is to be applied to the URL request.
     public let destination: Destination
 
+    /// The convention for encoding array parameters.
+    public let arrayEncoding: ArrayEncoding
+
     // MARK: Initialization
 
     /// Creates a `URLEncoding` instance using the specified destination.
     ///
     /// - parameter destination: The destination defining where the encoded query string is to be applied.
+    /// - parameter arrayEncoding: The convention to use for encoding array parameters.
     ///
     /// - returns: The new `URLEncoding` instance.
-    public init(destination: Destination = .methodDependent) {
+    public init(destination: Destination = .methodDependent, arrayEncoding: ArrayEncoding = .brackets) {
         self.destination = destination
+        self.arrayEncoding = arrayEncoding
     }
 
     // MARK: Encoding
@@ -160,8 +177,9 @@ public struct URLEncoding: ParameterEncoding {
                 components += queryComponents(fromKey: "\(key)[\(nestedKey)]", value: value)
             }
         } else if let array = value as? [Any] {
+            let arrayKey = (arrayEncoding == .brackets) ? "\(key)[]" : key
             for value in array {
-                components += queryComponents(fromKey: "\(key)[]", value: value)
+                components += queryComponents(fromKey: arrayKey, value: value)
             }
         } else if let value = value as? NSNumber {
             if value.isBool {
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -88,21 +88,39 @@ public struct URLEncoding: ParameterEncoding {
         case methodDependent, queryString, httpBody
     }
 
-    /// Configures how array parameters are encoded.
+    /// Configures how `Array` parameters are encoded.
     ///
     /// - brackets:        An empty set of square brackets is appended to the key for every value.
     ///                    This is the default behavior.
     /// - noBrackets:      No brackets are appended. The key is encoded as is.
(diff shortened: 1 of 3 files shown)


Vyazovoy:
👍 really useful PR!

jshier:
We’ve wanted this for a while, so thanks! In addition to Array encoding, a Bool encoding has also been requested in the past, 0, 1 vs. literal true/false, so if you want to add that we’d appreciate it. I’m on vacation right now so it may be a little while before a full review.

heiberg:
Sure! I'll update this with configurable Bool encoding shortly. No rush, have a nice vacation.

Source/ParameterEncoding.swift
```diff
@@ -82,6 +88,23 @@ public struct URLEncoding: ParameterEncoding {
         case methodDependent, queryString, httpBody
     }
 
+    /// Configures how array parameters are encoded.
```
jshier: I think having `Array` instead of array makes it a bit more clear.

Source/ParameterEncoding.swift
```diff
@@ -82,6 +88,23 @@ public struct URLEncoding: ParameterEncoding {
...
+    ///
+    /// - brackets:        An empty set of square brackets is appended to the key for every value.
+    ///                    This is the default behavior.
+    /// - noBrackets:      No brackets are appended. The key is encoded as is.
+    public enum ArrayEncoding {
+        case brackets, noBrackets
+    }
+
+    /// Configures how boolean parameters are encoded.
```
jshier: `Bool` here, like `Array`.

Source/ParameterEncoding.swift
```diff
@@ -82,6 +88,23 @@ public struct URLEncoding: ParameterEncoding {
...
+    ///
+    /// - numeric:         Encode `true` as 1 and `false` as 0. This is the default behavior.
```
jshier: I think ticks around `1` and `0` are appropriate here.

Source/ParameterEncoding.swift
```diff
@@ -99,15 +122,24 @@ public struct URLEncoding: ParameterEncoding {
     /// The destination defining where the encoded query string is to be applied to the URL request.
     public let destination: Destination
 
+    /// The convention for encoding array parameters.
```
jshier: It's a bit clearer to just say "The encoding to use for `Array` parameters."

Source/ParameterEncoding.swift
```diff
@@ -99,15 +122,24 @@ public struct URLEncoding: ParameterEncoding {
...
+    public let arrayEncoding: ArrayEncoding
+
+    /// The convention for encoding boolean parameters.
+    public let boolEncoding: BoolEncoding
```
jshier: As with `arrayEncoding`, it's a bit clearer to say "The encoding to use for `Bool` parameters."

Source/ParameterEncoding.swift
```diff
@@ -99,15 +122,24 @@ public struct URLEncoding: ParameterEncoding {
...
+
     // MARK: Initialization
 
     /// Creates a `URLEncoding` instance using the specified destination.
     ///
     /// - parameter destination: The destination defining where the encoded query string is to be applied.
+    /// - parameter arrayEncoding: The convention to use for encoding array parameters.
     ///
```
jshier: Need parameter docs for `boolEncoding` as well.

Source/ParameterEncoding.swift
```diff
@@ -261,6 +293,24 @@ public struct URLEncoding: ParameterEncoding {
             return false
         }
     }
+
+    private func encodeArrayKey(_ key: String) -> String {
```
jshier: It would be cleaner if this was simply an `encode` function on `ArrayEncoding`. Better encapsulation that way.

Source/ParameterEncoding.swift
```diff
@@ -261,6 +293,24 @@ public struct URLEncoding: ParameterEncoding {
...
+        switch arrayEncoding {
+        case .brackets:
+            return "\(key)[]"
+        case .noBrackets:
+            return key
+        }
+    }
+
+    private func encodeBoolValue(_ value: Bool) -> String {
```
jshier: Like the array encoding function, this should be an `encode` function on `BoolEncoding`.

Source/ParameterEncoding.swift
```diff
@@ -161,16 +193,16 @@ public struct URLEncoding: ParameterEncoding {
             }
         } else if let array = value as? [Any] {
             for value in array {
-                components += queryComponents(fromKey: "\(key)[]", value: value)
+                components += queryComponents(fromKey: encodeArrayKey(key), value: value)
```
jshier: Moving the `encode` functionality to `ArrayEncoding` would make this `arrayEncoding.encode(key)`, which I think is much clearer.

Source/ParameterEncoding.swift
```diff
@@ -161,16 +193,16 @@ public struct URLEncoding: ParameterEncoding {
...
             }
         } else if let value = value as? NSNumber {
             if value.isBool {
-                components.append((escape(key), escape((value.boolValue ? "1" : "0"))))
+                components.append((escape(key), escape(encodeBoolValue(value.boolValue))))
```
jshier: With `encode` on `BoolEncoding`, this becomes `boolEncoding.encode(value.boolValue)`.

Source/ParameterEncoding.swift
```diff
@@ -161,16 +193,16 @@ public struct URLEncoding: ParameterEncoding {
...
             } else {
                 components.append((escape(key), escape("\(value)")))
             }
         } else if let bool = value as? Bool {
-            components.append((escape(key), escape((bool ? "1" : "0"))))
+            components.append((escape(key), escape(encodeBoolValue(bool))))
```
jshier: With `encode` on `BoolEncoding`, this becomes `boolEncoding.encode(bool)`.

heiberg:
Excellent suggestions! I added commits to address them.

The only deviation from your suggestions is that I wanted to keep the external parameter name for the `encode` methods. So the usage becomes `arrayEncoding.encode(key: ...)` and `boolEncoding.encode(value: ...)` at the call sites.

One is intended for keys and the other is intended for values, and I think keeping the external parameter name help to differentiate and reduce the risk of mistakes when using them.

Issue #900, opened 2015-10-31
How can I do a SOAP request?


dasoga:
Hi everyone! (this is a question)
I want to do a SOAP request for example with this web service:
http://www.holidaywebservice.com//HolidayService_v2/HolidayService2.asmx?wsdl

I need to send one parameter "countryCode". I don't know how can i do that with alamofire. And how i get the response to parse the xml result.

Thanks for your help.

jshier:
Alamofire doesn't have any built in SOAP or XML support, so you'd have to write your own. Fortunately it's straightforward to [create a custom response serializer] (https://github.com/Alamofire/Alamofire/blob/master/README.md#response-serialization-1), but the XML parsing is up to you.

As for the parameters, take a look at the [documentation] (https://github.com/Alamofire/Alamofire/blob/master/README.md#parameters) and see which of the built in encoding match your use case.

dasoga:
Thanks for your response!

Issue #873, opened 2015-10-20
soap services in swift using alamofire?


aleemilu75:
Does Alamofire support Soap web service? will it create soap envelope itself or we need to provide it ?

cnoon:
It does not directly. You will need to create your own SOAP interface around Alamofire.

PR #2433, opened 2018-02-10
Added Module for managing soap request

ShakeebM:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/900
https://github.com/Alamofire/Alamofire/issues/873

### Goals :soccer:

Managing Soap Request with Alamofire is a difficult one. here I'm trying to solve this. I have added a support for Soap Web Service Request and response can receive as String format. later we can convert into XML 

- I have added a class `SoapRequestManager` which create soap request header and body when we passed inputs like 

> Web service URL
> Namespace (By default it will be http://tempUri.org/)
> Soap Parameters (It should pass as the normal `Parameters` class in swift
> Soap Method 

- Then I have added `soapRequest()` method in the `Alamofire` class. It can be called just like any other `Alamofire` `DataRequest`
### Testing Details :mag:
I have tested using the example swift project already given in the repository. i added a new table view cell for Soap Request. i used a sample web service that got from internet.


jshier:
@ShakeebM Thanks a lot for the work you've put in here. However, we aren't going to merge this PR, as we don't believe SOAP functionality is a good addition to the core of Alamofire. It's both not recommended in general and has a lot of edge cases in particular, so it would be a rather significant maintenance burden moving forward. However, I encourage you to turn this functionality into it's own open source library which integrates with Alamofire and which users can use for their SOAP needs. Thanks again! Cheers! 🍻

ShakeebM:
@jshier  thank you for the advice. I'am going to move on that way. i believe some guys are still in trouble with SOAP service. how ever i will be still here to continue contributions with Alamofire.
have a nice day!

ShakeebM:
Here is my extension with Alamofire Soap
https://github.com/ShakeebM/AlamofireSoap

PR #2436, opened 2018-02-14
fixed: cookies not shown in cURLRepresentation

kiruto:
When disabled cookies in session configuration and set cookies as a header, show cookies in curl query.

### Goals :soccer:
fixed: cookies not shown in cURLRepresentation

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -319,6 +319,7 @@ extension Request: CustomDebugStringConvertible {
         }
 
         var headers: [AnyHashable: Any] = [:]
+        var cookies: String?
 
         if let additionalHeaders = session.configuration.httpAdditionalHeaders {
             for (field, value) in additionalHeaders where field != AnyHashable("Cookie") {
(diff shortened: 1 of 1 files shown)


Source/Request.swift
```diff
@@ -327,15 +328,23 @@ extension Request: CustomDebugStringConvertible {
...
+                    cookies = value
+                }
             }
         }
 
         for (field, value) in headers {
             components.append("-H \"\(field): \(value)\"")
         }
 
+        if let c = cookies, !session.configuration.httpShouldSetCookies {
```
jshier: I think this can simply be added as an else to the cookie logic that ends on line 319.

```swift
if session.configuration.httpShouldSetCookies {
    if
        let cookieStorage = session.configuration.httpCookieStorage,
        let cookies = cookieStorage.cookies(for: url), !cookies.isEmpty
    {
        let string = cookies.reduce("") { $0 + "\($1.name)=\($1.value);" }

    #if swift(>=3.2)
        components.append("-b \"\(string[..<string.index(before: string.endIndex)])\"")
    #else
        components.append("-b \"\(string.substring(to: string.characters.index(before: string.endIndex)))\"")
    #endif
    }
} else if
    let sessionCookies = session.configuration.httpAdditionalHeaders?["Cookie"] as? String,
    let requestCookies = request.allHTTPHeaderFields?["Cookie"] {
    components.append("-b \"\(sessionCookies);\(requestCookies)\"")
}
```

With that, none of the other changes are needed. 

Also, is it possible for a user to use both the standard cookie storage and attach their own cookies, or would one overwrite the other? Perhaps this code should access the storage cookies, if available, as well as the session and request cookies, regardless of the `session.configuration.httpShouldSetCookies` setting?

jshier:
Closing due to age. If anyone else would like to take a look at this issue, feel free to open a new PR.

PR #2446, opened 2018-02-21
Remove pledgie.com

lukabratos:
Reason: http://pledgie.com/

Alternative? https://www.patreon.com/


jshier:
Thanks for the letting us know! We'll look for a new donation site, but I'm merging this now so people aren't misled.

PR #2468, opened 2018-03-14
Add swift_version to spec

hcanzonetta:
### Goals :soccer:
* Specify the swift version that will be used for the pod regardless of the swift version the app target uses.

This feature was added on CocoaPods 1.4.0 (http://blog.cocoapods.org/CocoaPods-1.4.0/)


jshier:
Unfortunately, since this setting only takes a single version, and Alamofire supports multiple versions, we won't be adding it. It harms anyone who already has a modern version of Swift set in their main project (it will be inherited by CocoaPods), as it's best to have the same version used for all dependencies, if possible. Once CocoaPods adds the multiple version implementation of this feature, we'll gladly add our supported versions.

PR #2488, opened 2018-04-13
Makes rethrows all methods that can be marked as rethrows

stephanecopin:
### Goals :soccer:
Allows to use `try` in non-escaping closures.
For example, the following code, before the change:
```
let response: DataResponse<Data> = <...>
let result: Result<Any>
switch response.result {
case .failure(let error):
	result = .failure(error)
case .success(let value):
	result = .success(try JSONSerialization.jsonObject(with: value))
}
let newResponse = DataResponse(
	request: response.request,
	response: response.response,
	data: response.data,
	result: result,
	timeline: response.timeline
)
```
Would be able to be changed to:
```
let response: DataResponse<Data> = <...>
let newResponse = try response.map { try JSONSerialization.jsonObject(with: value) }
```

(this is a simplified example for demonstrating what the change brings - for this specific example, it might make more sense to use `flatMap`, or just use `responseJSON` if possible)

Just some syntax additions, adding `throws` to the closure and `rethrows` to the method that accepts the closure.

### Testing Details :mag:
None, as this is purely a change for programmer's. The tests still passes.

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -158,12 +158,12 @@ extension DataResponse {
     ///
     /// - returns: A `DataResponse` whose result wraps the value returned by the given closure. If this instance's
     ///            result is a failure, returns a response wrapping the same failure.
-    public func map<T>(_ transform: (Value) -> T) -> DataResponse<T> {
+    public func map<T>(_ transform: (Value) throws -> T) rethrows -> DataResponse<T> {
         var response = DataResponse<T>(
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! I've removed the changes to the map functions, as they were specifically made non-throwing by design. However, throwing versions of the `with` and `if` functions would be nice.

Issue #2300, opened 2017-09-29
Normalize Response Types


jshier:
While separate response types may be useful for separating API, the `Default*` versions of these responses are quite awkward. We should reevaluate these APIs and try to normalize down to a simple inheritance tree.

jshier:
It seems like we could replace `DefaultDataResponse` and `DefaultDownloadResponse` with `DataResponse<Data?>` and `DownloadResponse<Data?>`.

jshier:
This was done as part of the rewrite.

Issue #2235, opened 2017-08-07
Audit Queue Usage


jshier:
In addition to locking queues, there may be other unnecessary queues used. We should explore how we use queues in Alamofire and see if we can eliminate some unnecessary usage. One example might be sharing a single underlying `DispatchQueue` between the `URLSessionDelegate` queue, the task queue, and the `Request` `OperationQueue`. If this can maintain performance it might be a good way to limit the resource impact of using Alamofire.

jshier:
Done in the rewrite.

Issue #2225, opened 2017-08-07
Consolidate URLSessionTaskMetrics into Timeline


jshier:
Our current integration of `URLSessionTaskMetrics` is awkward and inelegant, requiring an `AnyObject` property with internal shenanigans to expose the metrics to users. It's also a bit confusing to have that property in addition to the `Timeline` property exposed for older OSes. We should combine all statistics under `Timeline`, which would add additional properties to cover what's received from the `URLSessionTaskMetrics`.

jshier:
Like enhanced TLS errors, supporting this would require a significant refactor of Alamofire's internals in regards to what data goes where.

jshier:
Timeline has been removed in favor of full `URLSessionTaskMetrics` support in the rewrite.

PR #2521, opened 2018-05-21
Alamofire 5 Core Rewrite!

### LINKED ISSUE #2226: Asynchronous Request Creation
Perhaps the most inelegant bit of Alamofire's current API is the handling of multipart form uploads. Due to the asynchronous nature of the request creation, it doesn't support chaining like every other `Request` API, requiring users to get the actual upload request in the completion for the request creation. Alamofire should move to an asynchronous `Request` creation model for all request types. Not only could this greatly simplify multipart uploads, but would allow expensive operations in other requests as well, an issue we've had reported before. Most likely this will entail something like a `Future` API for `Request`, or at least an internal state machine so that method chaining can still work.

jshier:
### Goals :soccer:
This PR is a complete rewrite of core Alamofire types and behaviors. Fixes #2226, #2235, #2225, #2300, among others.

### Needed :white_check_mark:
- [x] Finalize which `EventMonitor` types will be provided, reimplement notifications as an `EventMonitor`, add `EventMonitor` tests.
- [x] Reimplement `description` and `debugDescription`
- [x] Reimplement `Response` mapping methods.
- [ ] Update all inline documentation.

tl;dr: `SessionManager`, `SessionDelegate`, and the `Request` types have been completely rewritten and most other areas of the library touched in some way. Oh, and iOS 10 is now the minimum (unless the `URLSessionTaskMetrics` linker bug on watchOS forces us to backtrack on metrics support).

#### Summary
This PR rewrites `SessionManager`, `SessionDelegate`, and the `Request` types for many reasons, including:

- Library evolution had lead to functionality and areas of responsibility being spread amongst many types within Alamofire. This made it increasingly difficult to add features, diagnose and fix bugs, and optimize performance. Most notably, the original subdelegate architecture has been completely removed, as it was a huge source of complexity and one of the biggest things preventing a Linux port. The `TaskDelegate` functionality has been reimplemented in the `SessionDelegate` or `Request`s themselves.
- Queue usage was distinctly suboptimal, with separate queues for the `URLSessionDelegate` and each `Request`, as well as global queue usage for things like multipart form requests and retry, easily leading to scenarios where users were able to negatively impact system performance without realizing it and without being able to do anything about it.
- Swift has evolved substantially over the lifetime of Alamofire and Alamofire hasn't kept pace. This is especially true in the area of error handling, where we using own custom constructs where standard `try/catch` would suffice. This rewrite updates those areas to better follow Swift convention and the advancements made over the years.
- Alamofire's usage of the `URLSessionDelegate` methods was preventing implementation of a few features and was burdened by trying to cover all facets of `URLSession`. This rewrite both prunes features that weren't tested and weren't likely being used (streaming) and reimplements security challenge handling by changing which delegate methods are implemented.

#### `SessionManager`
`SessionManager` has been reimplemented in order to simplify it's two areas of responsibility:

- Initializing dependent parts of Alamofire.
- Creating `Request`s.

...

### Testing Details :mag:
Existing tests have been updated but are largely untouched.


Source/Alamofire.swift
```diff
@@ -116,303 +116,309 @@ extension URLRequest {
...
-    usingThreshold encodingMemoryThreshold: UInt64 = MultipartUpload.multipartFormDataEncodingMemoryThreshold,
-    with urlRequest: URLRequestConvertible) -> UploadRequest
-{
-    return SessionManager.default.upload(
-        multipartFormData: multipartFormData,
-        usingThreshold: encodingMemoryThreshold,
-        with: urlRequest
-    )
+/// Global namespace containing API for the `default` `SessionManager` instance.
+public enum Alamofire {
```
GoCac: Hi, @jshier ,  there is a question - why do you choice `enum` rather than `struct` here？I wonder to know the difference between them. Thank you.

jshier: By convention, since Swift doesn’t have namespaces, enums are used instead, as they can’t be accidentally instantiated.

GoCac: got it. Thanks for your reply~👍

freak4pc:
Hey! I'm aware this is an old PR, but any whether or not this is documented somehow, migration-wise? I'm trying to figure out some of the latest changes to make sure everything is still ticking for us.

PR #2523, opened 2018-05-21
Docummentation added example

saulurias:
### Goals :soccer:
Provide developers an example of how to obtain the http status code from the response.

Took me a while to find the http code so I decided that could be good to have it on the Response Handling docummentation example.

### Testing Details :mag:
Not needed


jshier:
I'm sorry, I meant you should branch off of `master`, make your change, and then make a PR targeting `master`, not any of the `alamofire5` branches.

PR #2525, opened 2018-05-22
http status code docummentation example

saulurias:
### Goals :soccer:
Provide developers an example of how to obtain the http status code from the response.

Took me a while to find the http code so I decided that could be good to have it on the Response Handling docummentation example

### Testing Details :mag:
Not needed


getaaron:
Don't we want to encourage people to use validators, e.g. `.validate(statusCode: 200..<300)`?

jshier:
Thanks for the PR! However, this documentation is a high level overview of handling responses with `.response` methods, not an exhaustive listing of every property available. Therefore, this documentation would be better elsewhere.

PR #2537, opened 2018-06-03
Updated SPM installation instructions.

WayneEld:
The Swift Package Manager instructions were swapped around with regards to the Swift 4 vs Swift 3 dependency syntax. (Swift 4 begins with the ".Package", not Swift 3).  Implemented the installation in server side swift with the correct Swift 4 syntax and it works.


WayneEld:
Apologies I was wrong.

PR #2552, opened 2018-07-07
Update for Xcode 10 and Swift 4.2

jshier:
### Goals :soccer:
This PR fixes warnings and a test failure from Xcode 10 / Swift 4.2.


jshier:
Merged despite failure, as httpbin.org is being flaky, but tests pass locally.

PR #2563, opened 2018-07-31
[WIP] AF5 Feature - Redirects Per Request

cnoon:
This PR is a work in-progress, but currently adds the ability to AF5 to control redirect behavior per-request instead of only at the `SessionDelegate` level.

### Problem

The general issue with redirect handling is that you want to allow most endpoints to redirect, and deny several very specific ones. We recently ran into this issue and had to go to some extreme lengths to work around the lack of per-request redirect control in Alamofire. Commonly, the place where you are implementing the `SessionDelegate` or subclass of it will not know about all the URLs that are going to be piped through it, nor would you want it to. However, the `SessionDelegate` only has a single callback point for all requests hitting a redirect, so it would be awesome if Alamofire could automatically provide a way to control redirects per-request rather than having to do this through the `SessionDelegate`.

### Solution

Per-request redirect support! By adding a `Request.redirect` API that allows clients to specify redirect behavior for a specific request, it greatly simplifies the redirect logic. This gives clients full control over all redirects at the `Request` level, making it VERY easy to disable redirects for a single call without having to subclass the `SessionDelegate`.

### Callouts

@jshier I haven't added tests or anything for this behavior yet b/c I wanted to run the idea by you first to make sure this is going in the right direction. Feedback welcome!

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -80,6 +80,8 @@ open class Request {
         var uploadProgressHandler: (handler: ProgressHandler, queue: DispatchQueue)?
         /// `ProgressHandler` and `DispatchQueue` provided for download progress callbacks.
         var downloadProgressHandler: (handler: ProgressHandler, queue: DispatchQueue)?
+        /// `RedirectHandler` provided to handle per-request directs.
+        var redirectHandler: RedirectHandler?
         /// `URLCredential` used for authentication challenges.
         var credential: URLCredential?
         /// All `URLRequest`s created by Alamofire on behalf of the `Request`.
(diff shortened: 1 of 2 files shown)


jshier:
Given that this is an API that the redirect removed due to the removal of the closure API, I think it would be better to offer the same API on `SessionManager` and `Request` at the same time. A closure or protocol encapsulation should work, but users should be able to build a single redirect handler for all requests, or one per request using the same API, similar to how the retriers and adapters will eventually work out.

cnoon:
Definitely, that's a great callout. I'll take another pass with that consideration in mind.

mahadshahib:
i just wanna thank you so much i installed your version of alamofire and my resumedata problems fixed really reallyy thank you man love u

cnoon:
Closing this PR out since it was replaced by #2699 and merged into `master`.

PR #2569, opened 2018-08-05
Podspec explicitly specify documentation url

arr00:
Cocoadocs does not work anymore. To avoid a 404 when a user clicks the documentation link on cocoapods, you need to specify a url.


jshier:
Thanks! 🍻

PR #2585, opened 2018-09-07
Refactor the rewrite!

jshier:
### Goals :soccer:
This PR refactors core types following discussion of the rewrite merged in #2521. It also updates the project to require Swift 4.2 and updates our Travis environment to match.

This refactor pulls the `RequestTaskMap` that keeps track of active tasks and requests from the `SessionDelegate` to the `SessionManager`, which has been renamed `Session`. The `SessionStateProvider` protocol has been added to abstract the info needed by the `SessionDelegate` to connect the `URLSession` delegate methods with `Request` actions. This PR also abstracts the `FileManager` used by the `Session` to be injectable and renames the `Alamofire` enum that was added to encapsulate the convenience API to `AF` in order to avoid collision with the module name.

### Testing Details :mag:
No new tests, though tests were updated to work around the new non-deterministic hashing in the Swift 4.2 compiler. In test mode, the hashing is made deterministic, though this isn't a long term solution for ordered headers.


Source/Alamofire.swift
```diff
@@ -25,7 +25,7 @@
 import Foundation
 
 /// Global namespace containing API for the `default` `SessionManager` instance.
-public enum Alamofire {
+public enum AF {
```
cnoon: 👏🏻👏🏻👏🏻

Source/SessionStateProvider.swift
```diff
@@ -24,138 +24,21 @@
...
-            task.suspend()
-            request.didSuspend()
-        }
-    }
+public protocol SessionStateProvider: AnyObject {
+    func request(for task: URLSessionTask) -> Request?
+    func didCompleteTask(_ task: URLSessionTask)
+    var serverTrustManager: ServerTrustManager? { get }
+    func credential(for task: URLSessionTask, protectionSpace: URLProtectionSpace) -> URLCredential?
 }
```
cnoon: Great change!

Source/Session.swift
```diff
@@ -252,10 +261,12 @@ open class SessionManager {
 
     open func upload(multipartFormData: @escaping (MultipartFormData) -> Void,
                 usingThreshold encodingMemoryThreshold: UInt64 = MultipartUpload.encodingMemoryThreshold,
+                fileManager: FileManager = .default,
                 with request: URLRequestConvertible) -> UploadRequest {
         let multipartUpload = MultipartUpload(isInBackgroundSession: (session.configuration.identifier != nil),
                                               encodingMemoryThreshold: encodingMemoryThreshold,
                                               request: request,
+                                              fileManager: fileManager,
                                               multipartBuilder: multipartFormData)
```
cnoon: Cool. No reason to not allow users to use the top-level APIs with custom file managers if needed. Especially since you're leveraging default values. I dig it.

Source/Session.swift
```diff
@@ -323,33 +334,164 @@ open class SessionManager {
...
+
+    public func resumeRequest(_ request: Request) {
+        rootQueue.async {
+            guard !request.isCancelled, let task = self.requestTaskMap[request] else { return }
+
+            task.resume()
+            request.didResume()
         }
     }
 }
```
cnoon: This is a huge improvement! Much easier to follow responsibilities.

Source/Alamofire.swift
MarcSteven: AF?

cnoon: Avoid confusion with the Alamofire namespace. Now it's no longer `Alamofire.Alamofire.request...`

Issue #2477, opened 2018-04-01
Enhance Certificate Pinning


jshier:
Though functionally complete, there are some enhancements Alamofire's certificate pinning implementation could use.

- Per task failures with a specific error, rather than cancellation errors for all pinning errors.
- Similarly, pinning should fail on a per attempt basis for things like invalid cert format.
- Requiring all requests to be pinned, as right now, if the domain isn't explicitly pinning, the request will proceed with default `URLSessionPinning`.

abhishekthanvi8:
Hi is there any proper documentation on this issue? How can I use SSL Certificate Pinning with Alamofire 5? Need Urgent Help on it. Thanks

tpian928:
I use Alamofire for mutual authentication and I can't get **NSURLAuthenticationMethodClientCertificate** callback in new 4.7.3 version
```
manager.delegate.sessionDidReceiveChallenge = { (session,challenge) in
    if challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust {
            // it's ok
    }
    else if challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodClientCertificate {
           // can't receive NSURLAuthenticationMethodClientCertificate
    }
)
```

PR #2608, opened 2018-09-24
Alamofire 5: Server Trust Errors

jshier:
### Issue Link :link:
#2477

### Goals :soccer:
This PR adds new, individual errors for various server trust failures, including reasons and captured payloads.

This PR also removes some old property list APIs.

`AFError.serverTrustEvaluationFailure` and `AFError.ServerTrustFailureReason` have been added. Additionally, evaluation of trust policies has been refactored into the application of a policy and then validation. These validations are now run sequentially, so each have an opportunity to throw an error, instead of being set as a single policy and evaluated at once.

This PR specially avoids wrapping the `Security` APIs like `SecTrustResultType` with better descriptions or other public APIs, given the moving target and breadth of coverage that would be required. Unfortunately this means the exposed error API isn't as helpful as it could be, as raw `SecTrust`, `OSStatus`, and `SecTrustResultType` values have no public properties.

### Testing Details :mag:
There are still 3 test failures, `testThatPinnedLeafCertificatePassesEvaluationWithHostValidation`, `testThatPinnedIntermediateCertificatePassesEvaluationWithHostValidation`, `testThatPinnedRootCertificatePassesEvaluationWithHostValidation` that are supposed to pass with host validation on but don't. 

TODO:
- [x] Additional `AFError` convenience APIs.
- [x] Inline documentation.
- [ ] Finalization of public APIs.
- [ ] Test for exact errors?


Source/ServerTrustEvaluation.swift
```diff
@@ -65,7 +81,8 @@ public protocol ServerTrustEvaluating {
     ///   - trust: The `SecTrust` value to evaluate.
     ///   - host:  The host for which to evaluate the `SecTrust` value.
     /// - Returns: A `Bool` indicating whether the evaluator considers the `SecTrust` value valid for `host`.
-    func evaluate(_ trust: SecTrust, forHost host: String) -> Bool
+    /// - Throws:  An `AFError.serverTrustEvaluationFailed` with an associated `ServerTrustFailureReason`.
+    func evaluate(_ trust: SecTrust, forHost host: String) throws -> Bool
     #endif
 }
```
cnoon: You could create a second protocol for Evaluators to handle `performDefaultEvaluation` and `validateHost` logic. Looks duplicated in `DefaultTrustEvaluator`, `RevocationTrustEvaluator`, `PinnedCertificatesTrustEvaluator`, and `PublicKeysTrustEvaluator`. Could be something like:

```swift
public protocol CommonServerTrustEvaluating: ServerTrustEvaluating {
    #if os(Linux)
    // Implement this once Linux has API for evaluating server trusts.
    #else
    var performDefaultValidation: Bool { get }
    var validateHost: Bool { get }

    func performDefaultValidationIfEnabled() throws -> Bool
    func validateHostIfEnabled() throws -> Bool
    #endif
}
```

Then in all the evaluations, this:

```swift
if performDefaultValidation {
    try trust.validate(policy: .default) { (status, result) in
        AFError.serverTrustEvaluationFailed(reason: .defaultEvaluationFailed(output: .init(host, trust, status, result)))
    }
}
        
if validateHost {
    try trust.validate(policy: .hostname(host)) { (status, result) in
        AFError.serverTrustEvaluationFailed(reason: .hostValidationFailed(output: .init(host, trust, status, result)))
    }
}
```

Could become this:

```swift
try performDefaultValidationIfEnabled()
try validateHostIfEnabled()
```

Or something also those lines as far as naming. Not critical or anything, just spotted a way to DRY it up.

Source/ServerTrustEvaluation.swift
```diff
@@ -104,17 +121,18 @@ public final class DefaultTrustEvaluator: ServerTrustEvaluating {
...
+        }
+
+        if validateHost {
+            try trust.validate(policy: .hostname(host)) { (status, result) in
+                AFError.serverTrustEvaluationFailed(reason: .hostValidationFailed(output: .init(host, trust, status, result)))
+            }
+        }
 
-        return trust.isValid
+        return true
```
cnoon: This logic might point out a small design flaw in the APIs. The `evaluate` method does not gather the `Bool` result anywhere from either `validate` call. It just returns `true` if it doesn't encounter an error. This seems to indicate that the APIs should probably just throw, not also return a `Bool`. Is there a specific reason to keep both? Potentially clients adding their own evaluations and choosing to just return `true` or `false`? I can't think of a great reason off the top of my head to keep that. If it's convenience for the test suite, we should be able to wrap the calls in the tests to offer a Bool option.

Source/ServerTrustEvaluation.swift
```diff
@@ -358,8 +412,53 @@ public extension SecTrust {
     var isValid: Bool {
```
cnoon: Pretty sure this is only used in the test suite. Any harm in moving this API into the test suite directly to avoid confusion? I was thinking this was being called since there's a duplicate `isValid` function.

Source/ServerTrustEvaluation.swift
```diff
@@ -188,47 +221,65 @@ public final class RevocationTrustEvaluator: ServerTrustEvaluating {
...
+        if validateHost {
+            try trust.validate(policy: .hostname(host)) { (status, result) in
+                AFError.serverTrustEvaluationFailed(reason: .hostValidationFailed(output: .init(host, trust, status, result)))
+            }
+        }
+        
+        let serverCertificatesData = Set(trust.certificateData)
+        let pinnedCertificatesData = Set(certificates.data)
+        let certificatesPinned = !serverCertificatesData.isDisjoint(with: pinnedCertificatesData)
+        if !certificatesPinned {
```
cnoon: Might be more clear to refactor `certificatesPinned` to `pinnedCertificatesFound` or `pinnedCertificateFoundInServerCertificates` or `pinnedCertFoundInServerCerts`.

Source/ServerTrustEvaluation.swift
```diff
@@ -239,53 +290,66 @@ public final class PinnedCertificatesTrustEvaluator: ServerTrustEvaluating {
...
+            try trust.validate(policy: .hostname(host)) { (status, result) in
+                AFError.serverTrustEvaluationFailed(reason: .hostValidationFailed(output: .init(host, trust, status, result)))
+            }
+        }
 
-        outerLoop: for serverPublicKey in trust.publicKeys as [AnyHashable] {
-            for pinnedPublicKey in keys as [AnyHashable] {
-                if serverPublicKey == pinnedPublicKey {
-                    return true
+        let keysPinned: Bool = {
```
cnoon: Same comment here about naming. Maybe `keysPinned` to `pinnedKeyFoundInServerKeys`.

Source/ServerTrustEvaluation.swift
```diff
@@ -358,8 +412,53 @@ public extension SecTrust {
...
+        return (status.isSuccess) ? (result == .unspecified || result == .proceed) : false
+    }
+    
+    @discardableResult
+    func validate(policy: SecPolicy, errorProducer: (_ status: OSStatus, _ result: SecTrustResultType) -> Error) throws -> Bool {
+        return try apply(policy: policy).isValid(errorProducer: errorProducer)
+    }
+
+    @discardableResult
+    func isValid(errorProducer: (_ status: OSStatus, _ result: SecTrustResultType) -> Error) throws -> Bool {
```
cnoon: I'm fairly certain the `Bool` return value causes more harm than good here with these APIs. It would be worth a good think as to whether we really need them or not.

Tests/ServerTrustEvaluatorTests.swift
```diff
@@ -468,10 +418,10 @@ class ServerTrustPolicyPerformDefaultEvaluationTestCase: ServerTrustPolicyTestCa
 
         // When
         setRootCertificateAsLoneAnchorCertificateForTrust(serverTrust)
-        let serverTrustIsValid = serverTrustPolicy.evaluate(serverTrust, forHost: host)
+        let result = Result { try serverTrustPolicy.evaluate(serverTrust, forHost: host) }
```
cnoon: Fun!

Source/ServerTrustEvaluation.swift
```diff
@@ -396,3 +518,7 @@ public extension SecCertificate {
         return SecTrustCopyPublicKey(createdTrust)
     }
 }
+
+extension OSStatus {
+    var isSuccess: Bool { return self == errSecSuccess }
+}
```
cnoon: Should we consider making all these extensions `public` for all those out there that would like to make their own custom evaluators? I certainly like the flexibility to control these APIs internally, but these are fantastic convenience functions for everyone.

Tests/TLSEvaluationTests.swift
```diff
@@ -165,8 +146,8 @@ class TLSEvaluationExpiredLeafCertificateTestCase: BaseTestCase {
         // Then
         XCTAssertNotNil(error, "error should not be nil")
 
-        if let error = error as? AFError {
-            XCTAssertTrue(error.isCertificatePinningError, "should be .certificatePinningFailed")
+        if let error = error?.asAFError {
+            XCTAssertTrue(error.isServerTrustEvaluationError, "should be .certificatePinningFailed")
```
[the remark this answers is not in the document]
jshier: I kept them where I could to serve as a reminder we should probably be testing these scenarios for specific errors. I can remove them if you want but I think we'll want something like them at some point. Once everything is merged and we're into the beta process I'd like to evaluate test coverage and ensure we're looking at specific error scenarios, since that's where we're missing a lot of coverage I think.

Source/ServerTrustEvaluation.swift
jshier: I've moved it, though it used some convenience API I had to redo manually.

Source/ServerTrustEvaluation.swift
```diff
@@ -358,8 +412,53 @@ public extension SecTrust {
     var isValid: Bool {
         var result = SecTrustResultType.invalid
         let status = SecTrustEvaluate(self, &result)
+        
+        return (status.isSuccess) ? (result == .unspecified || result == .proceed) : false
```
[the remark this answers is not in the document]
jshier: Updated, though this extension has moved.

Source/ServerTrustEvaluation.swift
```diff
@@ -65,7 +81,8 @@ public protocol ServerTrustEvaluating {
     ///   - trust: The `SecTrust` value to evaluate.
     ///   - host:  The host for which to evaluate the `SecTrust` value.
     /// - Returns: A `Bool` indicating whether the evaluator considers the `SecTrust` value valid for `host`.
-    func evaluate(_ trust: SecTrust, forHost host: String) -> Bool
+    /// - Throws:  An `AFError.serverTrustEvaluationFailed` with an associated `ServerTrustFailureReason`.
+    func evaluate(_ trust: SecTrust, forHost host: String) throws -> Bool
```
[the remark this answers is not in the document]
jshier: API feels better without it.

Source/ServerTrustEvaluation.swift
```diff
@@ -371,20 +470,43 @@ public extension SecTrust {
...
     /// The `Data` values for all certificates contained in `self`.
     var certificateData: [Data] {
+        return certificates.data
+    }
+
+    var certificates: [SecCertificate] {
         return (0..<SecTrustGetCertificateCount(self)).compactMap { index in
             SecTrustGetCertificateAtIndex(self, index)
-        }.data
+        }
```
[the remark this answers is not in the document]
jshier: Updated.

Source/ServerTrustEvaluation.swift
jshier: You're right, the API feels better without the return value.

Source/ServerTrustEvaluation.swift
jshier: Updated naming.

Source/ServerTrustEvaluation.swift
jshier: Updated naming.

Source/ServerTrustEvaluation.swift
jshier: Updated.

Source/ServerTrustEvaluation.swift
jshier: I've DRY'd up a little bit more by moving the actual evaluations into `SecTrust` extensions, but I don't think complicating the protocols to get rid of the `if` checks would be worth it. Take another look.

Source/AFError.swift
```diff
@@ -125,17 +121,50 @@ public enum AFError: Error {
...
     case responseSerializationFailed(reason: ResponseSerializationFailureReason)
-    case certificatePinningFailed
+    case serverTrustEvaluationFailed(reason: ServerTrustFailureReason)
+}
+
+public extension Error {
+    public var asAFError: AFError? {
+        return self as? AFError
+    }
 }
```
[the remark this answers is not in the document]
jshier: I've added a doc comment, since it's public.

jshier:
@cnoon Ready for review again. I made most of the suggested changes, see my comments for those I did not.

Tests/TLSEvaluationTests.swift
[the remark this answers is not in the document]
cnoon: Sounds good @jshier 👌🏻

Source/AFError.swift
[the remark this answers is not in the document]
cnoon: I completely agree that users will be more likely to try to cast Error to `AFError` to implement custom error handling logic. I'm just not sold yet that it warrants a separate API to do that. How exactly does this make it easier? Are you thinking that it's easier because there's now an auto-complete option to do it?

I'm not super opposed to this addition or anything, I'm just not sure why you see it as being "easier" to use `asAFError` instead of `as? AFError`. I'm also not sure why using the property is better than having casts everywhere. Is that because we'd be able to refactor it if needed in the future? Just trying to understand...

Source/ServerTrustEvaluation.swift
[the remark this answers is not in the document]
cnoon: Sweet!

Source/ServerTrustEvaluation.swift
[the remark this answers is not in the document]
cnoon: 👌🏻

Source/ServerTrustEvaluation.swift
cnoon: Indeed. They would certainly need a namespace of some sort. I lean towards the `.af` property. Not something we'd need to do in this PR, just wanted to ask.

Source/AFError.swift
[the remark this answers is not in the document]
jshier: A property is more discoverable, easier to use (autocomplete, like you said), and saves the consideration of `as?` for users who aren't familiar with it. It's the typically suggested pattern for dealing with specific error types you have to cast from `Error`. Like I said, I envision using it a lot just in our own testing, even if users don't.

jshier:
I'd like to offer the API, but I'll do that in a separate PR that introduces an extension protocol we can use for all extensions under a `.af` property.

PR #2620, opened 2018-10-10
Add optional encoding completion callback queue argument for multipart form upload

jaltreuter:
### Goals :soccer:
This change adds an optional `queue` parameter to the `upload(multipartFormData:...)` functions.  If specified, that queue will be used for the encoding completion callback.  Otherwise the main queue will be used, as is the current functionality.

The `queue` argument has a default value of nil, and I've replaced existing calls to `DispatchQueue.main.async` with `(queue ?? DispatchQueue.main).async`, following the pattern seen in `DataRequest.response`.  Therefore merging this pull request will have no effect to existing usage of the `upload` function; it adds flexibility where users need it.

### Testing Details :mag:
I will admit I've tested this only minimally due to the simplicity of the change.  From what I've seen, calling `upload` with or without a `queue` specified results in the encoding callback coming from the expected queue (the specified queue or main queue, respectively).

### DIFF
--- a/Source/SessionManager.swift
+++ b/Source/SessionManager.swift
@@ -611,6 +611,7 @@ open class SessionManager {
         to url: URLConvertible,
         method: HTTPMethod = .post,
         headers: HTTPHeaders? = nil,
+        queue: DispatchQueue? = nil,
         encodingCompletion: ((MultipartFormDataEncodingResult) -> Void)?)
     {
         do {
@@ -623,7 +624,7 @@ open class SessionManager {
                 encodingCompletion: encodingCompletion
             )
         } catch {
-            DispatchQueue.main.async { encodingCompletion?(.failure(error)) }
+            (queue ?? DispatchQueue.main).async { encodingCompletion?(.failure(error)) }
         }
     }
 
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR! Do you have a specific use case in mind?

Additionally, while we could add this now, it's completely unnecessary in Alamofire 5, as all request creation has been made asynchronous, and that includes the multipart encoding. There's no longer a separate encoding completion handler, any errors are just included in the response like normal.

jaltreuter:
@jshier my use case is in my app's implementation of background fetching. `application(_ application: performFetchWithCompletionHandler:)` is called on the main queue, and the system will complain (and I believe will be less likely to trigger background fetch in the future) if the function returns before the passed completion block is called.  I want to perform multiple uploads serially, so I perform those uploads on a background queue while sleeping on the main queue until background execution time runs out or the upload is complete.  Blocks dispatched asynchronously to the main queue can't execute while this routine is happening, so I need a way to avoid dispatching to the main queue.

Thanks for the quick response and heads up on the changes in AF5!  Do you have an estimate on when it'll be released?

jshier:
Interesting. We'll discuss adding this, though like I said, it's not something that will make the transition to Alamofire 5.

Speaking of, feel free to try the `alamofire5` branch and see how it works for you, especially for this case. There are a lot of renames, so you may need to check the code from time to time. In any case, there are a few more features we want to add before release, but I'm not sure when it will actually ship.

jshier:
I'm going to go ahead and merge this ahead of what will likely be our last Alamofire 4 release.

PR #2621, opened 2018-10-11
Add re-designed Package.swift manifest for swift 4

wanbok:
### Goals :soccer:
```
warning: PackageDescription API v3 is deprecated and will be removed in the future; used by package(s): Alamofire
```
This message has appeared when a user uses Alamofire via PackageManager with Swift 4.
Because of [Package Manager Manifest API(https://swift.org/blog/swift-package-manager-manifest-api-redesign/) has been redesigned].

Added the Package@swift-4.swift that will affect for above swift version 4.

### DIFF
--- a/Package@swift-4.swift
+++ b/Package@swift-4.swift
@@ -1,0 +1,41 @@
+// swift-tools-version:4.0
+
+//
+//  Package.swift
+//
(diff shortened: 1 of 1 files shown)


jshier:
We'll likely need to add a 4.2 version as well, but this is a good start.

PR #2625, opened 2018-10-15
Improve cURLRepresentation implementation readability

giulio92:
### Goals :soccer:
This small `PR` will improve `cURLRepresentation` `func` implementation in `Request` core class.

Current implementation use multiple `if let` statements to check for variable `nil`-ness and if available perform some `for in` loops. We could achieve the same behavior using Swift native `Optional` features. In fact if we use `.forEach` instead of `for in` there's no need to check for variable `nil`-ness because the loop won't execute if the variable will be `nil`.

### Testing Details :mag:
Regular unit tests should be enough

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -320,19 +320,19 @@ extension Request: CustomDebugStringConvertible {
 
         var headers: [AnyHashable: Any] = [:]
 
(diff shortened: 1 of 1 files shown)


jshier:
Made some changes, but otherwise thanks for the PR!

PR #2626, opened 2018-10-15
Alamofire5 signposts

amrMohamedFoad:


jshier:
Support for signposts will be reevaluated after AF5 release.

Issue #2232, opened 2017-08-07
HTTPHeader APIs


jshier:
It may be useful to create and expose an `HTTPHeader` type to make our header interactions more type-safe, as the header types exposed from Apple's networking APIs are usually `[AnyHashable : Any]`. This would allow us to expose things like convenience methods for creating common headers. Something like:
```swift
let headers = HTTPHeaders.contentType(.json)
                         .contentEncoding(.all)
                         .userAgent("UserAgentString")
```

jshier:
I've implemented a preliminary version of the `HTTPHeaders` and `HTTPHeader` types. There are still some unanswered questions:

1. Do we need `HTTPHeaders`? We could just extend `Array where Element == HTTPHeader` and generate the dictionary that way. But if we wanted to offer a more builder-like API, `HTTPHeaders` is where it would live.
2. Do we want to consider name / value pairs as one type? I currently have them as separate types, which allows for maximum flexibility at the cost of a bit of safety.
3. How far do we go with providing static values? I included some examples of what we could provide, but we'd probably want to have them static values live elsewhere rather than polluting the type definitions with them, but I'm not sure.

```swift
struct HTTPHeaders {
    
    private var headers: [HTTPHeader] = []
    
    func asDictionary() -> [String : String] {
        return Dictionary.init(zip(headers.map { $0.name.rawValue }, headers.map { $0.value.rawValue }),
                               uniquingKeysWith: { (_, last) in last })
    }
    
}

extension HTTPHeaders: ExpressibleByArrayLiteral {
    
    init(arrayLiteral elements: HTTPHeader...) {
        headers = elements
    }
    
}

struct HTTPHeader {
    
    let name: Name
    let value: Value
    
    struct Name {
        
        let rawValue: String
        
        init(_ rawValue: String) {
            self.rawValue = rawValue
        }
        
        static let contentEncoding = Name("Content-Encoding") // Could be placed elsewhere
        
    }
    
    struct Value {
        
        let rawValue: String
        
        init(_ rawValue: String) {
            self.rawValue = rawValue
        }
        
        static let json = Value("application/json") // Could be placed elsewhere
        static let xml = Value("application/xml")
        
    }
    
}

extension HTTPHeader {
    
    static func contentEncoding(_ value: Value) -> HTTPHeader {
        return HTTPHeader(name: .contentEncoding, value: value)
    }
    
}

let headers: HTTPHeaders = [.contentEncoding(.json),
[...]

PR #2629, opened 2018-10-17
Alamofire 5: Real HTTPHeaders type.

jshier:
### Issue Link :link:
#2232 

### Goals :soccer:
This PR adds real `HTTPHeaders` and `HTTPHeader` types in an attempt to allow some convenience for users around headers.

This PR allows the following:
```swift
let headers: HTTPHeaders = ["Content-Type": "application/json"]
let headers: HTTPHeaders = [.contentType("application/json")]
```
`HTTPHeaders` also conforms to `Collection`, so it's iterable and indexable.

`HTTPHeaders` is a simple struct that wraps an `[HTTPHeader]` allowing us to maintain ordering and support sorting. It also provides convenience API for conversion to and from `[String: String]`, as well as `[HTTPHeader]`. These APIs are somewhat simpler than my previous attempt in the `httpheaders` branch, and talked about in #2232. This PR doesn't create `HTTPHeader.Name` or `HTTPHeader.Value` types, which means we get away with much less API surface. However, this does mean we can't quite offer the same level of convenience, but it's much simpler overall.

### Testing Details :mag:
Additional tests have been added under `HTTPHeadersTests`.


Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
 
 import Foundation
 
-public typealias HTTPHeaders = [String: String]
+public struct HTTPHeaders {
+    private var headers = [HTTPHeader]()
+    
+    public init() { }
+    
+    public init(_ headers: [HTTPHeader]) {
```
philtre: This initializer is the only place where it is possible to insert headers with duplicate names. All other initializers and mutating functions perform deduplication.

Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
...
+    public var dictionary: [String: String] {
+        let namesAndValues = headers.map { ($0.name, $0.value) }
+        
+        return Dictionary(namesAndValues, uniquingKeysWith: { (_, last) in last })
+    }
+}
+
+extension HTTPHeaders: ExpressibleByArrayLiteral {
+    public init(arrayLiteral elements: HTTPHeader...) {
+        self.init()
```
philtre: If deduplication is to be performed in `init(_ headers: [HTTPHeader])`, this can be changed to: `self.init(elements)`

Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
...
+}
+
+// MARK: - Defaults
+
+extension HTTPHeaders {
+    public static let `default`: HTTPHeaders = [.defaultAcceptEncoding,
+                                                .defaultAcceptLanguage,
+                                                .defaultUserAgent]
+    
     public static func authorization(username: String, password: String) -> HTTPHeaders {
```
philtre: This convenience method should probably be moved to `HTTPHeader`

Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
...
+    }
+    
+    public mutating func remove(name: String) {
+        guard let index = headers.index(of: name) else { return }
+        
+        headers.remove(at: index)
+    }
+    
+    mutating func sort() {
+        return headers.sort()
```
philtre: I wonder if it might be better to use custom sorting instead of adding `Comparable` conformance to `HTTPHeader`.
`return headers.sort{ $0.name < $1.name }`

Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
...
+    public let value: String
+}
+
+extension HTTPHeader: CustomStringConvertible {
+    public var description: String {
+        return "\(name): \(value)"
+    }
+}
+
+extension HTTPHeader: Comparable {
```
philtre: Not sure if it's entirely correct to conform `HTTPHeader` to `Comparable` if we're not also comparing the value. Since we're only sorting headers in one method, it might be better to use a custom sorting function (See line 62)

Source/HTTPHeaders.swift
jshier: Yes, this is an awkward place, as I ran into in the tests.

Source/HTTPHeaders.swift
jshier: Ah, because the synthesized `Equatable` does include `value`? I don't see an issue with including it here too, even though `name` should be enough.

philtre: Yes, synthesized `Equatable` compares both values, but also, you're assuming that *any* collection of `HTTPHeader` will be free of duplicate names.
Consider this test:

```swift
let headersArr = [
    HTTPHeader(name: "foo", value: "a"),
    HTTPHeader(name: "foo", value: "b"),
    HTTPHeader(name: "foo", value: "c"),
]

var headers1 = HTTPHeaders()
var headers2 = HTTPHeaders()
var headers3 = HTTPHeaders()

for header in headersArr {
    headers1.update(header)
}

for header in headersArr.sorted() {
    headers2.update(header)
}

for header in headersArr.reversed().sorted() {
    headers3.update(header)
}
```

Currently, the result is:
```
headers1: ["foo": "c"]
headers2: ["foo": "c"]
headers3: ["foo": "a"]
```

The expected result would be:
```
headers1: ["foo": "c"]
headers2: ["foo": "c"]
headers3: ["foo": "c"]
```
... with the following change:
```swift
extension HTTPHeader : Comparable {
    public static func <(lhs: HTTPHeader, rhs: HTTPHeader) -> Bool {
        return lhs.name == rhs.name ? lhs.value < rhs.value : lhs.name < rhs.name
    }
}
```

Alternatively, since `HTTPHeaders` guarantees that names will be unique, we could remove `Comparable` conformance and change the `sort` function to:
```swift
mutating func sort() {
    return headers.sort{ $0.name < $1.name }
}
```

Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
...
+        self.headers = headers
+    }
+    
+    public init(_ dictionary: [String: String]) {
+        self.init()
+        
+        headers = dictionary.map { HTTPHeader(name: $0.key, value: $0.value) }
+    }
+    
+    public mutating func update(name: String, value: String) {
```
philtre: Currently, `lowercased()` is being called twice for each comparison. It would make more sense to inline the comparison function and also make `update(name:value:)` the convenience function instead of `update(_:)`:

```swift
    mutating func update(_ header: HTTPHeader) {
        let lowercasedName = header.name.lowercased()
        guard let index = headers.firstIndex(where: { $0.name.lowercased() == lowercasedName } ) else {
            headers.append(header)
            return
        }
        headers.replaceSubrange(index...index, with: [header])
    }
    mutating func update(name: String, value: String) {
        update(HTTPHeader(name: name, value: value))
    }
```

Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
...
+    
+    mutating func sort() {
+        return headers.sort()
+    }
+    
+    func sorted() -> HTTPHeaders {
+        return HTTPHeaders(headers.sorted())
+    }
+    
+    func value(for name: String) -> String? {
```
philtre: To avoid calling `lowercased()` twice for each comparison and by using `first(where:)`:

```swift
func value(for name: String) -> String? {
    let lowercasedName = name.lowercased()
    return headers.first(where: { $0.name.lowercased() == lowercasedName} )?.value
}
```

Source/HTTPHeaders.swift
```diff
@@ -24,51 +24,215 @@
...
+        }
+        
+        headers.replaceSubrange(index...index, with: [HTTPHeader(name: name, value: value)])
+    }
+    
+    public mutating func update(_ header: HTTPHeader) {
+        update(name: header.name, value: header.value)
+    }
+    
+    public mutating func remove(name: String) {
```
philtre: To avoid calling `lowercased()` twice for each comparison:

```swift
mutating func remove(name: String) {
    let lowercasedName = name.lowercased()
    guard let index = headers.firstIndex(where: { $0.name.lowercased() == lowercasedName } ) else { return }
    headers.remove(at: index)
}
```

Source/HTTPHeaders.swift
jshier: I've switched the convenience but I made the `lowercased` change in my `index(of:)` extension.

Source/HTTPHeaders.swift
jshier: Done in the `Array` extension.

Source/HTTPHeaders.swift
jshier: Done in the `Array` extension.

Source/HTTPHeaders.swift
jshier: You're right, and I've made the change for manual comparison.

Source/HTTPHeaders.swift
jshier: Done.

Source/HTTPHeaders.swift
jshier: Changed, and everything is much nicer.

Source/HTTPHeaders.swift
jshier: Yep, changed.

jshier:
I've added inline documentation and tests, so I think this PR is ready for full review.

Source/HTTPHeaders.swift
```diff
@@ -24,40 +24,301 @@
...
+    ///
+    /// - Parameters:
+    ///   - username: The username of the header.
+    ///   - password: The password of the header.
+    /// - Returns:    The header.
+    public static func authorization(username: String, password: String) -> HTTPHeader {
+        let credential = Data("\(username):\(password)".utf8).base64EncodedString()
+
+        return authorization("Basic \(credential)")
+    }
```
cnoon: Would it make sense for the sake of completeness to add a third `authorization(bearerToken: String) -> HTTPHeader` API?

Source/HTTPHeaders.swift
```diff
@@ -24,40 +24,301 @@
...
+    /// Returns an `Accept-Language` header.
+    ///
+    /// Alamofire offers a default Accept-Language header that accumulates and encodes the system's preferred languages.
+    /// Use `HTTPHeader.defaultAcceptLanguage`.
+    ///
+    /// - Parameter value: The `Accept-Language` value.
+    /// - Returns:         The header.
+    public static func acceptLanguage(_ value: String) -> HTTPHeader {
+        return HTTPHeader(name: "Accept-Language", value: value)
+    }
```
cnoon: Could also add an `Accept-Charset` for completeness. That's the only other one we have in our internal Nike library that would make sense for the public I think.

Tests/HTTPHeadersTests.swift
```diff
@@ -0,0 +1,134 @@
+//
+//  HTTPHeaders.swift
```
cnoon: `HTTPHeaders.swift` -> `HTTPHeadersTests.swift`.

Source/HTTPHeaders.swift
```diff
@@ -24,40 +24,301 @@
...
+    /// - Parameter name: The name of the `HTTPHeader` to remove.
+    public mutating func remove(name: String) {
+        guard let index = headers.index(of: name) else { return }
+
+        headers.remove(at: index)
+    }
+
+    /// Sort the current instance by header name.
+    mutating public func sort() {
+        return headers.sort { $0.name < $1.name }
```
cnoon: I think you could ditch the `return` here.

Source/HTTPHeaders.swift
```diff
@@ -24,40 +24,301 @@
...
+    /// Case-insensitively find a header's value by name.
+    ///
+    /// - Parameter name: The name of the header to search for, case-insensitively.
+    /// - Returns:        The value of header, if it exists.
+    public func value(for name: String) -> String? {
+        guard let index = headers.index(of: name) else { return nil }
+
+        return headers[index].value
+    }
+
```
cnoon: Extra newline.

Source/HTTPHeaders.swift
```diff
@@ -24,40 +24,301 @@
...
+    /// Create an instance from a `[String: String]`. Duplicate case-insensitive names are collapsed into the last name
+    /// and value encountered.
+    public init(_ dictionary: [String: String]) {
+        self.init()
+
+        dictionary.forEach { update(HTTPHeader(name: $0.key, value: $0.value)) }
+    }
+
```
cnoon: I can't help but think the `HTTPHeaders` type is missing an `add(_ header: HTTPHeader)` API. I get that `update` can be used for both, but it almost seems like an `add` API makes sense as well for completion sake that could simply call `update` internally. Thoughts? I'm sure you've already rolled this around...

Source/HTTPHeaders.swift
```diff
@@ -103,8 +364,40 @@ extension Dictionary where Key == String, Value == String {
...
+    }
+}
+
+extension URLSessionConfiguration {
+    /// Returns `httpAdditionalHeaders` as `HTTPHeaders`.
+    public var httpHeaders: HTTPHeaders {
+        get { return (httpAdditionalHeaders as? [String: String]).map(HTTPHeaders.init) ?? HTTPHeaders() }
+        set { httpAdditionalHeaders = newValue.dictionary }
+    }
+}
```
cnoon: Great additions!

Source/MultipartFormData.swift
```diff
@@ -549,12 +546,12 @@ open class MultipartFormData {
...
 
-    private func contentHeaders(withName name: String, fileName: String? = nil, mimeType: String? = nil) -> [String: String] {
+    private func contentHeaders(withName name: String, fileName: String? = nil, mimeType: String? = nil) -> HTTPHeaders {
         var disposition = "form-data; name=\"\(name)\""
         if let fileName = fileName { disposition += "; filename=\"\(fileName)\"" }
 
-        var headers = ["Content-Disposition": disposition]
-        if let mimeType = mimeType { headers["Content-Type"] = mimeType }
+        var headers: HTTPHeaders = [.contentDisposition(disposition)]
+        if let mimeType = mimeType { headers.update(.contentType(mimeType)) }
```
cnoon: In this case, it would probably make more sense to see `headers.add(.contentType(mimeType))` rather than `update`.

Source/Response.swift
```diff
@@ -89,14 +89,13 @@ extension DataResponse: CustomStringConvertible, CustomDebugStringConvertible {
...
+            let sortedHeaders = response.httpHeaders.sorted()
 
             return """
-            Status Code: \(response.statusCode)
-                        Headers: \(sortedHeaders)
-            """
+                   Status Code: \(response.statusCode)
+                   Headers:
+                   \(sortedHeaders)
+                   """
```
cnoon: Extra indent I believe.

Source/Response.swift
```diff
@@ -274,14 +273,13 @@ extension DownloadResponse: CustomStringConvertible, CustomDebugStringConvertibl
...
+            let sortedHeaders = response.httpHeaders.sorted()
 
             return """
-            Status Code: \(response.statusCode)
-            Headers: \(sortedHeaders)
-            """
+                   Status Code: \(response.statusCode)
+                   Headers:
+                   \(sortedHeaders)
+                   """
```
cnoon: Extra indent.

Tests/HTTPHeadersTests.swift
```diff
@@ -0,0 +1,134 @@
...
+        XCTAssertEqual(headers.count, 3)
+    }
+
+    func testHeadersPreserveLastFormAndValueOfAName() {
+        // Given
+        var headers: HTTPHeaders = ["c": "a"]
+
+        // When
+        headers["C"] = "c"
+
```
cnoon: Extra newline.

Tests/SessionManagerTests.swift
```diff
@@ -195,7 +193,7 @@ class SessionManagerTestCase: BaseTestCase {
 
     func testDefaultUserAgentHeader() {
         // Given, When
-        let userAgent = HTTPHeaders.defaultHTTPHeaders["User-Agent"]
+        let userAgent = HTTPHeaders.default["User-Agent"]
```
cnoon: As a separate effort, it might be worth trying to pull the header fields into a strongly typed object instead of using magic strings. Something like a `HTTPHeader.Field` type.

```swift
struct HTTPHeader {
    struct Field {
        static let userAgent = "User-Agent"
    }
}
```

Etc. etc... This is something we did in our internal networking library and it's been helpful.

Tests/HTTPHeadersTests.swift
jshier: Fixed in [aaa3e6a](
https://github.com/Alamofire/Alamofire/pull/2629/commits/aaa3e6a10d2b14ad0dbc049dc5aa3b1b9bb8fa2e).

Source/HTTPHeaders.swift
jshier: Fixed in [aaa3e6a](
https://github.com/Alamofire/Alamofire/pull/2629/commits/aaa3e6a10d2b14ad0dbc049dc5aa3b1b9bb8fa2e).

Tests/HTTPHeadersTests.swift
jshier: Fixed in [aaa3e6a](
https://github.com/Alamofire/Alamofire/pull/2629/commits/aaa3e6a10d2b14ad0dbc049dc5aa3b1b9bb8fa2e).

Source/HTTPHeaders.swift
jshier: Added in [aaa3e6a](
https://github.com/Alamofire/Alamofire/pull/2629/commits/aaa3e6a10d2b14ad0dbc049dc5aa3b1b9bb8fa2e).

Source/HTTPHeaders.swift
jshier: Sure, added in [aaa3e6a](
https://github.com/Alamofire/Alamofire/pull/2629/commits/aaa3e6a10d2b14ad0dbc049dc5aa3b1b9bb8fa2e).

jshier:
@cnoon Ready for another look.

Source/HTTPHeaders.swift
jshier: Updated.

Source/Response.swift
cnoon: Ah...we'll catch all this in a SwiftLint cleanup before the release.

Issue #1713, opened 2016-10-20
Head request in Alamofire 4


muruganandham:
The below head request was throwing following error   

**responseSerializationFailed(Alamofire.AFError.ResponseSerializationFailureReason.inputDataNilOrZeroLength)**

```
  Alamofire.request(AppConstants.musicFileURL + musicFilename, method: .head, parameters: nil, encoding: JSONEncoding.default, headers:headers)

            .responseJSON { responseStr in

                print("HEAD \(responseStr)")

                //print FAILURE: responseSerializationFailed(Alamofire.AFError.ResponseSerializationFailureReason.inputDataNilOrZeroLength)

                guard responseStr.result.isSuccess else {
                    return
                }
        }
```

jshier:
Thanks for the report! HEAD requests are just rare enough to get through our testing in this instance. We'll work on an update this weekend.

luizv:
I got the same problem here. I was supposing the .head request would have a specific response type other than JSON, since it doesn't have a body to parse. But I can't find documentation about it.

Any thoughts or directions?

cnoon:
@jshier has put together #1775 to address this issue. We're not going to pull this PR into the Alamofire 4.1.0 because we still have some backwards compatibility issues to work out. However, I'd like to close this issue out and have everyone redirect all further comments to #1775.

Thanks everyone! 🍻

PR #2639, opened 2018-11-13
Alamofire 5: Customizable Empty Response Handling

jshier:
### Issue Link :link:
#1713

### Goals :soccer:
This PR adds customization points to the various serializers to support custom response code and request method handling for empty responses, while providing default values.

This PR refactors some of the response parsing, while keeping all but one test passing exactly the same. It adds properties to all of the response serializers to let users customize them at initialization. It also adds support for request method parsing, to support empty responses for HEAD requests.

### Testing Details :mag:
Test cases cover this code pretty thoroughly. One test was inverted, as file and data response parsing are now the same by default.

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -47,6 +47,11 @@ public protocol DownloadResponseSerializerProtocol {
 /// A serializer that can handle both data and download responses.
 public protocol ResponseSerializer: DataResponseSerializerProtocol & DownloadResponseSerializerProtocol { }
 
+extension ResponseSerializer {
+    public static var defaultEmptyResponseCodes: Set<Int> { return emptyResponseCodes }
+    public static var defaultEmptyRequestMethods: Set<HTTPMethod> { return emptyRequestMethods }
+}
+
 /// By default, any serializer declared to conform to both types will get file serialization for free, as it just feeds
 /// the data read from disk into the data response serializer.
 public extension DownloadResponseSerializerProtocol where Self: DataResponseSerializerProtocol {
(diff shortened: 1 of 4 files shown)


Source/ParameterEncoder.swift
```diff
@@ -221,6 +221,25 @@ public final class URLEncodedFormEncoder {
...
+        case plusReplaced
+        
+        /// Encodes the string according to the encoding.
+        ///
+        /// - Parameter string: The `String` to encode.
+        /// - Returns:          The encoded `String`.
+        func encode(_ string: String) -> String {
+            switch self {
+            case .percentEscaped: return string.replacingOccurrences(of: " ", with: "%20")
+            case .plusReplaced: return string.replacingOccurrences(of: " ", with: "+")
```
agnosticdev: Is the `plusReplaced`  case used to support encoding types for `application/x-www-form-urlencoded `  ?

jshier: It's actually a commit from the parent of this PR's branch (#2637). I rebranched and it made it part of this PR too.

Source/ResponseSerialization.swift
```diff
@@ -303,19 +308,40 @@ extension DataRequest {
...
-        guard let validData = data else {
-            throw AFError.responseSerializationFailed(reason: .inputDataNil)
+        
+        guard let data = data, !data.isEmpty else {
+            let requestAllowsEmptyResponses = request.flatMap { $0.httpMethod }
+                                                     .flatMap(HTTPMethod.init)
+                                                     .map { emptyRequestMethods.contains($0) }
+            let responseAllowsEmptyResponses = response.flatMap { $0.statusCode }
+                                                       .map { emptyResponseCodes.contains($0) }
+            if requestAllowsEmptyResponses ?? responseAllowsEmptyResponses ?? false {
```
cnoon: This is a great change overall! I saw some potential for reuse here given that each of the four response serializers repeated this logic. I took a stab at moving it all into the `ResponseSerializer` protocol and the protocol extension in PR #2646. It makes this logic more reusable for custom response serializers both inside and outside AF. Let me know what you think.

Source/ResponseSerialization.swift
```diff
@@ -303,19 +308,40 @@ extension DataRequest {
...
+    ///
+    /// - Parameters:
+    ///   - emptyResponseCodes:  The HTTP response codes for which empty responses are allowed. Defaults to
+    ///                          `[204, 205]`.
+    ///   - emptyRequestMethods: The HTTP request methods for which empty responses are allowed. Defaults to `[.head]`.
+    public init(emptyResponseCodes: Set<Int> = DataResponseSerializer.defaultEmptyResponseCodes,
+                emptyRequestMethods: Set<HTTPMethod> = DataResponseSerializer.defaultEmptyRequestMethods) {
+        self.emptyResponseCodes = emptyResponseCodes
+        self.emptyRequestMethods = emptyRequestMethods
+    }
```
cnoon: Really good idea to allow them to be injected!

jshier:
@cnoon This is ready for review again. I've merged your PR and made some updates. Namely:

- Removed the static default values as customization points on the `ResponseSerializer` protocol. It doesn't really make sense for those to be requirements as well as instance values, especially since we're not offering any functionality that takes advantage of them.
- Added an additional `emptyResponseAllowed(forRequest:response:)` method to not force users to use the same nil coalescing chain we had before.

Source/ResponseSerialization.swift
jshier: Updated.

Source/ResponseSerialization.swift
```diff
@@ -570,4 +652,5 @@ extension DataRequest {
 }
 
 /// A set of HTTP response status code that do not contain response data.
-private let emptyDataStatusCodes: Set<Int> = [204, 205]
+private let emptyResponseCodes: Set<Int> = [204, 205]
+private let emptyRequestMethods: Set<HTTPMethod> = [.head]
```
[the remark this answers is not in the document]
jshier: Moved into the suggested protocol.

Source/ResponseSerialization.swift
```diff
@@ -45,7 +45,33 @@ public protocol DownloadResponseSerializerProtocol {
...
 /// A serializer that can handle both data and download responses.
-public protocol ResponseSerializer: DataResponseSerializerProtocol & DownloadResponseSerializerProtocol { }
+public protocol ResponseSerializer: DataResponseSerializerProtocol & DownloadResponseSerializerProtocol {
+    var emptyRequestMethods: Set<HTTPMethod> { get }
+    var emptyResponseCodes: Set<Int> { get }
+}
+
+extension ResponseSerializer {
+    public static var defaultEmptyRequestMethods: Set<HTTPMethod> { return [.head] }
+    public static var defaultEmptyResponseCodes: Set<Int> { return [204, 205] }
```
cnoon: 👌🏻

Source/ResponseSerialization.swift
```diff
@@ -45,7 +45,33 @@ public protocol DownloadResponseSerializerProtocol {
...
+    }
+
+    public func responseAllowsEmptyResponseData(_ response: HTTPURLResponse?) -> Bool? {
+        return response.flatMap { $0.statusCode }
+                       .map { emptyResponseCodes.contains($0) }
+    }
+
+    public func emptyResponseAllowed(forRequest request: URLRequest?, response: HTTPURLResponse?) -> Bool {
+        return requestAllowsEmptyResponseData(request) ?? responseAllowsEmptyResponseData(response) ?? false
+    }
```
cnoon: Very wise...great change!

Source/ResponseSerialization.swift
```diff
@@ -347,23 +329,38 @@ extension DownloadRequest {
 /// data is considered an error. However, if the response is has a status code valid for empty responses (`204`, `205`),
 /// then an empty `String` is returned.
 public final class StringResponseSerializer: ResponseSerializer {
-    let encoding: String.Encoding?
-
-    /// Creates an instance with the given `String.Encoding`.
+    /// Optional string encodeing used to validate the response.
```
cnoon: `encodeing` -> `encoding`.

PR #2646, opened 2018-11-21
Alamofire 5: Empty Response Refactor (2nd Refactor)

cnoon:
This PR refactors the empty response refactor to move the logic all into the `ResponseSerializer` protocol and protocol extension. This change DRYs up the code and exposes all the functionality in a more convenient way for custom `ResponserSerializer` types.

I also removed the `AnyResponseSerializer` type since it no longer appears to be in use. Forgive me if this is not the case.

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -77,66 +77,6 @@ public extension DownloadResponseSerializerProtocol where Self: DataResponseSeri
     }
 }
 
-// MARK: - AnyResponseSerializer
-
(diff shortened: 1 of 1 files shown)


jshier:
My only concern with placing all of this under the `ResponseSerializer` protocol is that, if users don't want conformance with one or the other of the protocols, they lose the functionality. But if this is mostly for us, then yes, this looks good.

jshier:
I'm going to merge and then build off of this, as I think there are a few bits to improve.

MarcSteven:
Cool refactor!!Great design

PR #2662, opened 2018-12-07
Implemented protocol EmptyResponse so that an empty response value ca…

jvannoord:
### Goals :soccer:
An empty result should not need to be derived from the `Empty` type.

It is inconvenient (and restrictive) to require that all empty response value derived form `Empty`. Using a protocol allows any response value type to declare its own empty value.

The `EmptyResponse` protocol essential does what the `Empty` struct did; and the `Empty` struct has been extended to implement `EmptyResponse`. No implementations of Empty should be adversely affected. If we're worried about var vs let on the static property, it seems like a static func in its place would probable solve it.

### Testing Details :mag:
Added tests to test for objects that implement `EmptyResponse` and objects that incorrectly implement `EmptyResponse.`

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -583,11 +583,11 @@ public final class DecodableResponseSerializer<T: Decodable>: ResponseSerializer
                 throw AFError.responseSerializationFailed(reason: .inputDataNilOrZeroLength)
             }
 
-            guard let emptyResponseType = T.self as? EmptyResponse.Type else {
+            guard let emptyResponseType = T.self as? EmptyResponse.Type, let emptyValue = emptyResponseType.emptyValue as? T else {
                 throw AFError.responseSerializationFailed(reason: .invalidEmptyResponse(type: "\(T.self)"))
             }
 
-            return emptyResponseType.emptyValue as! T
+            return emptyValue
         }
 
         do {
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -515,10 +515,18 @@ extension DownloadRequest {
 }
 
 // MARK: - Empty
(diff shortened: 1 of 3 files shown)


Source/ResponseSerialization.swift
```diff
@@ -575,11 +583,11 @@ public final class DecodableResponseSerializer<T: Decodable>: ResponseSerializer
                 throw AFError.responseSerializationFailed(reason: .inputDataNilOrZeroLength)
             }
 
-            guard let emptyValue = Empty.value as? T else {
+            guard let emptyResponseType = T.self as? EmptyResponse.Type else {
```
jshier: It seems like there should be a more elegant want to handle this. At the least, perhaps combining the type cast and the value cast into the guard?

jvannoord: I updated the guard in 9a7bb81 just now. It'll take me a little time to get to writing the tests. Thanks for seeing the value in this.

jvannoord:
@jshier I added some tests. I didn't add anything specifically involving `Bool`, though.

jshier:
@jvannoord I only used `Bool` as an example of a type you could create a conformance for without having to write your own. I think the tests would be much cleaner using such types rather than custom written ones. Additionally, I mainly wanted to test that a conforming type was able to be used successfully while a non-conforming one could not.

Also, I need to do some renames for those tests, so let me get that in and you can rebase.

jshier:
@jvannoord I've pushed a small refactor for the tests, you can rebase. For the tests you're adding, I'd suggest just making `DecodableValue` conform to `EmptyResponse` and use that for most tests, and use a non-conforming type for a failure tests. I don't think the tests with different conforming types are necessary, especially since there's a test with `Empty` already.

jvannoord:
This PR is superseded by https://github.com/Alamofire/Alamofire/pull/2664.

PR #2664, opened 2018-12-12
Implementing protocol EmptyResponse so that an empty response value can be returned when an empty http status code is encountered

jvannoord:
### Goals :soccer:
An empty result should not need to be derived from the `Empty` type.

It is inconvenient (and restrictive) to require that all empty response value derived form `Empty`. Using a protocol allows any response value type to declare its own empty value.

The `EmptyResponse` protocol essential does what the `Empty` struct did; and the `Empty` struct has been extended to implement `EmptyResponse`. No implementations of Empty should be adversely affected. 

This PR supersedes https://github.com/Alamofire/Alamofire/pull/2662, which was originally based the `alamofire5` branch and had a number of requested changes.

### Testing Details :mag:
Added tests to test for objects that implement `EmptyResponse` and objects that incorrectly implement `EmptyResponse.` `Bool` is extended in order to test a built-in type; `Int` is not extended and tested for failure.

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -517,7 +517,7 @@ extension DownloadRequest {
 // MARK: - Empty
 /// A protocol for a type representing an empty response. Use `T.emptyValue` to get an instance.
 public protocol EmptyResponse {
-    static var emptyValue: EmptyResponse { get }
+    static func emptyResponseValue() -> Self
 }
 
 /// A type representing an empty response. Use `Empty.value` to get the instance.
@@ -526,7 +526,10 @@ public struct Empty: Decodable {
 }
 
 extension Empty: EmptyResponse {
-    public static var emptyValue: EmptyResponse { return value }
+    public static func emptyResponseValue() -> Empty {
+        return value
+    }
+    public static var emptyResponseValueVar: Empty { return value }
 }
 
 // MARK: - DataDecoder Protocol
@@ -583,11 +586,11 @@ public final class DecodableResponseSerializer<T: Decodable>: ResponseSerializer
                 throw AFError.responseSerializationFailed(reason: .inputDataNilOrZeroLength)
             }
 
-            guard let emptyResponseType = T.self as? EmptyResponse.Type, let emptyValue = emptyResponseType.emptyValue as? T else {
+            guard let emptyResponseType = T.self as? EmptyResponse.Type else {
                 throw AFError.responseSerializationFailed(reason: .invalidEmptyResponse(type: "\(T.self)"))
             }
 
-            return emptyValue
+            return emptyResponseType.emptyResponseValue() as! T
         }
 
         do {
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -515,10 +515,18 @@ extension DownloadRequest {
 }
 
 // MARK: - Empty
+/// A protocol for a type representing an empty response. Use `T.emptyValue` to get an instance.
+public protocol EmptyResponse {
+    static var emptyValue: EmptyResponse { get }
+}
 
 /// A type representing an empty response. Use `Empty.value` to get the instance.
 public struct Empty: Decodable {
     public static let value = Empty()
+}
+
+extension Empty: EmptyResponse {
+    public static var emptyValue: EmptyResponse { return value }
 }
 
 // MARK: - DataDecoder Protocol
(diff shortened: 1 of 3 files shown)


Source/ResponseSerialization.swift
```diff
@@ -515,12 +515,20 @@ extension DownloadRequest {
 }
 
 // MARK: - Empty
+/// A protocol for a type representing an empty response. Use `T.emptyValue` to get an instance.
+public protocol EmptyResponse {
+    static var emptyValue: EmptyResponse { get }
```
jshier: In order to work around the issue you identified in the tests, where types could accidentally return some other `EmptyResponse` type, perhaps this property should have type `Self`? I just don't know if that would break the casting that needed to pull the value out.

jvannoord: Guess I hadn't tried that. I'll play around with it.

Source/ResponseSerialization.swift
```diff
@@ -515,12 +515,23 @@ extension DownloadRequest {
 }
 
 // MARK: - Empty
+/// A protocol for a type representing an empty response. Use `T.emptyValue` to get an instance.
+public protocol EmptyResponse {
+    static func emptyResponseValue() -> Self
```
jshier: I like the transition to a function here, but this seems rather verbose. Perhaps just `emptyValue()`?

Source/ResponseSerialization.swift
```diff
@@ -515,12 +515,23 @@ extension DownloadRequest {
...
 /// A type representing an empty response. Use `Empty.value` to get the instance.
 public struct Empty: Decodable {
     public static let value = Empty()
 }
 
+extension Empty: EmptyResponse {
+    public static func emptyResponseValue() -> Empty {
+        return value
+    }
+    public static var emptyResponseValueVar: Empty { return value }
```
jshier: This looks to be unused.

Source/ResponseSerialization.swift
```diff
@@ -575,11 +586,11 @@ public final class DecodableResponseSerializer<T: Decodable>: ResponseSerializer
                 throw AFError.responseSerializationFailed(reason: .inputDataNilOrZeroLength)
             }
 
-            guard let emptyValue = Empty.value as? T else {
+            guard let emptyResponseType = T.self as? EmptyResponse.Type else {
                 throw AFError.responseSerializationFailed(reason: .invalidEmptyResponse(type: "\(T.self)"))
             }
 
-            return emptyValue
+            return emptyResponseType.emptyResponseValue() as! T
```
jshier: This can again be combined into the guard.

Tests/ResponseSerializationTests.swift
```diff
@@ -567,6 +592,67 @@ final class DecodableResponseSerializerTests: BaseTestCase {
...
+        }
+
+        static let emptyValue = DecodableClassWithEmptyResponse(string: "");
+
+        init(string: String) { self.string = string }
+
+        let string: String
+    }
+
+    func testThatJSONDecodableResponseSerializerSucceedsWhenDataIsNilWithEmptyResponseStatusCodeAndDecodableClassHasEmptyResponse() {
```
jshier: What is this testing? It seems unnecessary with all of the other tests around the feature.

Source/ResponseSerialization.swift
cs4alhaider: Agree with Jshier

Source/ResponseSerialization.swift
jvannoord: addressed in 4664bb578a70c3f8a1d3f74082b352d5eea8fd06

Source/ResponseSerialization.swift
jvannoord: addressed in 4664bb578a70c3f8a1d3f74082b352d5eea8fd06

Source/ResponseSerialization.swift
jvannoord: addressed in 4664bb578a70c3f8a1d3f74082b352d5eea8fd06

Tests/ResponseSerializationTests.swift
jvannoord: I was intending to specifically test for identity in case the case of a class being used. Removed in 53775470973f7dd57c2ea4a04e7cd2b8595fcfc8.

Source/ResponseSerialization.swift
jvannoord: addressed in 4664bb578a70c3f8a1d3f74082b352d5eea8fd06

Issue #2678, opened 2019-01-06
SSL Pinning without bundling cert files


tosbaha:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

I am trying to pin SSL certificate. Since I am using Let's Encrypt, I just want to pin public keys of Let'sEncryptAuthorityX3 and Let'sEncryptAuthorityX4 as a backup. I don't want to bundle cert, instead, I want to use public key with `customEvaluation` I don't want to bundle cert files because they can be replaced easily. 

## What did you expect to happen?

I want to extract public keys from the SecTrust but methods are private. I know I can copy and use those methods. However I think it would be nice to have those methods available. 

## What happened instead?

There is no easy way to extract public keys from certificate without duplicating Alamofire's private methods.

tosbaha:
I am trying to pin root lets encrypt certifcate. So I am trying to pin the public key of Let'sEncryptAuthorityX3 and Let'sEncryptAuthorityX4. For example according to my research here are the public keys

|NAME|Base64 Public Key|
|------|-------------------|
|DST Root CA X3| Vjs8r4z+80wjNcr1YKepWQboSIRi63WsWXhIMN+eWys=|
|Let's Encrypt Authority X3| YLh1dUR9y6Kja30RrAn7JKnbQG/uEtLMkBgFF2Fuihg=|
|Let's Encrypt Authority X4 (backup)|sRHdihwgkaib1P1gxX8HFszlD+7/gTfNvuAybgLPNis=|
|ISRG Root X1| C5+lpZ7tcVwmwQIMcRtPbsQtWLABXhQzejna0wHFr8M=|

I am not familiar with how these are generated but it looks like they can be extracted with `publicKeys` function in `ServerTrustPolicy.swift` in Alamofire 4. I guess they are sha256 of public keys but I am not so sure. As I said instead of embedding cert files in my bundle, I am looking for ways to just hardcode the public key of Let's encrypt certificate.

jshier:
Right. The hard part of that is going to be converting the Base64 `String` into a `SecKey` which you can use. Alamofire doesn't current do that. Extracting the `SecKey` from a `SecTrust` is relatively trivial in comparison, as it's just one more step than extracting the `SecKey` from a `SecCertificate`. We'd certainly be interested in such an extension if you get it working, so feel free to open a PR!

[...]

PR #2680, opened 2019-01-08
SPKI Support

tosbaha:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/2678

### Goals :soccer:
- Adds Subject Public Key Info Hash
- It is better to just include SPKI instead of including whole cert files for security and size wise.
- It is a standard. It can be used for [HPKP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Public_Key_Pinning)

- CommonCrypto added for SHA-256 Hash
- Since it is complicated to parse ASN1 header, I used the same methods used by [TrustKit](https://github.com/datatheorem/TrustKit/blob/master/TrustKit/Pinning/TSKSPKIHashCache.m)
- I basically ported TrustKit algorithm to Swift for SHA-256 Hash
- New error code is added for SPKI failure

### Testing Details :mag:
- Three tests are added for verifying SPKI hash of leaf certificate.

### DIFF
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -180,6 +180,8 @@ public enum AFError: Error {
         case certificatePinningFailed(host: String, trust: SecTrust, pinnedCertificates: [SecCertificate], serverCertificates: [SecCertificate])
         /// Public key pinning failed.
         case publicKeyPinningFailed(host: String, trust: SecTrust, pinnedKeys: [SecKey], serverKeys: [SecKey])
+        /// Subject PublicKey Info pinning failed
+        case subjectPublicKeyInfoPinningFailed(host: String, trust: SecTrust, pinnedKeys: [String], serverKeys:[SecKey])
     }
 
     case explicitlyCancelled
@@ -599,6 +601,8 @@ extension AFError.ServerTrustFailureReason {
             return "Certificate pinning failed for host \(host)."
         case let .publicKeyPinningFailed(host, _, _, _):
             return "Public key pinning failed for host \(host)."
+        case let .subjectPublicKeyInfoPinningFailed(host, _, _, _):
+            return "Subject Public key info pinning failed for host \(host)."
         }
     }
 }
--- a/Source/ServerTrustEvaluation.swift
+++ b/Source/ServerTrustEvaluation.swift
@@ -23,6 +23,7 @@
 //
 
 import Foundation
+import CommonCrypto
 
 /// Responsible for managing the mapping of `ServerTrustEvaluating` values to given hosts.
 open class ServerTrustManager {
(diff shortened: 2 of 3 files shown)


lexon:
Any updates here?

Source/AFError.swift
```diff
@@ -180,6 +180,8 @@ public enum AFError: Error {
         case certificatePinningFailed(host: String, trust: SecTrust, pinnedCertificates: [SecCertificate], serverCertificates: [SecCertificate])
         /// Public key pinning failed.
         case publicKeyPinningFailed(host: String, trust: SecTrust, pinnedKeys: [SecKey], serverKeys: [SecKey])
+        /// Subject PublicKey Info pinning failed
+        case subjectPublicKeyInfoPinningFailed(host: String, trust: SecTrust, pinnedKeys: [String], serverKeys:[SecKey])
```
scaio, line 184: nit: 
- Might want to rename `pinnedKeys` to `pinnedKeyHashes`
- Might want to expose serverKeyHashes as well to allow for easier debugging

Source/ServerTrustEvaluation.swift
```diff
@@ -326,6 +327,137 @@ public final class PublicKeysTrustEvaluator: ServerTrustEvaluating {
...
+            case .ecDsaSecp384r1:
+                return [
+                    0x30, 0x76, 0x30, 0x10, 0x06, 0x07, 0x2a, 0x86, 0x48, 0xce, 0x3d, 0x02,
+                    0x01, 0x06, 0x05, 0x2b, 0x81, 0x04, 0x00, 0x22, 0x03, 0x62, 0x00
+                ]
+            }
+        }
+    }
+
+    private func getAsn1HeaderBytes(publicKeyType:String,publicKeySize:Int) -> [UInt8]? {
```
scaio, line 393: Formatting is off here, need spaces after `publicKeyType:` and `publicKeySize:`

Source/ServerTrustEvaluation.swift
```diff
@@ -326,6 +327,137 @@ public final class PublicKeysTrustEvaluator: ServerTrustEvaluating {
...
+    ///                               Defaults to `true`.
+    public init(keys: [String],
+                performDefaultValidation: Bool = true,
+                validateHost: Bool = true) {
+        self.keys = keys
+        self.performDefaultValidation = performDefaultValidation
+        self.validateHost = validateHost
+    }
+
+    private enum Asn1Header {
```
scaio, line 360: Consider renaming this to `PublicKeyType`, renaming `bytes` to `headerBytes` and adding a `keySize: Int` var.

You can then have `PublicKeyType` be `RawRepresentable` by having the `init(rawValue: )` implementation check against the known `kSecAttrKeyType` values.

This would allow you to get rid of `getAsn1HeaderBytes` and have this instead in `hashSubjectPublicKeyInfoFromCertificate`:

```swift
       // ...
        guard let publicKeyAttributes = SecKeyCopyAttributes(publicKey) as? [CFString:Any],
            let rawPublicKeyType = publicKeyAttributes[kSecAttrKeyType] as? String,
            let publicKeyType = PublicKeyType(rawValue: rawPublicKeyType),
            let publicKeysize = publicKeyAttributes[kSecAttrKeySizeInBits] as? NSNumber,
            publicKeyType.keySize == publicKeysize
            else { return nil }
            // ...
```

Source/ServerTrustEvaluation.swift
```diff
@@ -326,6 +327,137 @@ public final class PublicKeysTrustEvaluator: ServerTrustEvaluating {
...
+            return Asn1Header.rsa4096.bytes
+        } else if publicKeyType == kSecAttrKeyTypeECSECPrimeRandom as String && publicKeySize == 256 {
+            return Asn1Header.ecDsaSecp256r1.bytes
+        } else if publicKeyType == kSecAttrKeyTypeECSECPrimeRandom as String && publicKeySize == 384 {
+            return Asn1Header.ecDsaSecp384r1.bytes
+        }
+        return nil // Unknown Algorithm
+    }
+
+    private func hashSubjectPublicKeyInfoFromCertificate(certificate:SecCertificate) -> String? {
```
scaio, line 406: nit: space needed after `certificate:` might also want to rename the function to `subjectPublicKeyHash(from certificate: SecCertificate)` so it reads more swifty

Source/ServerTrustEvaluation.swift
```diff
@@ -326,6 +327,137 @@ public final class PublicKeysTrustEvaluator: ServerTrustEvaluating {
...
+        let pinnedKeysInServerKeys: Bool = {
+            for certificate in trust.certificates {
+                for key in keys {
+                    if hashSubjectPublicKeyInfoFromCertificate(certificate: certificate) == key {
+                        return true
+                    }
+                }
+            }
+            return false
+        }()
```
scaio, line 450: Consider this instead:
```swift
    let serverKeyHashesSet = Set(trust.certificates.compactMap { subjectPublicKeyHash(from:  $0 })
    let pinnedKeysInServerKeys = !serverKeyHashesSet.intersection(Set(keys)).isEmpty
```

inepineee:
Nooo been waiting on this PR :'( Sad to read you aren't planning to do this feature

jshier:
If you need this functionality you should be able to pull it into your own codebase fairly easily. We are planning to implement this functionality, we just don't have a timeline for it.

PR #2687, opened 2019-01-10
Fixed the problem with Swift version detection

RomanPodymov:
### Goals :soccer:
This PR fixes the problem with Swift version detection when Alamofire is using via CocoaPods.

.swift_version file was deprecated in CocoaPods 1.4.0, we should use the `swift_version` attribute in podspec file to specify Swift version. Check [this link](https://blog.cocoapods.org/CocoaPods-1.4.0/) for more details about the Swift Version DSL in CocoaPods.

### Testing Details :mag:
No tests required, just deploy a new version via CocoaPods.


jshier:
@cnoon @RomanPodymov I've avoided the single Swift version attribute for CocoaPods as it conflicts with our ability to support multiple Swift versions through CP. Without the attribute, CP will use the integrating target's Swift version to set the frameworks. This works great in most cases, as Alamofire supports up-to-date versions of Swift. I've been waiting for multiple version support from CocoaPods which is slated for their 1.7 release and available on their `master`. However, their release process seems to have stalled and they haven't even shipped 1.6 yet. So I'm not sure what to do here. If we can only support one Swift version at a time, it seems like we'll need to make more frequent breaking releases. Of course, this says nothing of supporting not only Swift 5 but also that compiler's Swift 4 mode and the newly standardized `Result` type. At that point we'd likely want to support Swift 4.3(?) and 5 at the same time. 

So while this is useful, I don't think making this change buys us anything while limiting our options for future Swift version support through CP.

cnoon:
@jshier what about the case where the consuming project has explicitly specified Swift 4 or Swift 3? In those cases, wouldn't the lack of an explicit `swift_version` attribute in our podspec cause Alamofire not to build?

Also, isn't it possible to ship an app compiled against Swift 4.1 with linked frameworks compiled against Swift 4.2? If yes, and I'm pretty sure the answer is yes, wouldn't we want to explicitly move to 4.2 in our podspec to ensure AF compiles in all cases with CocoaPods?

RomanPodymov:
@jshier @cnoon 
If Alamofire can be built with any version of Swift, [we can do it like that](https://github.com/mxcl/PromiseKit/blob/master/.github/PromiseKit.podspec). But we should delete .swift_version file because it was deprecated.

RomanPodymov:
@jshier Can you guess why the `.swift_version` file is missing in [PromiseKit](https://github.com/mxcl/PromiseKit)? This library compiles with any Swift version supported by Xcode 10.1.

jshier:
@RomanPodymov Unfortunately Alamofire 5 does not, so without a `.swift-version` file, linting breaks. Even under a future version of Xcode that supports Swift 5 (which will only support 4.2 and 4 as well), CocoaPod's choice for a default language version (possibly 4, though it seems likely CP will have a bug and still try to set 3.2) will still likely fail to build. So we'll still need `.swift-version` to lint correctly, and we can't support a single version of Swift without limiting out users. Hopefully CP ships multi-version support before removing `.swift-version` support fully.

So thanks for the PR, and I appreciate the motivation, but it just doesn't work for us right now.

Issue #2677, opened 2019-01-04
When starting the app in airplane mode, the listener is not notified


alekop:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

```
let reachability = ReachabilityNetworkManager(host: "www.apple.com")!
reachability.listener = { state in
    print("isReachable: \(state)")
}
reachability.startListening()
```

## What did you expect to happen?

The `print` statement should have printed "isReachable: false"

## What happened instead?

Nothing is printed when starting with no connectivity. But if you start the app with connectivity, it will print. The initial notification is not called in airplane mode.

The issue is that `flags` and `previousFlags` both have a rawValue of 0, so the guard in notifyListener returns. The previous flags should probably default unknown, rather than 0, which means not-reachable.

## Alamofire Environment

**Alamofire version: 4.8.0**
**Xcode version: 10.1**
**Swift version: 4.2**
**Platform(s) running Alamofire: iOS**
**macOS version running Xcode: 10.13.6**

cnoon:
Hi @alekop,

Thank you for the report. Could you report which device and version of iOS you are running? Also, could you try the following code and see if it behaves better?

```swift
let reachability = ReachabilityNetworkManager()!
reachability.listener = { print("isReachable: \($0)") }
reachability.startListening()
```

> I tend to use the non-host specific initializer as the `host` version tends to be less reliable.

If you could report back on your findings I'd greatly appreciate it.

alekop:
Hi @cnoon,

I tried the no-hostname initializer, but it didn't make any difference. The issue seems to be here, in NetworkReachabilityManager.swift:

```
    listenerQueue.async {
        self.previousFlags = SCNetworkReachabilityFlags()
        self.notifyListener(self.flags ?? SCNetworkReachabilityFlags())
    }
```

Both `previousFlags`, and `flags` get initialized to the same value, which trips the guard in `notifyListener`:

```
    func notifyListener(_ flags: SCNetworkReachabilityFlags) {
        guard previousFlags != flags else { return }
        previousFlags = flags
        
        listener?(networkReachabilityStatusForFlags(flags))
    }
```

cnoon:
What versions of iOS have you tried this on @alekop? I'm guessing this is a change in functionality somewhere along the line.

alekop:
[...]

PR #2688, opened 2019-01-14
Bugfix - Network Reachability

cnoon:
### Issue Link :link:
This PR resolves Issue #2677.

### Goals :soccer:
The goals of this PR are to resolve the issue reported where the reachability listener closure is not being called when launching an app in airplane mode.

Prior to this change, the reachability listener was notified at launch if the device was not in airplane mode. When in airplane mode, the reachability flags were 0, which is what we were defaulting the previous flags to. Because of this, the `notifyListener` function would early out since the flags had not changed.

The core issue here was that the default value of `previousFlags` was poorly chosen. It should instead default to a value that represents and unknown status. One way to do this would be to make `previousFlags` an optional. However, this change would not be backwards compatible, so it was abandoned.

The best backwards compatible way to make this change was to set `previousFlags` to a `rawValue` that is not currently used. That way, we can guarantee that the originally computed flags will be different than the default previous flags. This change results in the listener being called at launch even when in airplane mode.

### Testing Details :mag:

All the original reachability tests are still passing as expected. Given the nature of this particular change, there's no way to test this specific change unless we would redesign the `NetworkReachabilityManager` to decouple the actual Reachability API calls and inject them. This type of change is out-of-scope for AF 4.x release train.


Example/Resources/Base.lproj/Main.storyboard
```diff
@@ -134,100 +139,121 @@
...
+                                                    <color key="textColor" red="0.66666668650000005" green="0.66666668650000005" blue="0.66666668650000005" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                                    <color key="highlightedColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                                </label>
+                                            </subviews>
+                                        </tableViewCellContentView>
+                                        <color key="backgroundColor" red="1" green="1" blue="1" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                                    </tableViewCell>
+                                </cells>
+                            </tableViewSection>
+                            <tableViewSection headerTitle="Reachability" id="7nc-cQ-nUY">
```
cnoon: Added a new `Reachability` section with a `Reachability Status` cell that prints out the reachability status when you tap it.

chrisballinger:
Wow good timing! I was actually looking into this yesterday because I was encountering the same issue, I'll take a look and see if your patch fixes it for us.

alekop:
Great! Thanks for the quick turnaround, @cnoon!

jshier:
:+1:

PR #2691, opened 2019-01-15
Simplify and refactor

jdisho:
This is my first PR. First of all, I want to thank you for your time investment in `Alamofire`. Amazing lib 👏
### Goals :soccer:
Simplifying and refactoring some basic things.

- Most of which are simplification of `switch` statements. The `switch` cases which are simplified are 2 case statements and can we simplified by using `guard`.
- I set the default value of queue to` .main` rather than `nil`. I did this because under the hood, `.main` is used when the queue value is `nil`.
- I remove `methodDependent` property in `ParameterEncoding` since it is a duplication of `default`. (If there is a reason why `methodDependent` and `default` are included, I would be interested to know why 😄)
- I simplified the `encodesParametersInURL` in `ParameterEncoding` by making it part of `Destination`. (I occurred that in the same way `encodesParametersInURL` is implemented in `ParameterEncoder`)

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -108,11 +108,8 @@ public struct URLEncoding: ParameterEncoding {
 
     // MARK: Properties
 
-    /// Returns a default `URLEncoding` instance.
+    /// Returns a default `URLEncoding` instance with a `.methodDependent` destination.
     public static var `default`: URLEncoding { return URLEncoding() }
-
-    /// Returns a `URLEncoding` instance with a `.methodDependent` destination.
-    public static var methodDependent: URLEncoding { return URLEncoding() }
 
     /// Returns a `URLEncoding` instance with a `.queryString` destination.
     public static var queryString: URLEncoding { return URLEncoding(destination: .queryString) }
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -101,7 +101,7 @@ public struct URLEncoding: ParameterEncoding {
             case .numeric:
                 return value ? "1" : "0"
             case .literal:
-                return value ? "true" : "false"
+                return value.description
             }
         }
     }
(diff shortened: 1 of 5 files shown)


Source/ResponseSerialization.swift
```diff
@@ -109,7 +109,7 @@ extension DataRequest {
     ///   - completionHandler: The code to be executed once the request has finished.
     /// - Returns:             The request.
     @discardableResult
-    public func response(queue: DispatchQueue? = nil, completionHandler: @escaping (DataResponse<Data?>) -> Void) -> Self {
+    public func response(queue: DispatchQueue = .main, completionHandler: @escaping (DataResponse<Data?>) -> Void) -> Self {
```
jshier: If we're going to make this change, it would be good to audit all API that takes optional values with a coalesced value in the body, and change them all to work like this.

Source/ParameterEncoding.swift
```diff
@@ -101,19 +109,16 @@ public struct URLEncoding: ParameterEncoding {
...
-                return value ? "true" : "false"
+                return value.description
             }
         }
     }
 
     // MARK: Properties
 
-    /// Returns a default `URLEncoding` instance.
+    /// Returns a default `URLEncoding` instance with a `.methodDependent` destination.
```
jshier: I'm not sure about this one. Perhaps keeping `default` and adding an `httpBody` convenience property?

jdisho:
Thank you for your response @jshier. I will try to address your comments as soon as possible.

Source/ParameterEncoding.swift
jdisho: I am not understanding your comment. I need more context. What is the issue about?

cnoon:
Thanks for putting this all together @jdisho. I'm going to take a stab at putting together a different PR with most of these changes while keeping your attribution. I'll link the new PR once it's up.

cnoon:
Hi @jdisho,

I just pushed up #2765 which pulls in these changes, culls out a few of them, and rebases on top of the #2716 PR. I made sure to maintain your attribution to the changeset. I'm going to go ahead and close out this PR for now. Please redirect all further comments to #2765. These changes will go out as part of AF5 beta 4.

Cheers. 🍻

PR #2693, opened 2019-01-16
Feature - Interceptors Per Request

cnoon:
This PR adds a new `RequestInterceptor` protocol that allows you to provide a custom adapter and retrier per request instead of only per session. We've been using per request interceptors in an AF fork at Nike for over a year. They have allowed us to move from many sessions each paired with a single interceptor to a single session where each request is paired with the appropriate interceptor. This enables us to use a single session with infinite numbers of authentication systems.

### Goals :soccer:

The goal of this PR is to allow clients to provide an adapter and retrier (or interceptor) per request in addition to per session. This change is meant to fix a design flaw in the original adapter and retrier architecture. Currently, in AF 4.x, you can only provide a single adapter and retrier per session. While this is a really useful feature, it is very limiting. For example, you may be making requests to several different services that use very different authentication mechanisms. Ideally, you could use a single session with different authentication mechanisms for the different requests. The current design makes this very difficult.

If you are only allowed a single adapter/retrier for your session, then you need to do introspection on the url request to see which authentication system it should be using. Generally, you could instead the domain of the URL and be able to choose the appropriate auth system. However, sometimes you have situations where different auth types are used for the same domain. Different relative paths in the URL will have different authentication systems. We currently face this challenge in our services at Nike. We have several different authentication systems in front of different api.nike.com services.

Therefore, trying to have a single interceptor figure out which auth system pairs to the request is not feasible. We'd have to maintain a map of all the URLs to the auth systems to power a single interceptor. When you have hundreds of services you are potentially calling, it quickly becomes obvious that this is the wrong solution.

Instead, you want to be able to specify an interceptor per request in these situations.

There are a few ways that this feature could be implemented:

1. Add `interceptor` parameter to top-level APIs
1. Leave top-level APIs alone and extend `URLRequestConvertible` to include `interceptor` property
1. Add new `InterceptableRequest` protocol that extends `URLRequestConvertible` that we’d check in implementation

I chose to go with the first option because the other two are clunky. It doesn't make sense to extend `URLRequestConvertible` to support interceptors. It also doesn't make sense to add another protocol that only extends `URLRequestConvertible` to tack on interceptors. It would make more sense to have a protocol that would encompass more pieces of info such as baseURL, relative path, query parameters, headers, body, interceptor, etc.

### Testing Details :mag:

The tests that have been added test the per request interceptor logic at the `Session` level.


Source/RequestInterceptor.swift
```diff
@@ -42,3 +54,8 @@ public protocol RequestRetrier {
     /// - parameter completion: The completion closure to be executed when retry decision has been determined.
     func should(_ manager: Session, retry request: Request, with error: Error, completion: @escaping RequestRetryCompletion)
 }
+
+// MARK: -
+
+/// A type that intercepts requests to potentially adapt and retry them.
+public protocol RequestInterceptor: RequestAdapter, RequestRetrier {}
```
cnoon: I thought it made the most sense to combine the `RequestAdapter`, `RequestRetrier`, and `RequestInterceptor` into a single file.

Source/Session.swift
```diff
@@ -174,37 +178,41 @@ open class Session {
...
+        return download(convertible, interceptor: interceptor, to: destination)
     }
 
     open func download<Parameters: Encodable>(_ convertible: URLConvertible,
                                               method: HTTPMethod = .get,
                                               parameters: Parameters? = nil,
                                               encoder: ParameterEncoder = JSONParameterEncoder.default,
                                               headers: HTTPHeaders? = nil,
+                                              interceptor: RequestInterceptor? = nil,
                                               to destination: DownloadRequest.Destination? = nil) -> DownloadRequest {
```
cnoon: I really struggled with the external API here. Given that all the other parameters are stacked in front of the destination with defaults, I didn't think it made sense to move the `interceptor` after the `destination`, even if it reads better. VERY open to suggestions...

Source/Session.swift
```diff
@@ -174,37 +178,41 @@ open class Session {
...
                                                       encoder: encoder,
                                                       headers: headers)
 
-        return download(convertible, to: destination)
+        return download(convertible, interceptor: interceptor, to: destination)
     }
 
     open func download(_ convertible: URLRequestConvertible,
+                       interceptor: RequestInterceptor? = nil,
                        to destination: DownloadRequest.Destination? = nil) -> DownloadRequest {
```
cnoon: Same challenge here with the external API. I chose to put the `interceptor` in front of the `destination` to be consistent with the previous API. Again, very open to suggestions.

Source/Session.swift
```diff
@@ -255,81 +265,98 @@ open class Session {
...
 
-        return upload(data, with: convertible)
+        return upload(data, with: convertible, interceptor: interceptor)
     }
 
-    open func upload(_ data: Data, with convertible: URLRequestConvertible) -> UploadRequest {
-        return upload(.data(data), with: convertible)
+    open func upload(_ data: Data,
+                     with convertible: URLRequestConvertible,
+                     interceptor: RequestInterceptor? = nil) -> UploadRequest {
```
cnoon: More struggling here with the external API. Some methods are more verbose than others, so trying to make some read nicely and others not is challenging.

Source/Session.swift
```diff
@@ -373,6 +400,8 @@ open class Session {
 
     func perform(_ request: DownloadRequest) {
         requestQueue.async {
+            guard !request.isCancelled else { return }
```
cnoon: @jshier please take a close look at this change. We were previously missing a cancellation check when performing download requests. I think we probably want one which is why I added it, but I wanted to make sure that that wasn't intentionally left out for some reason.

Source/Session.swift
```diff
@@ -442,6 +471,16 @@ open class Session {
...
+
+    // MARK: - Adapters and Retriers
+
+    func adapter(for request: Request) -> RequestAdapter? {
+        return request.interceptor ?? adapter
+    }
+
+    func retrier(for request: Request) -> RequestRetrier? {
+        return request.interceptor ?? retrier
+    }
```
cnoon: Nice and simple. Use the request's interceptor if specified, and fall back to the session if necessary.

Source/Session.swift
jshier: I agree, this spot makes sense.

Source/Session.swift
jshier: Swift guidelines say that defaulted arguments should come after non-defaulted ones, so this looks good.

Source/Session.swift
jshier: Looks good. Cancellation checks are hard to reason about consistently, so I usually just put them before major actions. As long as this location is consistent with the others (it seems to be) so we don't have behavioral differences between request types, this should be fine.

jshier:
This is great! I'm glad the internal rewrite is paying off like this. So I'm good with the feature as implemented, but I'd like to hold off merging until we can discuss #2694, as we may want to do some additional renaming or other refactors to differentiate from or merge with `RetryPolicy`.

Source/RequestInterceptor.swift
```diff
@@ -24,6 +24,18 @@
...
 import Foundation
 
+/// A type that can inspect and optionally adapt a `URLRequest` in some manner if necessary.
+public protocol RequestAdapter {
+    /// Inspects and adapts the specified `URLRequest` in some manner and calls the completion handler with the Result.
+    ///
+    /// - Parameters:
+    ///   - urlRequest: The `URLRequest` to adapt.
+    ///   - completion: The completion handler that must be called when adaptation is complete.
+    func adapt(_ urlRequest: URLRequest, completion: @escaping (_ result: Result<URLRequest>) -> Void)
```
AtomicCat: Things I wondered while integrating this branch into my code and updating my adaptor unit tests...

Is there any expectation that the completion should be called asynchronously, or can it be called inline? 
If called asynchronously, is there a preferred queue? Should the adapter be passed a queue on which the completion should be called?

jshier: No, it's specially designed for the queue not to matter. Alamofire immediately enqueues its work back onto the internal `rootQueue`.

AtomicCat: Maybe the docstrings should note this?

jshier: Sure. How about "The completion handler that must be called when adaptation is complete. It may be called from any queue."

cnoon:
Closing this PR out since it was replaced by #2704 and merged into `master`.

PR #2696, opened 2019-01-18
Use the hash(into:) API for hashing.

SlaunchaMan:
### Goals :soccer:
This PR fixes a warning in Xcode by using the new `hash(into:)` API for `Hashable` on the `Request` type.

Instead of implementing `hashValue`, which has been deprecated, we use `hash(into:)` to hash into a `Hasher`. The value of `hashValue` will then come from the `Hasher`.

### Testing Details :mag:
This shouldn’t affect any tests, as none were dependent on `Request` having a deterministic `hashValue`.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -497,8 +497,8 @@ extension Request: Equatable {
 }
 
 extension Request: Hashable {
-    public var hashValue: Int {
-        return id.hashValue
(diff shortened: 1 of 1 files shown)


SlaunchaMan:
@cnoon Upon more reading of the `Hashable` documentation, it looks like using `combine(_:)` here is the preferred method.

PR #2697, opened 2019-01-18
fixed syntax for swift build command

toddlahtinen:
just a small readme typo fix


jshier:
Fixed in #2729, but I'll give you credit too. Sorry about the delay!

PR #2699, opened 2019-01-23
Feature - Redirect Handlers

cnoon:
This PR adds a `RedirectHandler` protocol which is leveraged as a `redirectHandler` property on both the `Request` and `Session` APIs. It allows clients full control over redirect behavior at the `Session` level, as well as at the `Request` level through a chainable `redirect` API.

> This is an alternative approach to #2563.

### Goals :soccer:

The goals of this PR are to:

- Add support to AF5 for customizing redirect behavior at the `Session` level to match AF4.x functionality
- Add support to AF5 for customizing redirect behavior per `Request` to make it easier to implement simple redirect logic for a single request

Currently, in AF 4.x, you can control redirect behavior only at the `SessionDelegate` level by either subclassing or setting the `taskWillPerformHTTPRedirectionWithCompletion` closure property. While this gives you the ability to control redirect behavior, it's pretty clunky. The changes in AF5 have cleaned this up a bunch, but do not currently expose the ability to control redirect behavior.

> You can currently subclass `SessionDelegate`, but the properties needed to customize the functionality are `internal` or `private`.

#### RedirectHandler Protocol

In this PR, I added a `RedirectHandler` protocol which contains a single API for determining what to do with a redirect response.

```swift
public protocol RedirectHandler {
    func task(_ task: URLSessionTask,
              willBeRedirectedTo request: URLRequest,
              for response: HTTPURLResponse,
              completion: @escaping (URLRequest?) -> Void)
}
```

This protocol is then added to a `Session` through the initializer and made available through a `redirectHandler` property. It can also be added to a `Request` through the new `redirect` chainable API. The nice thing about having the logic abstracted into a protocol is that it's much easier to share the functionality between a `Session` and `Request` instead of having custom ways of handling redirects between the two.

#### Redirect Control per Request

This is a really cool addition to Alamofire. Right now, you have no way to control redirect behavior per request. It's only configuring at the `SessionManager` level (or per session). Therefore, you need to use request introspection on the `URLSessionTask`'s originalRequest and hope that you can dig the info you need out to determine the redirect behavior.

For example, you may need to provide custom redirect behavior for a bunch of different endpoints. In order to implement this, your closure has to keep a custom map of the different endpoints and their matching redirect behaviors. They also might all redirect differently. Then you may also have different behaviors with different test subdomains, etc. It can quickly grow out of control.

Instead, in most cases, you'd rather implement the custom redirect logic at the time of creation of the request. That makes it MUCH easier to simply pair up the custom redirect logic with the actual request. Then you no longer need to maintain a map and perform all the introspection at the session level.

#### Redirect Closure on `Request`

I purposefully left out support in this PR for `redirect` support through a closure API. While it would be a nice addition, it creates an odd mismatch between `Request` and `Session`. We wouldn't want to support both the protocol and closure variants in the `Session`, but we probably would in the `Request`. Therefore, it seemed odd to add the closure variant only to the `Request`. If we did decide to add it, it would be something like this:

```swift
extension Request {
    public typealias RedirectClosure = (URLSessionTask, URLRequest, HTTPURLResponse, @escaping (URLRequest?) -> Void) -> Void

    @discardableResult
    open func redirect(closure: RedirectClosure) -> Self {
        protectedMutableState.write { $0.redirectClosure = closure }
        return self
    }
}
```

Then we'd have to add another fallback case in the `SessionDelegate` implementation to first grab the `redirectHandler`, then the `redirectClosure`, and then the `redirectHandler` on the `Session`. We could also only store the `redirectClosure` on the `Request` and grab the function pointer to the `RedirectHandler` API and set it as the `redirectClosure`. Then you'd have to grab the `redirectClosure` for a `Request` and `redirectHandler` for a `Session`, each of which are called differently in the `SessionDelegate`.

> Again, I chose to avoid this complexity for redirects by not providing a closure API for redirects. We could certainly add it though if everyone feels the convenience outweighs the implementation complexity.

### Testing Details :mag:

I have intentionally left tests out of this PR until we agree that this is the direction we'd like to take for this particular feature. Once the approach is locked, I'll add the appropriate tests.

**Update**

Tests have been added for per request and per session redirect handlers as well as prioritization of request over session.

### DIFF
--- a/Source/RedirectHandler.swift
+++ b/Source/RedirectHandler.swift
@@ -1,0 +1,46 @@
+//
+//  RedirectHandler.swift
+//
+//  Copyright (c) 2014-2018 Alamofire Software Foundation (http://alamofire.org/)
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+/// A type that handles how an HTTP redirect response from a remote server should be redirected to the new request.
+public protocol RedirectHandler {
+    /// Determines how the HTTP redirect response should be redirected to the new request.
+    ///
+    /// The `completion` closure should be passed one of three possible options:
+    ///
+    ///   1. The new request specified by the redirect (this is the most common use case).
+    ///   2. A modified version of the new request (you may want to route it somewhere else).
+    ///   3. A `nil` value to deny the redirect request and return the body of the redirect response.
+    ///
+    /// - Parameters:
+    ///   - task:       The task whose request resulted in a redirect.
+    ///   - request:    The URL request object to the new location specified by the redirect response.
+    ///   - response:   The response containing the server's response to the original request.
+    ///   - completion: The closure to execute containing the new request, a modified request, or `nil`.
+    func task(_ task: URLSessionTask,
+              willBeRedirectedTo request: URLRequest,
+              for response: HTTPURLResponse,
+              completion: @escaping (URLRequest?) -> Void)
+}
--- a/Source/RedirectHandler.swift
+++ b/Source/RedirectHandler.swift
@@ -44,3 +44,53 @@ public protocol RedirectHandler {
               for response: HTTPURLResponse,
               completion: @escaping (URLRequest?) -> Void)
 }
+
+// MARK: -
+
+/// `Redirector` is a convenience `RedirectHandler` making it easy to follow, not follow, or modify a redirect.
+public struct Redirector {
+    /// Defines the behavior of the `Redirector` type.
+    ///
+    /// - follow:      Follows the redirect as defined in the response.
+    /// - doNotFollow: Does not follow the redirect defined in the response.
+    /// - modify:      Modifies the redirect request defined in the response.
+    public enum Behavior {
+        case follow
+        case doNotFollow
+        case modify((URLSessionTask, URLRequest, HTTPURLResponse) -> URLRequest?)
+    }
+
+    /// Returns a `Redirector` with a follow `Behavior`.
+    public static let follow = Redirector(behavior: .follow)
+    /// Returns a `Redirector` with a do not follow `Behavior`.
+    public static let doNotFollow = Redirector(behavior: .doNotFollow)
+
+    /// The `Behavior` of the `Redirector`.
+    public let behavior: Behavior
+
+    /// Creates a `Redirector` instance from the `Behavior`.
+    ///
+    /// - Parameter behavior: The `Behavior`.
+    public init(behavior: Behavior) {
+        self.behavior = behavior
+    }
+}
+
+// MARK: -
+
+extension Redirector: RedirectHandler {
+    public func task(_ task: URLSessionTask,
+              willBeRedirectedTo request: URLRequest,
+              for response: HTTPURLResponse,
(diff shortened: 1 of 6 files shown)


cnoon:
Added a new `Redirector` type (best name I could come up with) in d4aabf6d that conforms to `RedirectHandler` and offers some convenience behaviors as well as the ability to control redirects through a closure API.

```swift
public struct Redirector {
    public enum Behavior {
        case follow
        case doNotFollow
        case modify((URLSessionTask, URLRequest, HTTPURLResponse) -> URLRequest?)
    }

    public let behavior: Behavior

    public init(behavior: Behavior) {
        self.behavior = behavior
    }
}

extension Redirector: RedirectHandler {
    public func task(_ task: URLSessionTask,
              willBeRedirectedTo request: URLRequest,
              for response: HTTPURLResponse,
              completion: @escaping (URLRequest?) -> Void) {
        switch behavior {
        case .follow:
            completion(request)
        case .doNotFollow:
            completion(nil)
        case .modify(let closure):
            let request = closure(task, request, response)
            completion(request)
        }
    }
}
```

I think this will make it easier to clients to quickly disable redirects for a single request or modify the redirect using the closure API.

Another option here would be to remove the behavior concept and simply support a closure API. I thought the extra convenience though would be nice in certain situations. Given that `RedirectHandler` usage is supported per request, I think the most common use case is probably that you want to disable redirects on certain requests or perhaps the entire session, while maintaining them for all other requests. Therefore, I think the `.doNotFollow` behavior will "probably" be the most commonly used.

> This is certainly speculation, but thinking back to the use cases I've run into, I've only ever wanted to disable all redirects on a session, or for a request or two.

Source/RedirectHandler.swift
```diff
@@ -0,0 +1,79 @@
...
+// MARK: -
+
+public struct Redirector {
+    public enum Behavior {
+        case follow
+        case doNotFollow
+        case modify((URLSessionTask, URLRequest, HTTPURLResponse) -> URLRequest?)
+    }
+
+    public let behavior: Behavior
```
jshier: Similar to the caching behavior, we could offer static properties with prebuilt `Behavior` instances.

Source/RedirectHandler.swift
```diff
@@ -0,0 +1,79 @@
...
+    ///   1. The new request specified by the redirect (this is the most common use case).
+    ///   2. A modified version of the new request (you may want to route it somewhere else).
+    ///   3. A `nil` value to deny the redirect request and return the body of the redirect response.
+    ///
+    /// - Parameters:
+    ///   - task:       The task whose request resulted in a redirect.
+    ///   - request:    The URL request object to the new location specified by the redirect response.
+    ///   - response:   The response containing the server's response to the original request.
+    ///   - completion: The closure to execute containing the new request, a modified request, or `nil`.
+    func task(_ task: URLSessionTask,
```
jshier: Might it be useful to make this function `throws`? It would allow anyone needing error handling to get it while placing no burden on those who don't. It shouldn't make the implementation much more difficult either.

Source/Request.swift
```diff
@@ -80,6 +80,8 @@ open class Request {
         var uploadProgressHandler: (handler: ProgressHandler, queue: DispatchQueue)?
         /// `ProgressHandler` and `DispatchQueue` provided for download progress callbacks.
         var downloadProgressHandler: (handler: ProgressHandler, queue: DispatchQueue)?
+        /// `RetryHandler` provided for redirect responses.
+        var redirectHandler: RedirectHandler?
```
jshier: Per our previous conversations, do we want this to be a single optional instance or an array, to allow for easier compensation of multiple handlers?

Source/Request.swift
```diff
@@ -486,6 +494,18 @@ open class Request {
 
         return self
     }
+
+    // MARK: - Redirects
+
+    /// Sets the redirect handler for the `Request` which will be used if a redirect response is encountered.
+    ///
+    /// - Parameter handler: The redirect handler.
```
jshier: As with some of my other comments, I've been using the type for most parameter docs, so "The `RedirectHandler`." would be my doc here. 

Also, lining up all of the "The"s to match my previous alignment. Or we can do a documentation pass later.

Source/Session.swift
```diff
@@ -34,6 +34,7 @@ open class Session {
     public let adapter: RequestAdapter?
     public let retrier: RequestRetrier?
     public let serverTrustManager: ServerTrustManager?
+    public let redirectHandler: RedirectHandler?
```
jshier: Similar to the `Request` side, do we want the optional property or an array?

Source/RedirectHandler.swift
jshier: Upon further discussion, since `URLSession` doesn't allow errors here, there doesn't seem to be a clean way for us to either.

Source/RedirectHandler.swift
cnoon: Added in db4b3854. I also considered adding them to a protocol extension on `RedirectHandler` to make it even more convenient, but I thought that might be a bit much.

Source/Request.swift
cnoon: After discussion in Slack, we've decide that these should stay a single instance and use a `precondition` in the chained API to prevent clients from calling it twice.

Source/Request.swift
cnoon: Updated in dd39e600, and yes, we'll catch all this prior to release.

Source/Session.swift
cnoon: Based on Slack conversation, we've decided that this would overcomplicate this feature with little upside and would most likely lead to confusion on the client side.

Source/RedirectHandler.swift
MarcSteven: Yeah sometimes  we must deal with the error  code here firstly, so  agree your view here!

cnoon:
All comments have been addressed @jshier, should be all good to go!

PR #2700, opened 2019-01-23
Add info for debugDescription

rain2540:
add parameters info for debugDescription of DataResponse and DownloadResponse

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -88,7 +88,7 @@ extension DataResponse: CustomStringConvertible, CustomDebugStringConvertible {
     /// result.
     public var debugDescription: String {
         let requestDescription = request.map { "\($0.httpMethod!) \($0)" } ?? "nil"
-        let body = request?.httpBody.map { String(decoding: $0, as: UTF8.self) } ?? "None"
+        let requestBody = request?.httpBody.map { String(decoding: $0, as: UTF8.self) } ?? "None"
         let responseDescription = response.map { (response) in
             let sortedHeaders = response.httpHeaders.sorted()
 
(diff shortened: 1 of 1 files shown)


Source/Response.swift
```diff
@@ -88,6 +88,12 @@ extension DataResponse: CustomStringConvertible, CustomDebugStringConvertible {
     /// result.
     public var debugDescription: String {
         let requestDescription = request.map { "\($0.httpMethod!) \($0)" } ?? "nil"
+        let parameters = { () -> String in
```
jshier: You can greatly compact this code, and the naming should reflect the fact that it's the body parameters:
```swift
let body = request?.httpBody.map { String(decoding: $0, as: UTF8.self) } ?? "None"
...
[Request Body]: \(body)
```

Source/Response.swift
```diff
@@ -101,6 +107,7 @@ extension DataResponse: CustomStringConvertible, CustomDebugStringConvertible {
 
         return """
         [Request]: \(requestDescription)
+        [Parameters]: \(parameters)
         [Response]: \n\(responseDescription)
         [Data]: \(data?.description ?? "None")
```
jshier: Since you're adding the "Request Body", you should do the same thing here: generate a `String` from the response `Data` and update this field to be `[Response Body]`.

Source/Response.swift
```diff
@@ -272,6 +279,12 @@ extension DownloadResponse: CustomStringConvertible, CustomDebugStringConvertibl
     /// actions, and the response serialization result.
     public var debugDescription: String {
         let requestDescription = request.map { "\($0.httpMethod!) \($0)" } ?? "nil"
+        let parameters = { () -> String in
```
jshier: Same changes as mentioned earlier.

Source/Response.swift
rain2540: Thank you for your suggestion

Source/Response.swift
rain2540: I have added [Response Body]

PR #2701, opened 2019-01-23
Feature - Cached Responses

cnoon:
This PR adds a `CachedResponseHandler` protocol which is leveraged as a `cachedResponseHandler` property on both the `Request` and `Session` APIs. It allows clients full control over redirect behavior at the `Session` level, as well as at the `Request` level through a chainable `cachedResponse(with:)` API.

### Goals :soccer:

The goals of this PR are to:

- Add support to AF5 for customizing cached response behavior at the `Session` level to match AF4.x functionality
- Add support to AF5 for customizing cached response behavior per `Request` to make it easier to implement simple caching behaviors for a single request (most commonly to disable caching for a single request)

Currently, in AF 4.x, you can control response caching behavior only at the `SessionDelegate` level by either subclassing or setting the `dataTaskWillCacheResponse` closure property. While this gives you the ability to control response caching behavior, it's pretty clunky. The changes in AF5 have cleaned this up a bunch, but do not currently expose the ability to control response caching behavior.

> You can currently subclass `SessionDelegate`, but the properties needed to customize the functionality are `internal` or `private`.

#### `CachedResponseHandler` Protocol

In this PR, I added a `CachedResponseHandler` protocol which contains a single API for determining what to do with a redirect response.

```swift
public protocol CachedResponseHandler {
    func dataTask(_ task: URLSessionDataTask,
                  willCacheResponse response: CachedURLResponse,
                  completion: @escaping (CachedURLResponse?) -> Void)
}
```

This protocol is then added to a `Session` through the initializer and made available through a `cachedResponseHandler` property. It can also be added to a `Request` through the new `cacheResponse(with:)` chainable API. The nice thing about having the logic abstracted into a protocol is that it's much easier to share the functionality between a `Session` and `Request` instead of having custom ways of handling response caching between the two.

#### `ResponseCacher` Struct

In order to make it even easier to handle response caching, I added a `ResponseCacher` struct which conforms to `CachedResponseHandler` (if you have any better naming suggestions, I'm all ears). It makes it very easy to either cache or not cache the response, and also offers a closure behavior for fully customizing the cached response before saving it to the `URLCache`.

```swift
public struct ResponseCacher {
    public enum Behavior {
        case cache
        case doNotCache
        case modify((URLSessionDataTask, CachedURLResponse) -> CachedURLResponse?)
    }

    public let behavior: Behavior

    public init(behavior: Behavior) {
        self.behavior = behavior
    }
}
```

> This type is super simple and is merely provided for convenience. Clients can choose to use the `ResponseCacher` convenience type, or create their own type that conforms to `CachedResponseHandler`.

#### Response Caching Control per Request

This is a nice addition to Alamofire. Right now, you have no way to control response caching behavior per request. It's only configurable at the `SessionManager` level (or per session). Therefore, you need to use request introspection on the `URLSessionDataTask`'s originalRequest and hope that you can dig the info you need out to determine the response caching behavior.

> The reasoning is exactly the same as detailed in #2699 for redirects. Per request control can be very helpful in certain network architectures.

### Testing Details :mag:

I have intentionally left tests out of this PR until we agree that this is the direction we'd like to take for this particular feature. Once the approach is locked, I'll add the appropriate tests.

**Update**

Tests have been added for per request and per session cached response handlers as well as prioritization of request over session.

### DIFF
--- a/Source/CachedResponseHandler.swift
+++ b/Source/CachedResponseHandler.swift
@@ -1,0 +1,92 @@
+//
+//  CachedResponseHandler.swift
+//
+//  Copyright (c) 2019 Alamofire Software Foundation (http://alamofire.org/)
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+/// A type that handles whether the data task should store the HTTP response in the cache.
+public protocol CachedResponseHandler {
+    /// Determines whether the HTTP response should be stored in the cache.
+    ///
+    /// The `completion` closure should be passed one of three possible options:
+    ///
+    ///   1. The cached response provided by the server (this is the most common use case).
+    ///   2. A modified version of the cached response (you may want to modify it in some way before caching).
+    ///   3. A `nil` value to prevent the cached response from being stored in the cache.
+    ///
+    /// - Parameters:
+    ///   - task: The data task whose request resulted in the cached response.
+    ///   - response: The cached response to potentially store in the cache.
+    ///   - completion: The closure to execute containing cached response, a modified response, or `nil`.
+    func dataTask(_ task: URLSessionDataTask,
+                  willCacheResponse response: CachedURLResponse,
+                  completion: @escaping (CachedURLResponse?) -> Void)
+}
+
+// MARK: -
+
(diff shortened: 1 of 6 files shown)


Source/CachedResponseHandler.swift
```diff
@@ -0,0 +1,63 @@
...
+
+public protocol CachedResponseHandler {
+    func dataTask(_ task: URLSessionDataTask,
+                  willCacheResponse response: CachedURLResponse,
+                  completion: @escaping (CachedURLResponse?) -> Void)
+}
+
+// MARK: -
+
+public struct ResponseCacher {
```
jshier: I think some static properties could help users. 
```swift
static let cache = ResponseCacher(behavior: .cache)
```
etc.

Source/Request.swift
```diff
@@ -506,6 +515,14 @@ open class Request {
         protectedMutableState.write { $0.redirectHandler = handler }
         return self
     }
+
+    // MARK: - Cached Responses
+
+    @discardableResult
+    open func cacheResponse(with handler: CachedResponseHandler) -> Self {
```
jshier: I prefer `using` as the modifier for methods like this. e.g. `.cacheResponse(using: myHandler)`.

Source/CachedResponseHandler.swift
cnoon: Definitely, good suggestion.

cnoon: Added in e091feaa.

Source/Request.swift
cnoon: Updated in efa2a1db.

cnoon:
All comments have been addressed and tests have been added. Should be all good to go here @jshier. Let's just make sure we get the PR repointed at `master` before merging.

Tests/CachedResponseHandlerTests.swift
```diff
@@ -0,0 +1,226 @@
...
+                    storagePolicy: .allowed
+                )
+            }
+        )
+
+        let request = session.request(urlString).cacheResponse(using: cacher).response { resp in
+            response = resp
+            expectation.fulfill()
+        }
+
```
agnosticdev: I think this PR for Cached Responses looks great.  This will be an excellent addition.  One question I had about the `.modify` case above is that it looks like it allows you to cache a response that was not sent back from the server.  What would be the use case for this?   Would this be fine grained control for the client to reason about the response and then delegate the cache that is saved?

cnoon: Well, that's not really the use case for `.modify`. The `.cache` and `.doNotCache` behaviors are simply convenience cases that pass the cached response or `nil` on to the completion handler automatically. The `.modify` case allows you to do three different things, with full access to all the data that the URL session delegate API provides: 1) cache the response as is, 2) do not cache the response by passing `nil` to the completion, 3) modify the cached response in some way (maybe adjusting the cache-control header) and passing that on to be cached.

PR #2705, opened 2019-01-31
Allow changing boundary string in upload:multipartFormData callback

ondrejstocek:
### Goals :soccer:
Allow changing boundary string in upload:multipartFormData callback

Changes MultipartFormData::boundary definition from constant `let` to `var`

### Testing Details :mag:
No test added

### DIFF
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -98,7 +98,7 @@ open class MultipartFormData {
     public var contentLength: UInt64 { return bodyParts.reduce(0) { $0 + $1.bodyContentLength } }
 
     /// The boundary used to separate the body parts in the encoded form data.
-    public let boundary: String
+    public var boundary: String
 
     private let fileManager: FileManager
     private var bodyParts: [BodyPart]


jshier:
Thanks for the PR! Can you explain why you want this change? Additionally, since this would be newly mutable, we'd want some relevant tests over the functionality.

cnoon:
Hi @ondrejstocek,

I'm going to go ahead and close this PR out for now since we haven't heard back. We don't want to keep these PRs open forever if we can help it. If you comment back on why this should be re-opened at some point, we can get it opened back up and pushed to `master` if it makes sense to do so.

Cheers. 🍻

ondrejstocek:
Hello @cnoon,
some servers, I have to work with, don’t have multipart upload implemented correctly, and search just for 4-dash prefix instead of whole boundary string. It is possible because ALL main browsers generate boundary with at least two dashes at beginning. I can’t have any chance to change this behaviour on server side and it is literally NO chance to override this string without changing significant part of Alamofire source code. This minor change alows to change the boundary string in uploading initial callback.

cnoon:
Thank you for that explanation @ondrejstocek, that does indeed make sense. I'm going to re-open this PR and work to get it through today.

cnoon:
Okay @ondrejstocek, @jshier and I decided that we're going to push this change out in AF 4.x (will ship in 4.8.2), and that we're going to re-think the upload APIs around `MultipartFormData` to better support this use case. I just pushed up your change in 0f2135b0 to `hotfix`, and we'll get it released shortly.

I'm going to go ahead and close this PR out for now, and I'll link the new PR in AF5 that addresses this issue just for visibility.

Cheers. 🍻

cnoon:
Hi @ondrejstocek, I just pushed up #2764 which is a better long-term solution to this issue. We're also considering moving `MultipartFormData` to take an `Encoder`, but that will be a separate effort.

ondrejstocek:
Thank you @cnoon,
looking into #2764 diff and changes look great.

Issue #2214, opened 2017-07-28
Best practice for Request Retrier with custom error serialization logic


cody1024d:
Hey guys,

I have the following scenario:

* I want to be able to parse the body of an HTTP Error (for example, a bad request response will contain a reason I can display to the user).
* I also want to, on a 401, leverage the Request Retrier to implement re-authentication

Here is the problem I'm seeing:

The Request Retrier doesn't fire unless some sort of validate is called, so that the error is set on the TaskSessionDelegate.

However, if I do want to parse the error body, the best practice is to NOT call validate, and handle the error-case within the completionHandler/custom extension.

What is the best practice then, for being able to both, call to the retrier in certain cases, AND be able to parse an HTTP error response body?

cody1024d:
Ideally, I think, there would be a way for me to signal the Request Retrier to fire, without having to call validate() to create an error on the delegate?

cody1024d:
As of now, I'm solving this by, in my completion handler, if the response is unsuccessful, etc., I perform the same deserialization that the responseJson method does on the body (if the body was successful).  I then have another layer up stream that converts the JsonDictionary to the expected models.

I have to say this isn't ideal, because it is a bit ugly, but I guess does work for my particular case. (I know for sure that the server will not be returning ridiculous bodies with an error)

```swift
self.manager.request(fullURL, method: httpMethod, parameters: parameters, encoding: encodingType, headers: self.headers)
                    .validate()
                    .responseJSON(completionHandler: { (response) in
                        if let value = response.result.value, let statusCode = response.response?.statusCode, response.result.isSuccess {
                            observer.onNext(NetworkResponse(statusCode: statusCode, responseBody: value))
                            observer.onCompleted()
                        } else if let error = response.result.error {
                            if let data = response.data, let statusCode = response.response?.statusCode, let jsonResponse = try? JSONSerialization.jsonObject(with: data) {
                                observer.onNext(NetworkResponse(statusCode: statusCode, responseBody: jsonResponse))
                            } else {
                                observer.onError(error)
                            }
                        } else {
                            observer.onError(NetworkError.UnexpectedResponse)
                        }
                    })
```

cnoon:
Hi @cody1024d,

[...]

PR #2716, opened 2019-02-12
Feature - Response Serializer Retry Support

cnoon:
This PR demonstrates one potential way we can add retry support to response serializers. This approach is definitely a work in progress, so please keep that in mind. The test suite is passing and the approach is sound, it just requires us to make some decisions on how certain edge cases should behave.

### Issue Link :link:

The design flaw has several forms, but generally results in users having to parse JSON payloads in `Validation` closures to get the errors extracted to all the request to be retried. Issue #2214 is one example of many use cases. 

> Please link other related issues if you know of them. I'll try to find some more later.

### Goals :soccer:

To add retry support to response serializers without changing the public APIs. Response serializers should be able to retry requests with a `RequestRetrier` when serialization errors occur.

The implementation of this isn't too bad amazingly. 

#### Threading Model

The biggest change is around the threading model. Instead of tacking all the response serializers into an operation queue, we need to keep track of them as a collection of closures. I've created a simple `responseSerializers` array in the `MutableState` of a `Request` where we can append them. I also added `append`, `next`, and `remove` response serializer functions to safely access the response serializers in thread-safe ways. 

> I'm not at all tied to these APIs, just needed to put something together to demonstrate one way to do this.

#### Response Serializers

The response serializer closures are a bit different now than they were. First off, the closures are simply appended to the array of `responseSerializers` in the `Request`. They are no longer appended to the `internalQueue`. This allows us to iterate through them instead of resuming the queue.

The next change in them is that for both the `DataRequest` and `DownloadRequest` APIs where they take a `responseSerializer`, we call a `retryRequest(:dueTo:completion:)` API that will retry the `Request` if the `RequestRetrier` determines it should be retried. This is the exact same API that we use for retrying requests after validation.

If the retrier determines that the request should be retried, the request is restarted and all the response serializers stay in the array. If a response serializer completes successfully, its `completion` closure is called and it is removed from the array. This behavior presents some interesting questions in terms of multiple response serializers:

1. What if the next response serializer throws an error that trips retry?
  - Does it make sense to retry after the first response serializer has succeeded and been removed?
  - Should all response serializers be executed until an error trips retry or they all succeed?
2. If any response serializer trips retry, should they all be executed the next time around?
  - Currently, a successful response serializer is popped from the queue when it succeeds
3. If you can retry from a response serializer, does it make sense to allow multiple response serializers?
  - Supporting multiple greatly overcomplicates things
  - It's not difficult to make a custom response serializer that makes multiple passes if necessary

IMO, it is really weird if a response serializer succeeds and is completed, then a second one fails, the request is retried, and the first response serializer is unaware that that happened. However, I do understand that that might make sense in some cases. Personally, I would implement that as a single response serializer that might need to make multiple passes on the payload. This would allow me to call a single completion with a custom payload type, but that's just what I would do. I don't completely understand all the use cases for people using multiple response serializers for a single call, nor do I think it wise to add a precondition to only allow a single response serializer per request.

> If anyone has some really good examples of use cases where multiple response serializers make perfect sense, I'd love to hear them. The more info we have here the better the implementation we can land on for everyone.

### Testing Details :mag:

This PR is a work in progress, so I do not want to build out specific test cases until we've landed on the approach. Currently, the entire test suite is passing as is, so the functionality is sound. However, we do not currently have tests that exercise all the current edge cases.

---

## Updates

I've now been able to finalize this PR and have some answers to the questions posed above that are worth documenting in the PR.

1. What if the next response serializer throws an error that trips retry?

Completion closures are only called once all response serializers have completed. We never want to call the first response serializer's completion closure, then have the second response serializer trip a retry, and then the first response serializer is just out-of-luck.

2. If any response serializer trips retry, should they all be executed the next time around?

Yes, they will all be executed again. The response serializer completion closures will only be executed once all response serializers have completed.

3. If you can retry from a response serializer, does it make sense to allow multiple response serializers?

We are still going to allow it. It's difficult to foresee use cases where you would use more than one, but we've decided to continue to support it in AF5.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -86,6 +86,8 @@ open class Request {
         var redirectHandler: RedirectHandler?
         /// `CachedResponseHandler` provided to handle caching responses.
         var cachedResponseHandler: CachedResponseHandler?
+        /// TODO: CN - docstring
+        var responseSerializers: [() -> Bool] = []
         /// `URLCredential` used for authentication challenges.
         var credential: URLCredential?
         /// All `URLRequest`s created by Alamofire on behalf of the `Request`.
@@ -150,6 +152,20 @@ open class Request {
     public private(set) var cachedResponseHandler: CachedResponseHandler? {
         get { return protectedMutableState.directValue.cachedResponseHandler }
         set { protectedMutableState.write { $0.cachedResponseHandler = newValue } }
+    }
+
+    // Response Serializers
+
+    func appendResponseSerializer(_ responseSerializer: @escaping () -> Bool) {
+        protectedMutableState.write { $0.responseSerializers.append(responseSerializer) }
+    }
+
+    func nextResponseSerializer() -> (() -> Bool)? {
+        return protectedMutableState.directValue.responseSerializers.first
+    }
+
+    func removeResponseSerializer() {
+        _ = protectedMutableState.directValue.responseSerializers.removeFirst()
     }
 
     // Credential
@@ -356,8 +372,19 @@ open class Request {
     /// Called to trigger retry or finish this `Request`.
     func retryOrFinish(error: Error?) {
         if let error = error, delegate?.willAttemptToRetryRequest(self) == true {
-            delegate?.retryRequest(self, ifNecessaryWithError: error)
-            return
+            delegate?.retryRequest(self, dueTo: error) { retryResult in
+                switch retryResult {
+                case .doNotRetry:
+                    self.finish()
+
+                case .doNotRetryWithError(let error):
+                    self.finish(error: error)
+
+                default:
+                    // No-op: retry is already being executed
+                    break
+                }
+            }
         } else {
             finish()
         }
@@ -368,9 +395,39 @@ open class Request {
         if let error = error { self.error = error }
 
         // Start response handlers
-        internalQueue.isSuspended = false
+        processNextResponseSerializer()
 
         eventMonitor?.requestDidFinish(self)
+    }
+
+    func processNextResponseSerializer() {
+        guard let responseSerializer = nextResponseSerializer() else {
+            // There's a single hook right now for the multipart upload to delete the original file
+            internalQueue.isSuspended = false
+            return
+        }
+
+        serializationQueue.async {
+            let succeeded = responseSerializer()
+
+            if succeeded {
+                // Remove the respose serializer that succeeded from the queue
+                self.removeResponseSerializer()
+
+                // Move on to the next response serializer
+                self.processNextResponseSerializer()
+
+                // Open questions:
+                // 1) What if the next response serializer throws an error that trips retry?
+                //    - Does it make sense to retry after the first response serializer has succeeded and been removed?
+                //    - Should all response serializers be executed until an error trips retry or they all succeed?
+                // 2) If any response serializer trips retry, should they all be executed the next time around?
+                //    - Currently, a successful response serializer is popped from the queue when it succeeds
+                // 3) If you can retry from a response serializer, does it make sense to allow multiple response serializers?
+                //    - Supporting multiple greatly overcomplicates things
+                //    - It's not difficult to make a custom response serializer that makes multiple passes if necessary
+            }
+        }
     }
 
     /// Resets all task related state for retry.
@@ -651,7 +708,7 @@ public protocol RequestDelegate: AnyObject {
     var sessionConfiguration: URLSessionConfiguration { get }
 
     func willAttemptToRetryRequest(_ request: Request) -> Bool
-    func retryRequest(_ request: Request, ifNecessaryWithError error: Error)
+    func retryRequest(_ request: Request, dueTo error: Error, completion: @escaping (RetryResult) -> Void)
 
     func cancelRequest(_ request: Request)
     func cancelDownloadRequest(_ request: DownloadRequest, byProducingResumeData: @escaping (Data?) -> Void)
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -340,8 +340,8 @@ open class Request {
         retryOrFinish(error: self.error)
     }
 
-    /// Called when the `RequestDelegate` is retrying this `Request`. Calls `reset()`.
-    func requestIsRetrying() {
+    /// Called when the `RequestDelegate` is going to retry this `Request`. Calls `reset()`.
+    func prepareForRetry() {
         protectedMutableState.write { $0.retryCount += 1 }
 
         reset()
@@ -353,14 +353,10 @@ open class Request {
     func retryOrFinish(error: Error?) {
         if let error = error, let delegate = delegate {
             delegate.retryResult(for: self, dueTo: error) { retryResult in
-                guard let retryResult = retryResult else { self.finish(); return }
-
                 switch retryResult {
-                case .doNotRetry:
-                    self.finish()
-                case .doNotRetryWithError(let error):
-                    self.finish(error: error)
-                default:
+                case .doNotRetry, .doNotRetryWithError:
+                    self.finish(error: retryResult.error)
+                case .retry, .retryWithDelay:
                     delegate.retryRequest(self, withDelay: retryResult.delay)
                 }
             }
@@ -706,7 +702,7 @@ extension Request: CustomDebugStringConvertible {
 public protocol RequestDelegate: AnyObject {
     var sessionConfiguration: URLSessionConfiguration { get }
 
-    func retryResult(for request: Request, dueTo error: Error, completion: @escaping (RetryResult?) -> Void)
+    func retryResult(for request: Request, dueTo error: Error, completion: @escaping (RetryResult) -> Void)
     func retryRequest(_ request: Request, withDelay timeDelay: TimeInterval?)
 
     func cancelRequest(_ request: Request)
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -68,10 +68,6 @@ open class Request {
     /// The `Request`'s delegate.
     public weak var delegate: RequestDelegate?
 
-    /// `OperationQueue` used internally to enqueue response callbacks. Starts suspended but is activated when the
-    /// `Request` is finished.
-    let internalQueue: OperationQueue
-
     // MARK: - Updated State
 
     /// Type encapsulating all mutable state that may need to be accessed from anything other than the `underlyingQueue`.
@@ -86,8 +82,12 @@ open class Request {
         var redirectHandler: RedirectHandler?
         /// `CachedResponseHandler` provided to handle caching responses.
         var cachedResponseHandler: CachedResponseHandler?
-        /// TODO: CN - docstring
-        var responseSerializers: [() -> Bool] = []
+        /// Response serialization closures that handle parsing responses.
+        var responseSerializers: [() -> Void] = []
+        /// Current response serialization closure index used to track iteration.
+        var responseSerializerIndex: Int = 0
+        /// Response serialization completion closures executed once all response serialization is complete.
+        var responseSerializerCompletions: [() -> Void] = []
         /// `URLCredential` used for authentication challenges.
         var credential: URLCredential?
         /// All `URLRequest`s created by Alamofire on behalf of the `Request`.
@@ -152,20 +152,6 @@ open class Request {
     public private(set) var cachedResponseHandler: CachedResponseHandler? {
         get { return protectedMutableState.directValue.cachedResponseHandler }
         set { protectedMutableState.write { $0.cachedResponseHandler = newValue } }
-    }
-
-    // Response Serializers
-
-    func appendResponseSerializer(_ responseSerializer: @escaping () -> Bool) {
-        protectedMutableState.write { $0.responseSerializers.append(responseSerializer) }
-    }
-
-    func nextResponseSerializer() -> (() -> Bool)? {
-        return protectedMutableState.directValue.responseSerializers.first
-    }
-
-    func removeResponseSerializer() {
-        _ = protectedMutableState.directValue.responseSerializers.removeFirst()
     }
 
     // Credential
@@ -253,10 +239,6 @@ open class Request {
         self.id = id
         self.underlyingQueue = underlyingQueue
         self.serializationQueue = serializationQueue
-        internalQueue = OperationQueue(maxConcurrentOperationCount: 1,
-                                       underlyingQueue: underlyingQueue,
-                                       name: "org.alamofire.request-\(id)",
-                                       startSuspended: true)
         self.eventMonitor = eventMonitor
         self.interceptor = interceptor
         self.delegate = delegate
@@ -400,37 +382,36 @@ open class Request {
         eventMonitor?.requestDidFinish(self)
     }
 
+    /// Returns the next response serializer closure to execute if there's one left.
+    func nextResponseSerializer() -> (() -> Void)? {
+        var responseSerializer: (() -> Void)?
+
+        protectedMutableState.write { mutableState in
+            if mutableState.responseSerializerIndex < mutableState.responseSerializers.count {
+                responseSerializer = mutableState.responseSerializers[mutableState.responseSerializerIndex]
+                mutableState.responseSerializerIndex += 1
+            }
+        }
+
+        return responseSerializer
+    }
+
+    /// Processes the next response serializer and calls all completions if response serialization is complete.
     func processNextResponseSerializer() {
         guard let responseSerializer = nextResponseSerializer() else {
-            // There's a single hook right now for the multipart upload to delete the original file
-            internalQueue.isSuspended = false
+            // Execute all response serializer completions
+            protectedMutableState.directValue.responseSerializerCompletions.forEach { $0() }
+
+            // Cleanup the request
+            cleanup()
+
             return
         }
 
-        serializationQueue.async {
-            let succeeded = responseSerializer()
-
-            if succeeded {
-                // Remove the respose serializer that succeeded from the queue
-                self.removeResponseSerializer()
-
-                // Move on to the next response serializer
-                self.processNextResponseSerializer()
-
-                // Open questions:
-                // 1) What if the next response serializer throws an error that trips retry?
-                //    - Does it make sense to retry after the first response serializer has succeeded and been removed?
-                //    - Should all response serializers be executed until an error trips retry or they all succeed?
-                // 2) If any response serializer trips retry, should they all be executed the next time around?
-                //    - Currently, a successful response serializer is popped from the queue when it succeeds
-                // 3) If you can retry from a response serializer, does it make sense to allow multiple response serializers?
-                //    - Supporting multiple greatly overcomplicates things
-                //    - It's not difficult to make a custom response serializer that makes multiple passes if necessary
-            }
-        }
+        serializationQueue.async { responseSerializer() }
     }
 
-    /// Resets all task related state for retry.
+    /// Resets all task and response serializer related state for retry.
     func reset() {
         error = nil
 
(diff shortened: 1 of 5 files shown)


Source/Request.swift
```diff
@@ -368,11 +395,41 @@ open class Request {
...
+                // Open questions:
+                // 1) What if the next response serializer throws an error that trips retry?
+                //    - Does it make sense to retry after the first response serializer has succeeded and been removed?
+                //    - Should all response serializers be executed until an error trips retry or they all succeed?
+                // 2) If any response serializer trips retry, should they all be executed the next time around?
+                //    - Currently, a successful response serializer is popped from the queue when it succeeds
+                // 3) If you can retry from a response serializer, does it make sense to allow multiple response serializers?
+                //    - Supporting multiple greatly overcomplicates things
+                //    - It's not difficult to make a custom response serializer that makes multiple passes if necessary
+            }
```
jshier: Answers:
1. My thought is that each serializer should be called and allowed to retry until they all complete, whether from success or failure, with no retry. Each serializer would be fired for every response, since the new response may lead to different output for a given serializer. If the first serializer fails repeatedly and can't be retried anymore, we may need to be able to store that error with the serializer until we're ready to call all of them. Generally, though, I think the behavior seen by the user should be all of the response serializers called from the same response.

2. I believe so, yes, as new responses would likely yield new response objects from the serializers.

3. While it does make things more complicated, it's once of the fundamental APIs of Alamofire. However, I don't know if anyone uses multiple serializers in production rather than just for debugging. I could imagine a few uses, mainly around logging (`EventMonitor` would be better for that now), so I'm not sure how disruptive of a change it would be.

cnoon: Okay @jshier, I just pushed up some good changes in 24898921. It updates the logic to always run all the response serializers, and resets the `Request` along with the response serialization state if any of the response serializers want to retry the `Request`.

It requires a couple new APIs around response serialization which are `appendResponseSerializer` and `responseSerializerDidComplete`. These essentially take the place of the `internalQueue` previously queueing up all the closures. I also removed the `internalQueue` and replaced it with a `cleanup` function instead. I thought that might be more applicable now given that's all that the `internalQueue` now does.

What I really love about this approach is that custom response serializers shouldn't have to deal with the fact that retry is even a thing. Everything can hide behind the `DataResponseSerializerProtocol` and use the `response(queue:responseSerializer:completion:)` API.

Source/Request.swift
```diff
@@ -535,6 +578,30 @@ open class Request {
...
+        protectedMutableState.write { $0.responseSerializers.append(closure) }
+    }
+
+    /// Notifies the `Request` that the response serializer is complete.
+    ///
+    /// - Parameter completion: The closure to execute once all response serializers are executed.
+    open func responseSerializerDidComplete(completion: @escaping () -> Void) {
+        protectedMutableState.write { $0.responseSerializerCompletions.append(completion) }
+        processNextResponseSerializer()
+    }
```
cnoon: We could probably get away with making these APIs `internal` if we really wanted to hide all this functionality.

Source/ResponseSerialization.swift
```diff
@@ -144,25 +144,54 @@ extension DataRequest {
...
+                                                    result: result)
+
+                        self.responseSerializerDidComplete {
+                            (queue ?? .main).async { completionHandler(response) }
+                        }
+
+                    default:
+                        // No-op: request is being retried
+                        break
+                    }
```
cnoon: I wonder if we should change this so that it's actually up to the caller to execute the retry. As you can see here, it's a bit awkward to rely on the internal implementation that retry has actually fired.

jshier: I think you're right. Serializer calls and wrapping those calls in retry logic are separate concerns, so they're better separated, if we can.

Source/Request.swift
jshier: I think we can make a final decision once the APIs are finalized.

Source/Request.swift
```diff
@@ -86,6 +82,12 @@ open class Request {
         var redirectHandler: RedirectHandler?
         /// `CachedResponseHandler` provided to handle caching responses.
         var cachedResponseHandler: CachedResponseHandler?
+        /// Response serialization closures that handle parsing responses.
+        var responseSerializers: [() -> Void] = []
+        /// Current response serialization closure index used to track iteration.
+        var responseSerializerIndex: Int = 0
```
jshier: I'm not huge fan of tracking an index like this, but Swift doesn't have any other suitable type. 😕

Source/Request.swift
```diff
@@ -1001,6 +1057,19 @@ open class UploadRequest: DataRequest {
 
         return stream
     }
+
+    open override func cleanup() {
```
jshier: I like this separation. I think it'll play into a formalization of lifetime event's we'll need to do to enable arbitrary subclasses.

jshier:
This looks pretty good. I've added some specific feedback. Functionally this is great, I just think some cleaner separation is in order.

Source/Request.swift
cnoon: Moved to `internal` APIs in 4dea9036 since users should always go through the `response` APIs.

Source/ResponseSerialization.swift
cnoon: Refactored to be separate in 687c3f1a. So much better in every way.

Source/Request.swift
```diff
@@ -346,18 +340,30 @@ open class Request {
...
+                guard let retryResult = retryResult else { self.finish(); return }
+
+                switch retryResult {
+                case .doNotRetry:
+                    self.finish()
+                case .doNotRetryWithError(let error):
+                    self.finish(error: error)
+                default:
+                    delegate.retryRequest(self, withDelay: retryResult.delay)
+                }
```
cnoon: The act of retrying is now explicit by the caller rather than having to know that the delegate has already triggered the retry. Much better design overall.

> As a side note, it seems weird to me that we have an optional delegate. Nothing works if it isn't set, so I wonder if we can rethink the memory management (maybe even the naming), and make it a strong, non-optional property. Something to consider in another PR.

Source/Request.swift
```diff
@@ -368,19 +374,62 @@ open class Request {
...
+    /// Returns the next response serializer closure to execute if there's one left.
+    func nextResponseSerializer() -> (() -> Void)? {
+        var responseSerializer: (() -> Void)?
+
+        protectedMutableState.write { mutableState in
+            let responseSerializerIndex = mutableState.responseSerializerCompletions.count
+
+            if responseSerializerIndex < mutableState.responseSerializers.count {
+                responseSerializer = mutableState.responseSerializers[responseSerializerIndex]
+            }
```
cnoon: Now we are just building the index dynamically. 🤙🏻

Source/Request.swift
```diff
@@ -650,8 +706,8 @@ extension Request: CustomDebugStringConvertible {
 public protocol RequestDelegate: AnyObject {
     var sessionConfiguration: URLSessionConfiguration { get }
 
-    func willAttemptToRetryRequest(_ request: Request) -> Bool
-    func retryRequest(_ request: Request, ifNecessaryWithError error: Error)
+    func retryResult(for request: Request, dueTo error: Error, completion: @escaping (RetryResult?) -> Void)
+    func retryRequest(_ request: Request, withDelay timeDelay: TimeInterval?)
```
cnoon: These APIs are so much more explicit than the previous iterations. Now you simply request a retry result, or retry the request. Both are very clear as to what they actually do. Previously, it was pretty hard to tell what did what.

Source/ResponseSerialization.swift
```diff
@@ -144,25 +143,57 @@ extension DataRequest {
...
+                                        data: self.data,
+                                        metrics: self.metrics,
+                                        serializationDuration: (end - start),
+                                        result: result)
+
+            self.eventMonitor?.request(self, didParseResponse: response)
+
+            if let serializerError = result.error, let delegate = self.delegate {
+                delegate.retryResult(for: self, dueTo: serializerError) { retryResult in
+                    var didComplete: (() -> Void)? = { completionHandler(response) }
```
cnoon: I pulled this into a local variable to DRY up the following:

```swift
self.responseSerializerDidComplete { 
    queue.async { completionHandler(response) } 
}
```

Quite complex to DRY it up properly, but this is working now as expected. Let me know if it's just too hard to follow this way and I can make it a bit more verbose.

Source/ResponseSerialization.swift
```diff
@@ -218,24 +248,57 @@ extension DownloadRequest {
...
+            let queue: DispatchQueue = queue ?? .main
+            let response = DownloadResponse(request: self.request,
+                                            response: self.response,
+                                            fileURL: self.fileURL,
+                                            resumeData: self.resumeData,
+                                            metrics: self.metrics,
+                                            serializationDuration: (end - start),
+                                            result: result)
+
+            if let serializerError = result.error, let delegate = self.delegate {
```
cnoon: I'm explicitly grabbing a reference to `delegate` here b/c everything falls apart if we optional chain and `delegate` is `nil` for some reason.

> I'd really recommend considering making that delegate a strong non-optional reference.

Tests/SessionTests.swift
```diff
@@ -1123,20 +1153,262 @@ class SessionTestCase: BaseTestCase {
...
+        XCTAssertTrue(session.requestTaskMap.isEmpty)
+
+        let errors: [AFError] = [json1Response, json2Response].compactMap { $0?.error?.asAFError }
+        XCTAssertEqual(errors.count, 2)
+
+        for error in errors {
+            XCTAssertTrue(error.isRequestAdaptationError)
+            XCTAssertEqual(error.localizedDescription, "Request adaption failed with error: URL is not valid: /adapt/error/2")
+        }
+    }
```
cnoon: These tests should cover all the use cases I can think of. Everything is working exactly as expected. 👍🏻

Source/Request.swift
cnoon: Oh man me too! I was actually able to just get rid of it altogether in 2ec42dd2 by using the counts on the two arrays.

Source/Request.swift
jshier: We can discuss it more, but in general, we need a way to break the reference cycle between the `Session` and its `Request`s, and the `weak` delegate does that. We could treat it as `unowned`, but we can't guarantee lifetimes like that. If we do want a strong delegate, both a rename / redesign of the API is in order, as well as another solution to the cycle issue.

Source/Request.swift
```diff
@@ -346,18 +340,30 @@ open class Request {
         retryOrFinish(error: self.error)
     }
 
-    /// Called when the `RequestDelegate` is retrying this `Request`.
+    /// Called when the `RequestDelegate` is retrying this `Request`. Calls `reset()`.
     func requestIsRetrying() {
```
jshier: We may want to rename this to better reflect what it's used for. Perhaps `prepareForRetry`?

Source/ResponseSerialization.swift
```diff
@@ -144,25 +143,57 @@ extension DataRequest {
...
+            let response = DataResponse(request: self.request,
+                                        response: self.response,
+                                        data: self.data,
+                                        metrics: self.metrics,
+                                        serializationDuration: (end - start),
+                                        result: result)
+
+            self.eventMonitor?.request(self, didParseResponse: response)
+
+            if let serializerError = result.error, let delegate = self.delegate {
```
jshier: It seems like this would be better as a `guard`, since the `else` would have a simple call to `self.responseSerializerDidComplete { queue.async { completionHandler(response) } }`.

Source/Request.swift
```diff
@@ -650,8 +706,8 @@ extension Request: CustomDebugStringConvertible {
 public protocol RequestDelegate: AnyObject {
     var sessionConfiguration: URLSessionConfiguration { get }
 
-    func willAttemptToRetryRequest(_ request: Request) -> Bool
-    func retryRequest(_ request: Request, ifNecessaryWithError error: Error)
+    func retryResult(for request: Request, dueTo error: Error, completion: @escaping (RetryResult?) -> Void)
```
jshier: Why is the `RetryResult` optional in the completion handler?

Source/ResponseSerialization.swift
```diff
@@ -144,25 +143,57 @@ extension DataRequest {
...
+                delegate.retryResult(for: self, dueTo: serializerError) { retryResult in
+                    var didComplete: (() -> Void)? = { completionHandler(response) }
+
+                    if let retryResult = retryResult {
```
jshier: Again this could be a `guard`, as there doesn't appear to be an `else`. You could `defer` the `didComplete` check.

Source/ResponseSerialization.swift
```diff
@@ -144,25 +143,57 @@ extension DataRequest {
...
+                            let response = DataResponse(request: self.request,
+                                                        response: self.response,
+                                                        data: self.data,
+                                                        metrics: self.metrics,
+                                                        serializationDuration: (end - start),
+                                                        result: result)
+
+                            didComplete = { completionHandler(response) }
+
+                        default:
```
jshier: We should probably be explicit here since this `default` only handles a single extra case.

Source/ResponseSerialization.swift
```diff
@@ -144,25 +143,57 @@ extension DataRequest {
...
+                            let result = Result<Serializer.SerializedObject>.failure(retryError)
+
+                            let response = DataResponse(request: self.request,
+                                                        response: self.response,
+                                                        data: self.data,
+                                                        metrics: self.metrics,
+                                                        serializationDuration: (end - start),
+                                                        result: result)
+
+                            didComplete = { completionHandler(response) }
```
jshier: This is what `didComplete` is already set to, no?

Source/ResponseSerialization.swift
jshier: I think a lot of the same structural changes I recommended above could be applied here.

Source/Session.swift
```diff
@@ -513,41 +513,34 @@ extension Session: RequestDelegate {
...
-                        request.requestIsRetrying()
-                        self.perform(request)
-                    }
-
-                    if let retryDelay = result.delay {
-                        self.rootQueue.after(retryDelay) { retry() }
-                    } else {
-                        self.rootQueue.async { retry() }
-                    }
+                if let retryResultError = result.error {
```
jshier: Also a good candidate for a `guard`.

Tests/SessionTests.swift
```diff
@@ -876,8 +888,10 @@ class SessionTestCase: BaseTestCase {
         waitForExpectations(timeout: timeout, handler: nil)
 
         // Then
+        XCTAssertEqual(handler.adaptCalledCount, 2)
         XCTAssertEqual(handler.adaptedCount, 2)
-        XCTAssertEqual(handler.retryCount, 2)
+        XCTAssertEqual(handler.retryCalledCount, 3)
```
jshier: Can you briefly outline why these counts have changed? Is it due to changes in what we're counting as retries between 4 and 5?

Source/Request.swift
cnoon: Absolutely, we certainly don't want to introduce a reference cycle. My thought was more around your final point of just rethinking the API to be less of a delegate and more of a core set of functionality that's required by the `Request` type.

Source/Request.swift
cnoon: Good call, will update.

Source/ResponseSerialization.swift
cnoon: Will update.

Source/Request.swift
cnoon: That's a great catch. For some reason when I built it, I didn't think of just having cases where the `Request` didn't have a `RequestRetrier` attached to it to just return `.doNotRetry`. For some reason I figured you'd want to get `nil` back instead. Turns out, that's silly. Much better to just return `.doNotRetry` in that case. Will fix. 👍🏻

Source/ResponseSerialization.swift
cnoon: Will update.

Source/ResponseSerialization.swift
cnoon: Yes, but the `response` is different in the first one. This one captures the new repackaged response with the retry error.

Tests/SessionTests.swift
cnoon: Absolutely. The reason why the retry counts have increased by one in almost all cases here is that AF5 now asks the retrier if it would like to retry in cases where response serialization fails. Since we didn't previously support that, most of these tests will see one more retry pass.

As an additional callout, I added `adaptCalledCount` and `retryCalledCount` to the already existing `adaptCount` and `retryCount` properties. The `Called` variants track if the method was ever called, where the non-Called variants only track whether the adapt or retry successfully execute.

Source/Request.swift
cnoon: Renamed in fea5fc86.

Source/ResponseSerialization.swift
cnoon: Switched to `guard` in fea5fc86.

Source/Request.swift
cnoon: Removed optional in fea5fc86.

Source/ResponseSerialization.swift
cnoon: Switched to `guard` in fea5fc86.

Source/ResponseSerialization.swift
cnoon: Good call, switched in fea5fc86.

Source/ResponseSerialization.swift
cnoon: Applied same changes in fea5fc86.

Source/Session.swift
cnoon: Switched to `guard` in fea5fc86.

Source/Session.swift
```diff
@@ -513,41 +513,35 @@ extension Session: RequestDelegate {
...
 
-    public func willAttemptToRetryRequest(_ request: Request) -> Bool {
-        return retrier(for: request) != nil
-    }
-
-    public func retryRequest(_ request: Request, ifNecessaryWithError error: Error) {
-        guard let retrier = retrier(for: request) else { request.finish(); return }
+    public func retryResult(for request: Request, dueTo error: Error, completion: @escaping (RetryResult) -> Void) {
+        guard let retrier = retrier(for: request) else {
+            rootQueue.async { completion(.doNotRetry) }
```
jshier: Any particular reason is async? It seems like you could just immediately call the completion handler, since `retryResult` should always be called on the root queue, right?

Source/Request.swift
```diff
@@ -346,18 +340,26 @@ open class Request {
...
+
         eventMonitor?.requestIsRetrying(self)
     }
 
     /// Called to trigger retry or finish this `Request`.
     func retryOrFinish(error: Error?) {
-        if let error = error, delegate?.willAttemptToRetryRequest(self) == true {
-            delegate?.retryRequest(self, ifNecessaryWithError: error)
-            return
+        if let error = error, let delegate = delegate {
```
jshier: This could be a `guard` as well. Sorry, missed it the first time.

Source/Session.swift
cnoon: Yep, you're absolutely right, you could. I chose to make it async to make the behavior consistent between both cases. If you call the `completion` closure on the calling thread, then the completion will run at the call-site before the next line of local scope is executed. In actual async cases, the local scope logic is executed before the completion closure is called. As a general guiding principle, it is wise to always call completion closures in either a consistent synchronous or asynchronous case to ensure client logic executes in a consistent manner.

Source/Request.swift
cnoon: Fixed in 68b44f49.

Source/Session.swift
jshier: 👍

PR #2718, opened 2019-02-13
Merge and cleanup append(_ fileURL, …) functions

rivera-ernesto:
Now `fileName` an `mimeType` are optional parameters that are generated if needed.  

### Issue Link :link:
Before, besides having two functions, we had to specify both `fileName` an `mimeType` or not specify neither. Now we have the freedom to specify one, both or neither.

### Goals :soccer:
Cleaner single function with very Swift optional parameters that gracefully fallback to proper default values.

As described in the new merged function documentation:
> The filename in the `Content-Disposition` HTTP header is generated from the last path component of the `fileURL`. The `Content-Type` HTTP header MIME type is generated by mapping the `fileURL` extension to the system associated MIME type.

### Testing Details :mag:
None needed. Code that used to call one or the other function now successfully use the new function without any changes.

Similarly, other functions should get merged together with optional parameters. This is no Objective-C, so no need to repeat functions with more/less parameters.


Source/MultipartFormData.swift
```diff
@@ -189,43 +189,22 @@ open class MultipartFormData {
...
+    /// system associated MIME types.
     ///
     /// - parameter fileURL:  The URL of the file whose content will be encoded into the multipart form data.
     /// - parameter name:     The name to associate with the file content in the `Content-Disposition` HTTP header.
     /// - parameter fileName: The filename to associate with the file content in the `Content-Disposition` HTTP header.
     /// - parameter mimeType: The MIME type to associate with the file content in the `Content-Type` HTTP header.
-    public func append(_ fileURL: URL, withName name: String, fileName: String, mimeType: String) {
+    public func append(_ fileURL: URL, withName name: String, fileName: String? = nil, mimeType: String? = nil) {
+        let fileName = fileName ?? fileURL.lastPathComponent
+        let mimeType = mimeType ?? self.mimeType(forPathExtension: fileURL.pathExtension)
```
cnoon: You're missing the `isEmpty` checks from the previous function. Could you add those along with the same body part error logic here?

Source/MultipartFormData.swift
```diff
@@ -189,43 +189,22 @@ open class MultipartFormData {
...
     /// - Encoded file data
     /// - Multipart form boundary
     ///
-    /// The filename in the `Content-Disposition` HTTP header is generated from the last path component of the
-    /// `fileURL`. The `Content-Type` HTTP header MIME type is generated by mapping the `fileURL` extension to the
-    /// system associated MIME type.
-    ///
-    /// - parameter fileURL: The URL of the file whose content will be encoded into the multipart form data.
-    /// - parameter name:    The name to associate with the file content in the `Content-Disposition` HTTP header.
-    public func append(_ fileURL: URL, withName name: String) {
```
cnoon: Could you also collapse the `Data` methods in the same way you have for the `URL` methods?

Source/MultipartFormData.swift
rivera-ernesto: No need for the checks anymore:

Source/MultipartFormData.swift
```diff
@@ -189,43 +189,22 @@ open class MultipartFormData {
...
+    /// `fileURL`. The `Content-Type` HTTP header MIME type, if needed, is generated by mapping the `fileURL` extension to the
+    /// system associated MIME types.
     ///
     /// - parameter fileURL:  The URL of the file whose content will be encoded into the multipart form data.
     /// - parameter name:     The name to associate with the file content in the `Content-Disposition` HTTP header.
     /// - parameter fileName: The filename to associate with the file content in the `Content-Disposition` HTTP header.
     /// - parameter mimeType: The MIME type to associate with the file content in the `Content-Type` HTTP header.
-    public func append(_ fileURL: URL, withName name: String, fileName: String, mimeType: String) {
+    public func append(_ fileURL: URL, withName name: String, fileName: String? = nil, mimeType: String? = nil) {
+        let fileName = fileName ?? fileURL.lastPathComponent
```
rivera-ernesto: ...here `fileName` gets checked, if `nil` we use `fileURL.lastPathComponent` as before.

Source/MultipartFormData.swift
rivera-ernesto: ...same here for `mimeType`.

Source/MultipartFormData.swift
rivera-ernesto: Sure, let me do that.

Source/MultipartFormData.swift
```diff
@@ -177,7 +139,7 @@ open class MultipartFormData {
     /// - parameter name:     The name to associate with the data in the `Content-Disposition` HTTP header.
     /// - parameter fileName: The filename to associate with the data in the `Content-Disposition` HTTP header.
     /// - parameter mimeType: The MIME type to associate with the data in the `Content-Type` HTTP header.
-    public func append(_ data: Data, withName name: String, fileName: String, mimeType: String) {
+    public func append(_ data: Data, withName name: String, fileName: String? = nil, mimeType: String? = nil) {
         let headers = contentHeaders(withName: name, fileName: fileName, mimeType: mimeType)
```
rivera-ernesto: Really easy `append(_ data: Data, ...)` functions merge as `contentHeaders(...)` already handles optional `fileName` and `mimeType`s.

Source/MultipartFormData.swift
rivera-ernesto: Done.

rivera-ernesto:
Just a gentle ping.

cnoon:
Hi @rivera-ernesto, I just opened #2766 with a few of these changes to go into beta 4. I made sure to maintain your attribution. I'm going to go ahead and close out this PR. Please direct all future comments to #2766.

Cheers. 🍻

Source/MultipartFormData.swift
cnoon: There is, it's a change in behavior and the error would no longer be thrown. Regardless, I've opened up #2766 to move this through as quickly as possible. Thanks for all your help here! 🍻

Issue #2719, opened 2019-02-15
status code validation inefficient


StevenArmandLee:
* [ x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

on https://github.com/Alamofire/Alamofire/blob/3e488dcd5caf8d76f3da79c7b6d7f4d784ca3296/Source/Validation.swift#L70-L94

The function is inefficient, the reasons are:

- acceptableStatusCode creating array with size of n
- On every request, we validate the status code with  `if acceptableStatusCodes.contains(response.statusCode) {` with complexity of O(n)

recommend to use comparison instead of contains.

PR #2720, opened 2019-02-15
Changed array to range

StevenArmandLee:
### Issue Link :link:
status code validation inefficient #2719

### Goals :soccer:
increase the efficiency of the code, `contain` in array is O(n) where as `contain` in Range is comparison 
https://github.com/apple/swift/blob/3b6c6cc0a5424fbac81213376969432b0b16ad5e/stdlib/public/core/Range.swift#L165-L168 

### Testing Details :mag:
Ran the existing unit test and all passed

### DIFF
--- a/Source/Validation.swift
+++ b/Source/Validation.swift
@@ -67,7 +67,7 @@ extension Request {
 
     // MARK: Properties
 
-    fileprivate var acceptableStatusCodes: [Int] { return Array(200..<300) }
+    fileprivate var acceptableStatusCodes: Range<Int> { return 200..<300 }
 
     fileprivate var acceptableContentTypes: [String] {
         if let accept = request?.value(forHTTPHeaderField: "Accept") {
--- a/Source/Validation.swift
+++ b/Source/Validation.swift
@@ -67,7 +67,7 @@ extension Request {
 
     // MARK: Properties
 
-    fileprivate var acceptableStatusCodes: Range { return 200..<300 }
+    fileprivate var acceptableStatusCodes: Range<Int> { return 200..<300 }
 
     fileprivate var acceptableContentTypes: [String] {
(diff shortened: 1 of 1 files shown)


StevenArmandLee:
tested with measure, using `String` giving us 0.00223s where `Range` gives us 0.0000106s with test code of `acceptableStatusCodes.contains(299)`

agnosticdev:
Very interesting @StevenArmandLee.  Are the time measurements the elapsed times measured from the call to `acceptableStatusCodes` to validate the response code?

For example:
```swift
let dateStart = Date()
...
let elapsedTime = dateStart.timeIntervalSinceNow
print("\(abs(elapsedTime))")
```

StevenArmandLee:
@agnosticdev I did not test the full function, I just the contains method with measure block provided by XCTest
here is my test code
```
func testSpeed() {
        var acceptableStatusCodes: [Int] { return Array(200..<300) }
//        var acceptableStatusCodes: Range<Int> { return 200..<300 }
        self.measure {
            acceptableStatusCodes.contains(299)
            acceptableStatusCodes.contains(299)
            acceptableStatusCodes.contains(299)
            acceptableStatusCodes.contains(299)
            acceptableStatusCodes.contains(299)
            acceptableStatusCodes.contains(299)
            acceptableStatusCodes.contains(299)
        }
    }
```

the reason I use 299 is to present the worst case scenario.

agnosticdev:
Excellent, thank you for providing that testing information @StevenArmandLee

StevenArmandLee:
Will this PR being reviewed?

cnoon:
Sure @StevenArmandLee, it's just not our highest priority right now. Thanks for understanding.

jshier:
This is a relatively minor improvement but an improvement nonetheless. Thanks for the PR!

PR #2728, opened 2019-02-25
Fix #2682: Cancel all requests on session invalidation or deinit.

### LINKED ISSUE #2682: Alamofire 5: Invalidated sessions don't complete requests
Due to Alamofire 5's async request creation, `Session` instances can be `deinit`'d without completing their requests, unlike AF4. It's important this behavior exists so users can catch this implicit behavior, as it's one of the biggest causes of GitHub issues.

jshier:
### Issue Link :link:
#2682

### Goals :soccer:
This PR ensures that all `Request`s finish with a specific error when the owning `Session` is `deinit`d or the underlying `URLSession` is invalidated.

Previously, after the Alamofire 5 refactor, `Request`s were no longer properly being finished, due to the separation of the `URLSessionDelegate` implementation from the `Session` itself, which meant the `Session` would be `deinit`d by the time the `SessionStateProvider` could act on the invalidation.

This PR adds an additional `SessionStateProvider` method, `cancelRequestsForSessionInvalidation(with error: Error?)` and a new `AFError` case, `invalidatedSession(error: Error?)` to properly support the new failure.

### Testing Details :mag:
A test was added for the invalidation after deinit case. I'm not sure the actual session invalidation case can be tested.

### DIFF
--- a/Source/RequestTaskMap.swift
+++ b/Source/RequestTaskMap.swift
@@ -26,63 +26,67 @@ import Foundation
 
 /// A type that maintains a two way, one to one map of `URLSessionTask`s to `Request`s.
 struct RequestTaskMap {
-    private var requests: [URLSessionTask: Request]
-    private var tasks: [Request: URLSessionTask]
+    private var tasksToRequests: [URLSessionTask: Request]
+    private var requestsToTasks: [Request: URLSessionTask]
 
-    init(requests: [URLSessionTask: Request] = [:], tasks: [Request: URLSessionTask] = [:]) {
-        self.requests = requests
-        self.tasks = tasks
+    var requests: [Request] {
+        return Array(tasksToRequests.values)
+    }
+
+    init(tasksToRequests: [URLSessionTask: Request] = [:], requestsToTasks: [Request: URLSessionTask] = [:]) {
+        self.tasksToRequests = tasksToRequests
+        self.requestsToTasks = requestsToTasks
     }
 
     subscript(_ request: Request) -> URLSessionTask? {
-        get { return tasks[request] }
+        get { return requestsToTasks[request] }
         set {
             guard let newValue = newValue else {
-                guard let task = tasks[request] else {
+                guard let task = requestsToTasks[request] else {
                     fatalError("RequestTaskMap consistency error: no task corresponding to request found.")
                 }
 
-                tasks.removeValue(forKey: request)
-                requests.removeValue(forKey: task)
+                requestsToTasks.removeValue(forKey: request)
+                tasksToRequests.removeValue(forKey: task)
 
                 return
             }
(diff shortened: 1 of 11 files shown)


jshier:
Two unanswered questions remain:
1. Should we finish or cancel ongoing `Request`s? Since the `Session` is going away, there's no way to call back from the `URLSession` delegates to do anything, so whatever we do has to happen immediately. However, just `finish`ing the `Request`s seems potentially problematic.
2. Do we want additional tests for active vs. inactive `Request`s, many simultaneous `Request`s?

Source/Session.swift
```diff
@@ -43,10 +43,10 @@ open class Session {
     var requestTaskMap = RequestTaskMap()
     public let startRequestsImmediately: Bool
 
-    public init(startRequestsImmediately: Bool = true,
-                session: URLSession,
+    public init(session: URLSession,
                 delegate: SessionDelegate,
                 rootQueue: DispatchQueue,
+                startRequestsImmediately: Bool = true,
```
jshier: Moving this parameter to follow Swift guideline stating defaulted parameters should come after non-defaulted ones.

Source/RequestTaskMap.swift
```diff
@@ -26,63 +26,67 @@ import Foundation
...
-    private var requests: [URLSessionTask: Request]
-    private var tasks: [Request: URLSessionTask]
+    private var tasksToRequests: [URLSessionTask: Request]
+    private var requestsToTasks: [Request: URLSessionTask]
 
-    init(requests: [URLSessionTask: Request] = [:], tasks: [Request: URLSessionTask] = [:]) {
-        self.requests = requests
-        self.tasks = tasks
+    var requests: [Request] {
+        return Array(tasksToRequests.values)
```
jshier: This may have a performance impact, so we could just return the `values` collection, since all we need to do is loop over the contents.

jshier:
@cnoon Thinking about this more, this seems like only a partial solution, as it only works for `Request`s mapped to `*Task`s. It's also racy, given that the `*Task` isn't created until after the async `URLRequest` creation. I think a full solution will require us to keep track of all in-flight `Request`s, but before we do that we need to properly define the `Request` lifetime and what in-flight means. I'm thinking some sort of `weak` storage, but I'm not sure if that guarantees the lifetimes we'd want. Any ideas?

In the mean time I think I'm going to refactor this PR to separate the invalidation error from the `deinit` error, given they're really separate causes. I'll ping you again when that's done. If this brief version works out, we should merge it soon.

jshier:
I've updated this PR to fully fix the validator leaks as well as separate the error case for deinitialization from invalidation.

This isn't a full solution but fixes the biggest issues.

Issue #2698, opened 2019-01-19
[V5] SPM generated Xcode project fails to compile with minimum target compiler errors


kdawgwilk:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

1.
```
$ swift package init --type=executable
```
2. Add Alamofire 5 as a dependency.  Here is the Package.swift:

```swift
let package = Package(
    name: "PackageName",
    dependencies: [
        // Dependencies declare other packages that this package depends on.
        // .package(url: /* package url */, from: "1.0.0"),
        .package(url: "https://github.com/Alamofire/Alamofire.git", from: "5.0.0-beta.1")
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages which this package depends on.
        .target(
            name: "PackageName",
            dependencies: ["Alamofire"]),
        .testTarget(
            name: "PackageNameTests",
            dependencies: ["PackageName"]),
    ]
)
```
3.
```
$ swift package generate-xcodeproj
```
4. Build in Xcode

## What did you expect to happen?

Compiles without error

## What happened instead?

25 Compiler errors similar to the following
```
/path/to/project/ProjectName/.build/checkouts/Alamofire.git-3767994841630926798/Source/EventMonitor.swift:62:95: 'URLSessionTaskMetrics' is only available on OS X 10.12 or newer
```

## Alamofire Environment

**Alamofire version:** 5.0.0-beta.1
**Xcode version:** 10.1
**Swift version:** 4.2.1
**Platform(s) running Alamofire:** macOS
**macOS version running Xcode:** 10.14.2

## Demo Project

Swift Package Manager fresh project so no demo required

### Option 1

Build outside of Xcode from CLI and specify target:
```
$ swift build -Xswiftc "-target" -Xswiftc "x86_64-apple-macosx10.12"
```
### Option 2

Create `.xcconfig` and generate project using it:

Config.xcconfig
```
MACOSX_DEPLOYMENT_TARGET = 10.12
```
Generate Xcode project using config
```
$ swift-package generate-xcodeproj --xcconfig-overrides Config.xcconfig
```

## Possible Solution

[...]

PR #2736, opened 2019-03-06
Update swiftLanguageVersions to swift-tools-version:4.2

mattt:
This PR fixes incompatibilities with Swift 5 compiler due to an API change in the Package.swift manifest API.

`swiftLanguageVersions` changed from taking `[Int]` to `[SwiftVersion]` in PackageDescription API Version 4.2. For more information, see https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescriptionV4_2.md#swift-language-version.

I've tested this with Swift 4.2.1 (swiftlang-1000.11.42) and Swift 5 (swiftlang-1001.0.63.8), and it appears to work for both. It was difficult to verify because of build errors related to #2698, but I was able to work around that by passing an explicit macOS 10.12 target with the `-Xswiftc` flag:

```terminal
$ swift build -Xswiftc -target -Xswiftc x86_64-apple-macosx10.12
Compile Swift Module 'Alamofire' (30 sources)
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -37,5 +37,5 @@ let package = Package(
             name: "Alamofire",
             path: "Source")
     ],
-    swiftLanguageVersions: [3, 4]
+    swiftLanguageVersions: [.v3, .v4]
 )


jshier:
Thanks Mattt!

Once Xcode 10.2 ships we'll update to the v5 syntax and fix #2698.

mattt:
@jshier Thanks for the quick merge!

Please see #2737 if you're interested in a solution that isn't blocked by Xcode 10.2 GM.

PR #2746, opened 2019-03-15
Delay to adapt request

over140:

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -361,11 +361,14 @@ open class DataRequest: Request {
         let urlRequest: URLRequest
 
         func task(session: URLSession, adapter: RequestAdapter?, queue: DispatchQueue) throws -> URLSessionTask {
-            do {
-                let urlRequest = try self.urlRequest.adapt(using: adapter)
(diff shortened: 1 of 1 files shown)


Source/Request.swift
```diff
@@ -361,11 +361,14 @@ open class DataRequest: Request {
         let urlRequest: URLRequest
 
         func task(session: URLSession, adapter: RequestAdapter?, queue: DispatchQueue) throws -> URLSessionTask {
-            do {
-                let urlRequest = try self.urlRequest.adapt(using: adapter)
-                return queue.sync { session.dataTask(with: urlRequest) }
-            } catch {
-                throw AdaptError(error: error)
+            let request = self.urlRequest
```
agnosticdev: Is there a benefit to wrapping the entire do/catch block in `queue.sync {...}` instead of just `session.dataTask(with: urlRequest)`?

cnoon:
Hi @over140,

I'm going to go ahead and close this PR out for now. Without your feedback, we can't move forward, and we don't want to leave these PRs open forever. If you comment back and it makes sense to move this change through, we'll happily re-open and work it into `master`.

Cheers. 🍻

Issue #2750, opened 2019-03-17
[Alamofire 5]: An Alamofire Namespace


jshier:
To better support Alamofire shipping extensions of types we don't own, we need a namespace similar to those offered by libraries like RxSwift or Kingfisher. This would both properly limit the visibility of those extensions to reduce conflicts and give us a common location for extensions in other Alamofire libraries like AlamofireImage. Following what RxSwift does, we could do something like this:
```swift
struct AlamofireExtension<ExtendedType> {
    let type: ExtendedType
    
    init(_ type: ExtendedType) {
        self.type = type
    }
}

protocol AlamofireExtended {
    associatedtype ExtendedType
    
    static var af: AlamofireExtension<ExtendedType>.Type { get set }
    
    var af: AlamofireExtension<ExtendedType> { get set }
}

extension AlamofireExtended {
    static var af: AlamofireExtension<Self>.Type {
        get { return AlamofireExtension<Self>.self }
        set { }
    }
    
    var af: AlamofireExtension<Self> {
        get { return AlamofireExtension(self) }
        set { }
    }
}
```

jshier:
Completed in #2758.

PR #2758, opened 2019-03-24
[Alamofire 5]: AlamofireExtended extension point.

jshier:
### Issue Link :link:
#2750

### Goals :soccer:
This PR adds the `AlamofireExtended` protocol and `AlamofireExtension` type to enable `.af` extension points on any conforming type. This allows us to put all of our extensions of types we don't own behind a single entry point.

Types conforming to `AlamofireExtended` receive the `.af` static and instance properties. Extending `AlamofireExtension where ExtendedType` is necessary to add functionality to those extension points on particular types. Our extensions of various Security types, as well as `URLSessionConfiguration`, have been updated to use this pattern. 

### Testing Details :mag:
Tests of the extended API have been updated, but no specific tests have been added, as the extension protocol and type have no functionality without extensions.

### DIFF
--- a/Source/AlamofireExtended.swift
+++ b/Source/AlamofireExtended.swift
@@ -1,7 +1,7 @@
 //
-//  URLSessionConfiguration+Alamofire.swift
+//  AlamofireExtended.swift
 //
-//  Copyright (c) 2014-2018 Alamofire Software Foundation (http://alamofire.org/)
+//  Copyright (c) 2019 Alamofire Software Foundation (http://alamofire.org/)
 //
 //  Permission is hereby granted, free of charge, to any person obtaining a copy
 //  of this software and associated documentation files (the "Software"), to deal
@@ -22,14 +22,34 @@
 //  THE SOFTWARE.
 //
 
-import Foundation
+/// Type that acts as a generic extension point for all `AlamofireExtended` types.
+public struct AlamofireExtension<ExtendedType> {
+    /// Stores the type or metatype of any extended type.
+    let type: ExtendedType
 
-extension URLSessionConfiguration: AlamofireExtended { }
-extension AlamofireExtension where ExtendedType: URLSessionConfiguration {
-    public static var `default`: URLSessionConfiguration {
-        let configuration = URLSessionConfiguration.default
-        configuration.httpHeaders = .default
+    init(_ type: ExtendedType) {
+        self.type = type
+    }
+}
 
-        return configuration
+/// Protocol describing the `af` extension points for Alamofire extended types.
+public protocol AlamofireExtended {
+    associatedtype ExtendedType
+
+    /// Static Alamofire extension point.
+    static var af: AlamofireExtension<ExtendedType>.Type { get set }
+    /// Instance Alamofire extension point.
+    var af: AlamofireExtension<ExtendedType> { get set }
+}
+
+public extension AlamofireExtended {
+    static var af: AlamofireExtension<Self>.Type {
+        get { return AlamofireExtension<Self>.self }
+        set { }
+    }
+
+    var af: AlamofireExtension<Self> {
+        get { return AlamofireExtension(self) }
+        set { }
     }
 }
--- a/Source/ServerTrustEvaluation.swift
+++ b/Source/ServerTrustEvaluation.swift
@@ -85,24 +85,6 @@ public protocol ServerTrustEvaluating {
     #endif
 }
 
-extension Array where Element == ServerTrustEvaluating {
-    #if os(Linux)
-    // Add this same convenience method for Linux.
-    #else
-    /// Evaluates the given `SecTrust` value for the given `host`.
-    ///
-    /// - Parameters:
-    ///   - trust: The `SecTrust` value to evaluate.
-    ///   - host:  The host for which to evaluate the `SecTrust` value.
-    /// - Returns: Whether or not the evaluator considers the `SecTrust` value valid for `host`.
-    func evaluate(_ trust: SecTrust, forHost host: String) throws {
-        for evaluator in self {
-            try evaluator.evaluate(trust, forHost: host)
-        }
-    }
-    #endif
-}
-
 // MARK: - Server Trust Evaluators
 
 /// An evaluator which uses the default server trust evaluation while allowing you to control whether to validate the
@@ -120,10 +102,10 @@ public final class DefaultTrustEvaluator: ServerTrustEvaluating {
 
     public func evaluate(_ trust: SecTrust, forHost host: String) throws {
         if validateHost {
-            try trust.validateHost(host)
+            try trust.af.performValidation(forHost: host)
         }
 
-        try trust.performDefaultEvaluation(forHost: host)
+        try trust.af.performDefaultValidation(forHost: host)
     }
 }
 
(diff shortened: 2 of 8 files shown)


Source/ServerTrustEvaluation.swift
```diff
@@ -352,9 +334,30 @@ public final class DisabledEvaluator: ServerTrustEvaluating {
...
+    ///   - trust: The `SecTrust` value to evaluate.
+    ///   - host:  The host for which to evaluate the `SecTrust` value.
+    /// - Returns: Whether or not the evaluator considers the `SecTrust` value valid for `host`.
+    func evaluate(_ trust: SecTrust, forHost host: String) throws {
+        for evaluator in self {
+            try evaluator.evaluate(trust, forHost: host)
+        }
+    }
+    #endif
+}
```
cnoon: Seems like we should probably hide this internally or inside `AlamofireExtended` as well. Pretty sure it can't be done with `AlamofireExtended` which is probably why you didn't already do it if I had to guess.

Source/Session.swift
```diff
@@ -74,7 +74,7 @@ open class Session {
         delegate.stateProvider = self
     }
 
-    public convenience init(configuration: URLSessionConfiguration = .alamofireDefault,
+    public convenience init(configuration: URLSessionConfiguration = URLSessionConfiguration.af.default,
```
cnoon: `URLSessionConfiguration.af.default` -> `.af.default`?

Source/URLRequest+Alamofire.swift
```diff
@@ -24,9 +24,9 @@
...
 import Foundation
 
-extension URLRequest {
+public extension URLRequest {
     var method: HTTPMethod? {
-        guard let httpMethod = self.httpMethod else { return nil }
+        guard let httpMethod = httpMethod else { return nil }
         return HTTPMethod(rawValue: httpMethod)
     }
 }
```
cnoon: Are we sure we don't want to hide this in `AlamofireExtension` as well?

Source/ServerTrustEvaluation.swift
jshier: My standard for hiding was: if we don't own either the extended or returned types, put it inside the extension point. If we do, like for this array convenience, or the `headers`, I didn't.

Source/Session.swift
jshier: Unfortunately that doesn't seem to work. Using the `.af.default` results in a compiler error about needing more type context.

Source/URLRequest+Alamofire.swift
jshier: Since we own the `HTTPMethod` type, I left it out, as usage can be disambiguated at the use site if there's a collision. We could certainly put everything inside, it's just makes it more awkward.

jshier:
@cnoon Back to you. If you disagree about which extensions should be inside the extension point, let me know and I can move everything inside.

Source/ServerTrustEvaluation.swift
cnoon: Thanks for the clarification. I think that's a really sound approach.

Source/URLRequest+Alamofire.swift
cnoon: Yep, makes sense.

Source/AlamofireExtended.swift
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/// Protocol describing the `af` extension points for Alamofire extended types.
+public protocol AlamofireExtended {
+    associatedtype ExtendedType
+
+    /// Static Alamofire extension point.
+    static var af: AlamofireExtension<ExtendedType>.Type { get set }
+    /// Instance Alamofire extension point.
+    var af: AlamofireExtension<ExtendedType> { get set }
```
cnoon: Do we actually need `set` functionality here? Would it be possible to remove it and just use computed `get` properties in the protocol extension?

Source/AlamofireExtended.swift
```diff
@@ -0,0 +1,55 @@
...
+
+public extension AlamofireExtended {
+    static var af: AlamofireExtension<Self>.Type {
+        get { return AlamofireExtension<Self>.self }
+        set { }
+    }
+
+    var af: AlamofireExtension<Self> {
+        get { return AlamofireExtension(self) }
+        set { }
```
cnoon: I've never built a no-op `set` implementation before. Makes sense that it can just be empty. I wonder if we could just switch these to this:

```swift
static var af: AlamofireExtension<Self>.Type { return AlamofireExtension<Self>.self }
var af: AlamofireExtension<Self> { return AlamofireExtension(self) }
```

jshier: I copied the [implementation from RxSwift](https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Reactive.swift), except for the comment in the `set`, which states: "// this enables using Reactive to "mutate" base type". So while I'm not sure we need it, it may be useful.

cnoon: I wanted to learn more about this so I put together a simple playground and learned a few things. First off, this is a cool pattern for sure, very powerful. Secondly, there are a couple hidden tricks that aren't super obvious at first.

### 1) Setting only works for reference types

The `ExtendedType` in the `AlamofireExtension` is a `let`, so you can only mutate the state of the `ExtendedType` instance if it's a reference type. You could switch it to a `var`, but that causes more harm than good.

Also, you MUST have the `set` declaration in `AlamofireExtended` for being able to modify state on the `ExtendedType` instance. Otherwise you get the following error:

> "Cannot assign to property: 'af' is a get-only property"

### 2) Value types can use these APIs, but "setting" state doesn't work as expected

Since a new `AlamofireExtension` type is made on each `af` call, the internal type is copied if it's a value type. Therefore, setting things on one instance won't probably end up what you want.

```swift
import Foundation
import Security

public struct AlamofireExtension<ExtendedType> {
    let type: ExtendedType
    init(_ type: ExtendedType) { self.type = type }
}

public protocol AlamofireExtended {
    associatedtype ExtendedType

    static var af: AlamofireExtension<ExtendedType>.Type { get set }
    var af: AlamofireExtension<ExtendedType> { get set }
}

public extension AlamofireExtended {
    static var af: AlamofireExtension<Self>.Type {
        get { return AlamofireExtension<Self>.self }
        set { }
    }

    var af: AlamofireExtension<Self> {
        get { return AlamofireExtension(self) }
        set { print("Called with new value: \(newValue)") }
    }
}

extension URLRequest: AlamofireExtended {}
public extension AlamofireExtension where ExtendedType == URLRequest {
    var method: String {
        get { return type.httpMethod ?? "GET" }
//        set { type.httpMethod = newValue }
    }
}

extension URLSession: AlamofireExtended {}
extension AlamofireExtension where ExtendedType: URLSession {
    var fancyDescription: String {
        get { return type.sessionDescription ?? "" }
        set { type.sessionDescription = newValue }
    }
}

var request = URLRequest(url: URL(string: "https://httpbin.org/post")!)
request.httpMethod = "POST"

let method = request.af.method
//print("before set")
//request.af.method = "HEAD"
//print("after set")

let method2 = request.af.method

var session = URLSession(configuration: .default)
let desc = session.af.fancyDescription
print("before set")
session.af.fancyDescription = "This is cool stuff!"
print("after set")
let desc2 = session.af.fancyDescription
```

Source/AlamofireExtended.swift
cnoon: Yep, we want the `set` functionality. See other comments for more info.

PR #2764, opened 2019-03-26
Refactor - Upload MultipartFormData with Custom Boundary

cnoon:
This PR updates the upload APIs for `MultipartFormData` to support a builder closure for the more convenient API, and a `MultipartFormData` instance for the more complex one. This allows users to create custom boundaries if necessary. The builder closure does not expose the boundary as a parameter.

### Issue Link :link:
The original issue was raised in #2705.

### Goals :soccer:
We want to allow users that MUST change the boundary to still use the top-level APIs. We also want to avoid having to add new APIs to handle this use case if possible.

Pretty simple change overall, we just switched the complex API over from using a builder closure to taking an instance. This also allows us to remove the `fileManager` parameter from the API.

### Testing Details :mag:
Test suite is still passing as expected.


cnoon:
Good call, I'll add a test doing just that.

cnoon:
Added test in 47c71e02.

PR #2765, opened 2019-03-26
🐷 Refactor - Switches and Queues

cnoon:
This PR is a simplification of #2691 where some of the changes have been culled out due to either being changes we do not agree with, or changes that were incorrect.

### Issue Link :link:
This is a subset of #2691 that maintains attribution to @jdisho.

### Goals :soccer:
To refactor the switches into guards where applicable and also move away from optional queues in the response serialization logic.

No real details here worth calling out other than this PR is piggy-backed on #2716 since that PR contains a lot of changes with regards to the DispatchQueue usage in response serialization.

### Testing Details :mag:
Test suite still looks good. These changes do not change any actual functionality.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -69,6 +69,14 @@ public struct URLEncoding: ParameterEncoding {
     /// - httpBody:        Sets encoded query string result as the HTTP body of the URL request.
     public enum Destination {
         case methodDependent, queryString, httpBody
+
+        func encodesParametersInURL(for method: HTTPMethod) -> Bool {
+            switch self {
+            case .methodDependent: return [.get, .head, .delete].contains(method)
+            case .queryString:     return true
+            case .httpBody:        return false
+            }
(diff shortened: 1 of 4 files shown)


Source/ParameterEncoding.swift
```diff
@@ -159,7 +164,7 @@ public struct URLEncoding: ParameterEncoding {
 
         guard let parameters = parameters else { return urlRequest }
 
-        if let method = HTTPMethod(rawValue: urlRequest.httpMethod ?? "GET"), encodesParametersInURL(with: method) {
+        if let method = HTTPMethod(rawValue: urlRequest.httpMethod ?? "GET"), destination.encodesParametersInURL(for: method) {
```
jshier: We now have an extension on `URLRequest` which should return the method, without having to initialize it.

cnoon: Good catch, addressed in 9ea95f6f.

Issue #2752, opened 2019-03-17
[Alamofire 5]: Remove Our Result Type and Require Swift 5


jshier:
We finally have a `Result` type in Swift. Supporting it, while straightforward, will require a bit of work. Most likely we'll need to `typealias` our current result signature as `AFResult<T>` and replace all of our usage with it. At the same time it seems like a good idea to start requiring Swift 5 to reduce our maintenance requirements going forward.

Given that Swift's `Result` is far more limited than ours, we'll need to reevaluate the APIs we ship with it. We can use the namespace from #2750 to ship some functionality, we just need to decide what.

AtomicCat:
Any objections to me opening a PR for this that builds on #2758?

jshier:
No, feel free.

AtomicCat:
Tagging @cnoon for completness...

jshier:
As an aside, creating `typealias AFResult<T> = Result<T, Error>` helped with my previous experiments in this transition. Create the `typealias`, do a find replace for all of our `Result<T>` usage, and things should get closer to building.

cnoon:
Go for it @AtomicCat! @jshier and I were both hoping we could swap this in AF5.

chrisvoronin:
Yep just ran into this conflict today as upgraded to Swift 5.0

jshier:
Completed in #2774.

PR #2769, opened 2019-03-27
Feature/swift result

AtomicCat:
### Issue Link :link:
#2752

### Goals :soccer:
This PR removes the custom `Result<Value>` type and replaces it with a typealias to the new Swift 5 Result type: `typealias AFResult<T> = Result<T, Error>`.

The existing extensions on `Result` are also updated to work with `AFResult`

In addition, all internal references to `Result` have been replaced with references to `AFResult`

The goal was to simply replace the existing `Result` type and retain the extended functionality as well; this would remove a now redundant type and keep the existing functionality that the code relied upon.

However in the process of implementing the changes I've come to realize this may not be the best course of action.

- Ideally the extensions on `AFResult` would be namespaced behind an `af` sub-struct as outlined in #2750. However, I was unable to come up with any Swift syntax that actually compiled as expected due to the generic nature of `Result`. Perhaps someone else will have better luck than I.

- In preserving the existing extensions which the Alamofire code depends on, there are impedance mismatches with the new Swift `Result` type. For example, both implementations have a `flatMap` and `flatMapError` functions, but the semantics and the signatures of the supplied closures differ. The AF implementations take a closure that yields the `Success` or `Failure` types and can `throw`, whereas the Swift variants yield a new `Result` and cannot throw. There is existing code that relies on the old behaviors, so simplifying to the Swift versions is potentially non-trivial.

- A number of the existing `Result` extensions were explicit in the docstrings about returning an unmodified `Result` (e.g. `return self`), but it's not obvious why that was called out. I was unable to bridge the type mismatches with the new code and in many cases it now returns new `AFResult` instance with the expected value.

- Given the lack of extension namespacing, it is unclear if publicly extending the Swift `Result` is a good idea. While, switching to an `internal` access level would allow the existing code to function as-is, it would break any client code relying on the old functionality.

### Testing Details :mag:
The unit tests for `AFResult` pass, though I have a lot of random failures on other unrelated tests. The `master` branch also has a number of similar failures. The corporate network I'm on is pretty locked down, so that could be a factor.


AtomicCat:
I thought this would be easy <ha>, but I think it's revealed that some architectural/philosophical decisions need to be made by people closer to the project than I am.

jshier:
You'll need to merge master to fix the conflicts. (Fun!)

If we can't namespace the extensions I would just get rid of them and use our `typealias` to cover the API. There are few internal bits that will need to change but I'm not concerned about breaking `Result` usage. It's unfortunate but necessary. We could manually prefix the methods with `af_` but I think I'd want to remove everything first and see what we want to keep.

cnoon:
I agree @jshier. I think we should remove public extensions on `Result` so that we only vend AFResult as a typealias with no additional functionality, and keep any internal extensions that are necessary. I'd also hope though we'd work to remove the dependence internally on old, extended APIs on `AFResult` and instead move over to the new APIs that we have available.

Source/AFResult.swift
```diff
@@ -0,0 +1,247 @@
...
+    /// success or failure in addition to the value or error.
+    public var debugDescription: String {
+        switch self {
+        case .success(let value):
+            return "SUCCESS: \(value)"
+        case .failure(let error):
+            return "FAILURE: \(error)"
+        }
+    }
+}
```
cnoon: We'll probably want to do this a different way using internal APIs. Otherwise we are really polluting the Result behavior publicly. Not something that you need to solve in this PR @AtomicCat.

Source/AFResult.swift
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Returns the associated error value if the result is a failure, `nil` otherwise.
+    public var error: Error? {
+        switch self {
+        case .success:
+            return nil
+        case .failure(let error):
+            return error
+        }
+    }
```
cnoon: Hopefully in a future PR we can get rid of all these extensions across the board. At a minimum, get them moved to internal APIs.

Source/AFResult.swift
jshier: We shouldn't override these at all, as it was decided the compiler generated strings are good enough when `Result` was added to the standard library.

jshier:
Ultimately I think we want to remove all extensions except those we use internally, those too if we can replace them with functionality from `Swift.Result`. When I looked at this it was only the properties that didn't have an easy equivalent, so you can cut out everything except that which breaks the build and we can reexamine.

Source/AFResult.swift
AtomicCat: The `debugDescription`s of the Response types rely on these extensions. The Swift Result doesn't appear to conform to either of the string convertible protocols.

jshier: Right. Rather than accessing the property directly, we can use string interpolation to get the description of the value. That would match what users expect for the type’s description.

Source/AFResult.swift
AtomicCat: I've changed the access level to `internal`.

Source/AFResult.swift
cnoon: Here's one way that could look:

```swift
extension String.StringInterpolation {
    mutating func appendInterpolation(afResult value: AFResult, verbose: Bool = false) {
        if verbose {
            switch self {
	        case .success(let value): return "SUCCESS: \(value)"
	        case .failure(let error): return "FAILURE: \(error)"
	    }
        } else {
            switch self {
	        case .success: return "SUCCESS"
	        case .failure: return "FAILURE"
	    }
        }
    }
}
```

jshier: Sorry, I didn't mean we should do anything custom here, just our current uses of `result.description` or `result.debugDescription` could just be replaced by `"\(result)". Unless there's an important bit of data we want to provide, but I don't think there is.

Source/AFResult.swift
jshier: No need for an explicit access level, just don't make the `extension` `public`.

Source/AFResult.swift
```diff
@@ -0,0 +1,247 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+public typealias AFResult<T> = Result<T, Error>
```
ejensen: Can the `Error` type constraint be made the more specific `AFError` type? 

I believe Alamofire always vends `AFError`s masked behind `Error`. 
Which currently requires receivers to implicitly know that they should using the `error.asAFError` extension to get the detailed error type.

The switch to explicitly specifying the `Failure` case as `AFError` would eliminate the need for the extension on `Error` and the implicit knowledge of its existence

jshier: Not really, no, as we don't convert underlying system error to our error type in most situations. It would also require us to wrap all user errors as well, as we don't wrap things like serializer failures in our own error type either. And even if we did wrap everything in `AFError`, the user would still need to extract the underlying associated error as an `Error` and then cast to the type they need. So I'm not sure adding a layer of indirection through our error type would help.

cnoon: This is something we do in our error system built on top of AF @jshier and it's actually super helpful. This way, you can easily categorize the error types you care about. We'd probably need to introduce something like `.user` and `.system` cases into `AFError` that would wrap the underlying error. That would allow us to lock `AFResult` to `AFError` exclusively and would make it easier for clients, even though it makes it more difficult for us in AF. 

Worth considering...

ejensen:
> If we can't namespace the extensions I would just get rid of them and use our `typealias` to cover the API. There are few internal bits that will need to change but I'm not concerned about breaking `Result` usage. It's unfortunate but necessary. We could manually prefix the methods with `af_` but I think I'd want to remove everything first and see what we want to keep.

I took a shot at namespacing the extensions under `Result.af` in commit https://github.com/Alamofire/Alamofire/commit/db8b23198e94f602611a01bee2836134b6a659cc
It doesn't use the `AlamofireExtended` protocol since the generics were fighting each other. But it results in the same signature & functionality.

Commit https://github.com/Alamofire/Alamofire/commit/d38c4e979a4877a96186cd41b6b5d5df4d34a899 updates the Documentation to show how `internal`-izing the extensions affect the public API.

cnoon:
Ah sorry @ejensen, I should have posted here that I've been working on that as well. The current gameplan is to move all the APIs on `AFResult` to internal, and remove all of them we can. I've been able to remove most of them since some are duplicates, but there are a few that are worth keeping for internal purposes. 

The biggest question is whether we can to keep the `value` and `error` convenience properties. Right now we're leaning towards keeping them internal, and clients can easily add those extensions if they want them. Interested to get your thoughts on the matter as to which ones you'd like to still be public (probably under an `af` namespace).

Source/AFResult.swift
jshier: Yeah, thinking about it more, the doubly generic `Result` does give us more flexibility here. I just need to think out all of the ramifications of encapsulating all errors.

cnoon:
Thanks for all the leg-work here @AtomicCat and @ejensen! I've pulled all these changes into #2774 while maintaining both of your attributions. Let's continue the discussion on that PR. I'm going to close this one out for now.

PR #2771, opened 2019-03-27
Swift 5!!!

cnoon:
This PR updates the two Xcode projects, source code, test code, SPM package, and podspec to Swift 5. I have a little demo using SPM verifying things are working, so we should be good on that front.

### Goals :soccer:
Move the entire project onto Xcode 10.2 and Swift 5 only.

Migrator and package verification.

### Testing Details :mag:
Just requires the test suite to pass. No new tests are needed. I had to fix up some user agent tests and encoder tests where the key-value pairs were no longer sorted.


cnoon:
Just realized I missed the travis updates. Will fix.

Tests/ParameterEncoderTests.swift
```diff
@@ -780,3 +780,10 @@ private struct FailingOptionalStruct: Encodable {
         }
     }
 }
+
+private func XCTAssertQueryEqual(_ query1: String?, _ query2: String?) {
```
jshier: Nice!

Issue #2770, opened 2019-03-27
default emptyResponseAllows returns false unless HTTP Method is .head 


jvannoord:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

I'm deserializing a response from an HTTP PUT Method and returns an empty status code (204).

## What did you expect to happen?

My instinct was that ResponseSerializer.emptyResponseAllowed(forRequest:response:) would return true if either the http method was HEAD or the status code was 204/205. 

If I were to have written this method, it would have looked like this:

`requestAllowsEmptyResponseData(request) == true || responseAllowsEmptyResponseData(response) == true`

## What happened instead?

It appears that in the default case, the code returns true only if the http method is HEAD AND the status code to be 204/205. Below is the current code.

`requestAllowsEmptyResponseData(request) ?? responseAllowsEmptyResponseData(response) ?? false`

Again, I'm merely asking if this is logical AND is intentional or not.

The workaround is easy: just need to explicitly enumerate all of the HTTP methods that can return empty responses.

## Alamofire Environment

**Alamofire version:** 5.0.0-beta3
**Xcode version:** 10.1 / 10.2
**Swift version:** 4.2
**Platform(s) running Alamofire:** iOS

jshier:
My first thought is that this is a bug from being too clever during cleanup of these methods. It should certainly be what you've put. Seems like we'll need a fix and some testing around this.

jshier:
@jvannoord Give the `bug/2770-empty-response-logic` branch a shot, that should fix it. We'll roll this into the next beta, I just need to write some tests for it. Thanks for the report!

jvannoord:
Whoa, just like that! Thanks! I was gonna volunteer, but you. are. on. it.

ruslankaminsky:
I've just encountered this issue today morning, after updating from v4.8 to v5 beta 3. And noticed that server responses with code 204 and empty body started to be handled as errors with "Alamofire.AFError.ResponseSerializationFailureReason.inputDataNilOrZeroLength" error. 
I'm happy that it will be fixed soon.

cnoon:
This issue should be resolved in the 5.0.0-beta.4 release. Please test it out and let us know!

Cheers. 🍻

Issue #2270, opened 2017-09-11
Issue with Alamofire + AFNetworking


mtr1012:
I'm working on a project using both Objective C and Swift 3. We use both Alamofire and AFNetworking for network request. 

Dev Environment :
 'AFNetworking', '3.1.0'
 'Alamofire', '4.4.0'

My issue is: When a request (request 1) make with Alamofire is in progress (uploading file), i excute an another request with AFNetworking (request 2). Request 2 not excute immediately until request 1 finish. When request 1 finish request 2 will trigger automatically

Can anyone give me some idea on this issue? 

Many thanks !

mtr1012:
Please helppppppp

jshier:
Are you using the same `URLSession` between both AFNetworking and Alamofire? If you, you may just be running into system throttling for overall network connections, or other `URLSession` details. I suggest you read [Apple's `URLSession` documentation](https://developer.apple.com/documentation/foundation/urlsession) to see what may be happening. 

Also, we use our GitHub project for bug reports and feature requests. In the future, you should open questions like this on [Stack Overflow](https://stackoverflow.com) and tag `alamofire`.

Cheers. 🍻

---

From our [Contribution Guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md)

## Asking Questions

We don't use GitHub as a support forum. For any usage questions that are not specific to the project itself, please ask on [Stack Overflow](https://stackoverflow.com) instead. By doing so, you'll be more likely to quickly solve your problem, and you'll allow anyone else with the same question to find the answer. This also allows maintainers to focus on improving the project for others.

mtr1012:
i'm not using the same URLSession

Swift: i'm using background configuration
let configuration = URLSessionConfiguration.background(withIdentifier: bundleIdentifier)

// Using background session instead of default URLSession

let sessionManager = Alamofire.SessionManager(configuration: configuration)

Obj C: i'm using default session
AFHTTPSessionManager *manager = [AFHTTPSessionManager manager];

please help @jshier

PR #2772, opened 2019-03-28
[#2270]: Fix emptyResponseAllowed logic.

jshier:
### Issue Link :link:
#2770 

### Goals :soccer:
This PR fixes the `emptyResponseAllowed` logic to no longer use nil coalescing but direct value comparison instead.

### Testing Details :mag:
Needs tests, so this will be a draft PR until they're written.


jshier:
@cnoon Ready for review!

Source/URLRequest+Alamofire.swift
```diff
@@ -26,7 +26,7 @@ import Foundation
 
 public extension URLRequest {
     var method: HTTPMethod? {
-        guard let httpMethod = httpMethod else { return nil }
-        return HTTPMethod(rawValue: httpMethod)
+        get { return httpMethod.flatMap(HTTPMethod.init) }
+        set { httpMethod = newValue?.rawValue }
     }
```
cnoon: Not that we need to address this in this PR, but we should consider always returning a `.get` and not having this be optional. It defaults to `.get` even when not set.

Tests/ResponseSerializationTests.swift
```diff
@@ -285,20 +324,59 @@ final class DataResponseSerializationTestCase: BaseTestCase {
             XCTFail("error should not be nil")
         }
     }
-
-    func testThatStringResponseSerializerSucceedsWhenDataIsNilWithEmptyResponseStatusCode() {
+
```
cnoon: Looks like some tabs snuck in throughout this file.

Source/URLRequest+Alamofire.swift
jshier: Huh. Not only does it default to `GET` but even after setting it to `nil` it still returns `GET`. What weird API. I can fix.

jshier: Thinking about it more, since we only support a strict set of methods, it seems like this extension **should** return `nil` when the user has set a method string to a value we don't support.

cnoon: Oh great point! I really don't like the Apple API in this case.

cnoon: What does it even do if you set it to `INVALID` or something? Does it just fallback to `GET`?

cnoon: If it does just fall back to `GET`, then we probably can too.

jshier: Nope, it returns the `String` entered, whatever it is.

cnoon: 😩

cnoon: Optional it is 🤨

PR #2774, opened 2019-03-28
AFResult

cnoon:
This PR is a combination of #2769 from @AtomicCat and the changes in https://github.com/Alamofire/Alamofire/commit/db8b23198e94f602611a01bee2836134b6a659cc from @ejensen with changes from myself with what should be exposed.

### Issue Link :link:
#2752 

### Goals :soccer:
There are a few:

- Replace the `Alamofire.Result` type with the `Foundation.Result` type build by our own @jshier!
- Remove any `public` extension APIs that duplicate the new `Foundation.Result` APIs.

The main details of this implementation are described in-depth in #2752.

### Testing Details :mag:
The test suite has been updated accordingly.

> NOTE: I'll swing back here and add more detail to this PR shortly. _have to run to a meeting_


Source/Request.swift
```diff
@@ -781,7 +781,7 @@ open class DataRequest: Request {
 
             let result = validation(self.request, response, self.data)
 
-            result.withError { self.error = $0 }
+            if case .failure(let error) = result { self.error = error }
```
jshier: Why not keep `withError` (and `withValue`) as extensions?

Source/AFResult.swift
```diff
@@ -0,0 +1,125 @@
...
+        let result: Result<Success, Failure>
+    }
+
+    public static var af: AFExtension.Type { return AFExtension.self }
+    public var af: AFExtension { return AFExtension(result: self) }
+}
+
+// MARK: - Public APIs
+
+extension Result.AFExtension {
```
jshier: The public / internal distinction here seems arbitrary. Why don't we keep anything we want to make public namespaced, and any other extensions we need should be internal and non-namespaced? So we could just mark this extension public and move anything internal to the `AFResult` extension.

Source/Response.swift
```diff
@@ -42,13 +42,13 @@ public struct DataResponse<Value> {
     public let serializationDuration: TimeInterval
 
     /// The result of response serialization.
-    public let result: Result<Value>
+    public let result: AFResult<Value>
 
     /// Returns the associated value of the result if it is a success, `nil` otherwise.
-    public var value: Value? { return result.value }
+    public var value: Value? { return result.af.value }
```
jshier: Might we want to get rid of this API and have users pull from the result themselves?

Source/AFResult.swift
AtomicCat: If the internal extensions are only needed to support unit tests, moving those additions to the test target might be a good idea.

Source/Response.swift
AtomicCat: If the new intent is to not provide functionality beyond what Apple thought appropriate for Result, then removing this would be inline with that. It depends on if you think providing a shortcut or bandaid over Result is appropriate. 

I'd guess that someone who was unhappy with the Result API would already have their own extension that added `value` to Result and this would be unnecessary.

Source/AFResult.swift
```diff
@@ -0,0 +1,125 @@
...
+    /// Returns the associated value if the result is a success, `nil` otherwise.
+    public var value: Success? {
+        guard case .success(let value) = result else { return nil }
+        return value
+    }
+
+    /// Returns the associated error value if the result is a failure, `nil` otherwise.
+    public var error: Error? {
```
ejensen: Can the return type be the generic type `Failure` instead of `Error`? It would return the specific type that implements `Error` in cases where the `Failure` generic type is more specific than `Error`

Tests/AFResultTests.swift
```diff
@@ -0,0 +1,185 @@
...
+        // Given, When
+        let result = AFResult<String>.failure(error)
+
+        // Then
+        XCTAssertNotNil(result.af.error, "result error should not be nil for failure case")
+    }
+
+    // MARK: - Initializer Tests
+
+    func testThatInitializerFromThrowingClosureStoresResultAsASuccess() {
```
ejensen: The following tests cover `Swift.Result` methods instead of anything in Alamofire. Candidates for deletion?
```
testThatInitializerFromThrowingClosureStoresResultAsASuccess
testThatInitializerFromThrowingClosureCatchesErrorAsAFailure
testThatFlatMapTransformsSuccessValue
```

Source/Request.swift
cnoon: We could, but we already have a Swift construct that is essentially the equivalent. I was trying to remove all the extensions that we could replace with either common Swift API or Result API. I'm not opposed to bringing it back, but I would advise against making it a public API if we do bring it back.

Source/AFResult.swift
cnoon: That's exactly what I did @jshier. I deleted any of them we don't actually need, kept the ones public in the namespaced extension, and moved the initializer to be an internal API on `AFResult`. Are you referring to what I did in the test suite by adding internal API behind the namespace? I could certainly move those to just extend `AFResult`.

cnoon: And @AtomicCat, the initializer at the bottom of the `AFResult` file is needed in the framework as well as the test target.

Source/Response.swift
cnoon: We use this as a convenience all over the place in the test suite. I think it would be unwise for us to make the assumption that the public doesn't need this, b/c they will. The question is just whether we provide it through an `af` namespace, or we require them to extend `Result` themselves. In the best practice mindset, I'd lean towards removing it. In the pragmatic sense, I think everyone else will add it if we don't provide it.

Tests/AFResultTests.swift
cnoon: Ah thanks, I didn't look closely enough. Will fix.

Source/Response.swift
jshier: That's not what I meant. If we offer public `Result` extensions for accessing the associated values, we don't need to offer these properties on `Response`, since users could just access `response.result.af.value`.

Source/AFResult.swift
cnoon: But that won't be the case with Alamofire as is would it since the `AFResult` is locked to `Error` currently? Even if it was an AFError, you'd still have to cast it wouldn't you?

Source/AFResult.swift
jshier: My point was that any extension that isn't public doesn't need to be namespaced and so can be moved directly into the `AFResult` extension four our internal use. If we want to offer any public extensions, we should discuss which features those would be. I think that extensions here could all be public, and so the extension should be made public. I've seen `isSuccess` / `isFailure` misused quite a bit in the wild, so keeping those internal would be good.

Source/AFResult.swift
jshier: This is an extension on `Result`, which is `Result<Success, Failure: Error>`, so we should use the generic here so the extension is useful on all `Result`s.

cnoon: Oh I follow now, good point. I guess I was assuming people would use the `af` namespace on non-AFResult types.

Source/Request.swift
cnoon: We decided in Slack to remove them for now.

Source/AFResult.swift
cnoon: We decided to remove the public extensions altogether in Slack.

Source/AFResult.swift
```diff
@@ -0,0 +1,108 @@
...
+
+extension AFResult {
+    /// Returns the associated value if the result is a success, `nil` otherwise.
+    var value: Success? {
+        guard case .success(let value) = self else { return nil }
+        return value
+    }
+
+    /// Returns the associated error value if the result is a failure, `nil` otherwise.
+    var error: Error? {
```
jshier: This should be `Failure`, even for our internal use.

Source/AFResult.swift
```diff
@@ -0,0 +1,108 @@
...
+    ///     let possibleData: AFResult<Data> = .success(Data(...))
+    ///     let possibleObject = possibleData.flatMapError {
+    ///         try someFailableFunction(taking: $0)
+    ///     }
+    ///
+    /// - Parameter transform: A throwing closure that takes the error of the instance.
+    ///
+    /// - Returns: An `AFResult` instance containing the result of the transform. If this instance is a success, returns
+    ///            the same success.
+    func flatMapError<T: Error>(_ transform: (Failure) throws -> T) -> AFResult<Success> {
```
jshier: Do we use this internally?

Tests/AFResultTests.swift
cnoon: Deleted the entire set of tests.

Source/AFResult.swift
cnoon: We decided not to vend any APIs publicly so the point is moot.

Tests/AFResult+Alamofire.swift
```diff
@@ -0,0 +1,47 @@
...
+    var isFailure: Bool {
+        return !isSuccess
+    }
+
+    var value: Success? {
+        guard case .success(let value) = self else { return nil }
+        return value
+    }
+
+    var error: Error? {
```
jshier: Again, we can keep the `Failure` generic, so if we do ever vend `Result<T, AFError>`, we'll be ready.

Source/Request.swift
cnoon: We decided to use the Swift conventions for now and remove the `withValue` and `withError` APIs.

Source/AFResult.swift
cnoon: Yep, we do in the functional response APIs a couple of times. I figured we'd want to keep the extension rather than duplicate the logic since it's complex.

jshier: Only usage I can see for `flatMapError` ATM is in the tests.

Source/AFResult.swift
cnoon: Fixed in 456847e2.

Tests/AFResult+Alamofire.swift
cnoon: Fixed in 456847e2.

Source/AFResult.swift
jshier: NVM, I had search limited to tests.

PR #2775, opened 2019-03-28
Compose `Session` of several protocols

svenmuennich:
### Goals :soccer:
The main goal of this PR is to make mocking `Session` easier for unit testing by moving most of its public methods and properties to new protocols and extensions while maintaining the current public API (`5.0.0-beta.4`).

I added four new protocols
1. `RequestSession`,
2. `DownloadSession`,
3. `UploadSession` and
4. `SessionProtocol` (not sure about the name here).

The first three consist of methods previously part of the `Session` class, all copied without modifying the declarations to keep the public interface the same (actually, one method is added). The methods where split into protocols as follows:

#### `RequestSession`

* `request(_:, method:, parameters:, encoding:, headers:, interceptor:)`
* `request(_:, method:, parameters:, encoder:, headers:, interceptor:)`
* `request(_:, interceptor:)`

#### `DownloadSession`

* `download(_:, method:, parameters:, encoding:, headers:, interceptor:, to:)`
* `download<Parameters:>(_:, method:, parameters:, encoder:, headers:, interceptor:, to:)`
* `download(_:, interceptor:, to:)`
* `download(resumingWith:, interceptor:, to:)`

#### `UploadSession`

* `upload(_:, to:, method:, headers:, interceptor:)`
* `upload(_:, with:, interceptor:)`
* `upload(_:, to:, method:, headers:, interceptor:)`
* `upload(_:, with:, interceptor:)`
* `upload(_:, to:, method:, headers:, interceptor:)`
* `upload(_:, with:, interceptor:)`
* `upload(multipartFormData:, usingThreshold:, fileManager:, to:, method:, headers:, interceptor:)`
* `upload(multipartFormData:, usingThreshold:, with:, interceptor:)`
* `upload(_:, with:, interceptor:)`
* `upload(_:, interceptor:)` (**new in public interface**)

`SessionProtocol` conforms to the three protocols described above and also declares some properties, which I think are not implementation specific but should be available to consumers of `SessionProtocol`s:

* `var session: URLSession { get }`
* `var delegate: SessionDelegate { get }`
* `var interceptor: RequestInterceptor? { get }`
* `var eventMonitor: CompositeEventMonitor { get }`
* `var defaultEventMonitors: [EventMonitor] { get }`

`Session` now conforms to `SessionProtocol` (and as a result also to `RequestSession`, `DownloadSession` and `UploadSession`), keeping the same public interface as before.

Most methods declared in the new interfaces have a default implementation provided by protocol extensions, since they just _forward_ their calls to other methods of the same protocol, leaving only 4 methods (plus the properties) to be implemented when adopting `SessionProtocol` in a custom implementation:

* `request(_:, interceptor:)`
* `download(_:, interceptor:, to:)`
* `download(resumingWith:, interceptor:, to:)`
* `upload(_:, interceptor:)`

The only _breaking change_ I needed to make was changing the nested type `Session.RequestConvertible` to be a top level type `RequestConvertible`.

### Testing Details :mag:
So far I haven't added any tests because I first wanted to get some feedback on the general idea of this PR. Also, since neither the public interface nor the inner workings of `Session` are changed by this PR, I don't think any new tests are necessary.


svenmuennich:
Please note that I haven't rebased this PR on `5.0.0-beta.4` since I'm still stuck with Xcode `10.1`. I will be happy to take care of this if you decide this PR should be merged :)

svenmuennich:
Do you have any feedback on this one? It would be great to get this merged in time for the release of `v5.0.0` 🙂

jshier:
@svenmuennich 

While we've briefly discussed this PR, we have yet to reach a final decision. Since these changes are additive and wouldn't affect users of existing Alamofire APIs, we could ship these protocols at any time, so they wouldn't be included in the 5.0 release either way. 

However, personally, I don't think something like this should ship with Alamofire, for both general and Alamofire-specific reasons.

In general, I don't believe protocol-abstraction is a great way to test things for a few reasons. First and foremost, it's not a scalable approach to testing. If you abstract every API boundary using protocols, you double the number of types required for your entire project. Protocol abstractions also tend to require every other type they use to also be protocol abstracted so that everything can be controlled at test time. A type which has a few dependencies injected at initialization would likely need at least as many protocols to be properly abstracted. This is a huge burden on the tools, on the developers, and on the project itself due to the long term maintenance cost. 

Second, it actually prevents the testing of important application characteristics. For me, it's better for tests to be as realistic as possible. Protocol abstraction prevents this realism and so can prevent very real issues from becoming visible, particularly for something like networking. In this regard I think it's similar to testing asynchronous APIs by making them synchronous. Technically it work, but you're not testing the things that are actually happening in your app.

Finally, the complexity of protocol abstracted tests can sometimes approach the complexity of using the real thing, so using the real thing would be a better choice. I think this will be especially true of anything trying to use the protocols you've proposed, given the necessity of reproducing various internal Alamofire types for use in the protocols. For instance, the various `Request` types aren't very useful on their own, they have to be produced and have lifetime events managed by `Session`. Reproducing that in your protocol abstraction will likely be significant work.

So, in general, I only use protocols abstractions to work around systems that have side effects I can't otherwise control. Things like disk or keychain storage, or perhaps view controller presentation (though I have yet to find a good approach there). For networking there are a variety of existing libraries that use `URLProtocol` or customized `URLSession` subclasses to enable testing. I find these approaches superior across the board, as they are both far more realistic and less of a burden on my projects to use.

As to Alamofire in particular, again there are a few reasons I wouldn't want this sort of abstraction built into the library. First, like I mentioned before, it's a maintenance burden. Every API change we want to make would have to be done twice, once for real and once in the protocols and their underlying testable implementation. There are only two core contributors to this project and we do this as our night job, so something that makes it harder for us to develop the library and requires more support in the long term is not something likely to be added.

Second, on a similar note, Alamofire already works with `URLProtocol` and specifically supports initializing a `Session` with a specific `URLSession` instance  in order to support testing. Both of these methods are far less of a burden on the library and far more realistic, which makes them superior testing methods to me. 

Third, users really shouldn't need to abstract all of Alamofire. Any even slightly sophisticated user of the library should already have their own internal generic abstraction they use to make requests, so abstraction should be as simple as abstracting that generic method. Even if they don't write it themselves, it seems like most users could get away with abstracting one of Alamofire's `request` methods, making the rest of the protocol abstractions unnecessary.

So while this is an interesting approach, I'm not sure it's a good fit for the library as a whole. However, it may be valuable as a separate dependency on top of Alamofire, so I encourage you to build out enough of it to achieve your goals. And if there are improvements we can make to Alamofire to make it more testable in some ways, or easier to abstract, let us know!

svenmuennich:
Thank you very much for taking the time to give such a thorough response. I have worked with Alamofire 5 for several month now **without my proposed changes** and it all worked out perfectly.

And sorry for coming back to this just now. This should have been closed a long time ago.

PR #2778, opened 2019-03-30
Bugfix - Cancellation Loop

cnoon:
This PR fixes an issue in the response serialization retry logic where calling `cancel` and a few other APIs would result in the response serializers running again.

### Goals :soccer:
Fix the issue where calling `request.cancel()` inside response serializer would result in the response serializer running again.

I should have realized this when I was building it, but it came up when moving our networking library onto AF5 and running the test suite over it. When we finish running all the response serializers and determine that we do not need to retry the request, we then execute all the response serializer completions sequentially. However, we don't exactly know what are in those completions. Therefore, we have to assume the worst and that they call back into the request to possibly do things like `request.cancel()`. I actually had a few unit tests doing this and found some dragons.

Solution is to clear out all the response serializers and completions from the mutable state prior to actually executing the completions. This is how the logic should have been implemented in the beginning.

### Testing Details :mag:
I added a test that fails without this change demonstrating the issue.

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -1104,7 +1104,6 @@
 			developmentRegion = en;
 			hasScannedForEncodings = 0;
 			knownRegions = (
-				English,
 				en,
 				Base,
 			);
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -396,8 +396,19 @@ open class Request {
     /// Processes the next response serializer and calls all completions if response serialization is complete.
     func processNextResponseSerializer() {
         guard let responseSerializer = nextResponseSerializer() else {
-            // Execute all response serializer completions
-            protectedMutableState.directValue.responseSerializerCompletions.forEach { $0() }
+            // Execute all response serializer completions and clear them
+            protectedMutableState.write { mutableState in
+                let completions = mutableState.responseSerializerCompletions
+
+                // Clear out all response serializers and response serializer completions in mutable state since the
+                // request is complete. It's important to do this prior to calling the completion closures in case
+                // the completions call back into the request triggering a re-processing of the response serializers.
+                // An example of how this can happen is by calling cancel inside a response completion closure.
+                mutableState.responseSerializers.removeAll()
+                mutableState.responseSerializerCompletions.removeAll()
+
+                completions.forEach { $0() }
+            }
 
             // Cleanup the request
             cleanup()
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -397,8 +397,10 @@ open class Request {
     func processNextResponseSerializer() {
         guard let responseSerializer = nextResponseSerializer() else {
             // Execute all response serializer completions and clear them
+            var completions: [() -> Void] = []
+
             protectedMutableState.write { mutableState in
-                let completions = mutableState.responseSerializerCompletions
+                completions = mutableState.responseSerializerCompletions
 
                 // Clear out all response serializers and response serializer completions in mutable state since the
                 // request is complete. It's important to do this prior to calling the completion closures in case
@@ -406,9 +408,9 @@ open class Request {
                 // An example of how this can happen is by calling cancel inside a response completion closure.
                 mutableState.responseSerializers.removeAll()
                 mutableState.responseSerializerCompletions.removeAll()
-
-                completions.forEach { $0() }
             }
+
+            completions.forEach { $0() }
 
             // Cleanup the request
             cleanup()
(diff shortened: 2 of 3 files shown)


Source/Request.swift
```diff
@@ -396,8 +396,19 @@ open class Request {
...
+                let completions = mutableState.responseSerializerCompletions
+
+                // Clear out all response serializers and response serializer completions in mutable state since the
+                // request is complete. It's important to do this prior to calling the completion closures in case
+                // the completions call back into the request triggering a re-processing of the response serializers.
+                // An example of how this can happen is by calling cancel inside a response completion closure.
+                mutableState.responseSerializers.removeAll()
+                mutableState.responseSerializerCompletions.removeAll()
+
+                completions.forEach { $0() }
```
cnoon: These should not be called inside the lock. Need to pull the completions out of the lock and then execute.

Alamofire.xcodeproj/project.pbxproj
```diff
@@ -1104,7 +1104,6 @@
 			developmentRegion = en;
 			hasScannedForEncodings = 0;
 			knownRegions = (
-				English,
```
cnoon: This was an issue Xcode caught and wanted to fix so I made the change.

Source/Request.swift
cnoon: Addressed in 881e75fa.

Tests/SessionTests.swift
```diff
@@ -1435,6 +1435,43 @@ class SessionTestCase: BaseTestCase {
...
+        let request = session.request("https://httpbin.org/get", interceptor: handler)
+        request.validate()
+
+        request.responseJSON { resp in
+            request.cancel()
+
+            response = resp
+            completionCallCount += 1
+
+            DispatchQueue.main.after(0.01) { expectation.fulfill() }
```
jshier: Why is this delay necessary? Is it an attempt to finish the internal race to set the cancellation state? Also, how sensitive is this delay? We've had issues in the past with static delays being too short to reliably pass in the test environment.

Tests/SessionTests.swift
```diff
@@ -1435,6 +1435,43 @@ class SessionTestCase: BaseTestCase {
...
+        let expectation = self.expectation(description: "request should complete")
+        var response: DataResponse<Any>?
+        var completionCallCount = 0
+
+        // When
+        let request = session.request("https://httpbin.org/get", interceptor: handler)
+        request.validate()
+
+        request.responseJSON { resp in
+            request.cancel()
```
jshier: It seems like we may want to more generally fix `cancel()` by preventing its actions when the `Request` has already finished. It may be out of scope for this fix, but it makes sense to me that we may want to add an additional stage to state, perhaps `.completed`, to indicate that a `Request` has finished and started calling its completion handlers.

Tests/SessionTests.swift
cnoon: Great question. The reason the delay is necessary is to allow the `responseJSON` completion closure to be called a second time before the expectation is fulfilled. Without the delay, we'd have a race condition where the assertion on `completionCallCount` would most likely pass the assertion test before the retry logic actually ran the `responseJSON` completion closure again. In all my testing locally, this test will fail if we do not ship the provided fix. So that is good at least.

I personally do not like these types of tests that require waits to see if bad things happen, but unfortunately, this is one of those types of cases. We need a slight delay before we move on with the assertions to make sure additional calls aren't occurring. Thankfully, this will never produce false-positives if CI runs this particular test much slower. It would perhaps miss a regression if the delay was too short, but it would not produce a false-positive failure.

Tests/SessionTests.swift
cnoon: I agree and would advocate we consider that outside this particular PR. I think we need to spend some time thinking about the `resume`, `suspend`, and `cancel` APIs and the current safety around them. I think there are a lot of good additions to AF5 in this area, but I think they could still use another rev before we officially release.

jshier: Sounds good!

Tests/SessionTests.swift
jshier: Thinking about it again, I think we can do without the delay as, if the completion did fire again, the fulfillment count of the expectation would go up and the test would fail. I think that would probably be a bit cleaner than the delay, if you think it would work.

cnoon: It's cleaner, but way more racey, which is why I added the delay. By adding a delay, we know for certain that the code would execute a second time on a local machine which would produce the failure. Without the delay, it generally won't fail, depends on how fast the test can get shut down. My goal here was to produce a test that reliably fails with the current design flaw. Then fix the code so the test reliably passes.

PR #2779, opened 2019-03-30
Bugfix - Resumed Request State

cnoon:
This PR fixes an issue where the `Request.State` was never being set to `.resumed` when `startRequestsImmediately` is set to `true`.

### Goals :soccer:
When setting `startRequestsImmediately` to `true`, the request's state should reflect it's actual state.

There is a huge amount of complexity in this simple little method. I tried to actually document what's going on in the code because I thought it might be worthwhile to have. With the lack of thread-safety written in the state management in `Request` and the decoupling from the tasks themselves, we don't have any choice other than to run `resume` or `suspend` an extra time if the user is resuming the requests themselves. The `request.resume()` call is racing with the `Request` and `URLSessionTask` creation. Depending on which one wins, we can get into a scenario where `didResume` or `didSuspend` can be called twice even though only one `resume` or `suspend` call was actually made.

I don't really think this is an issue overall, it just required me to make some tests more robust once I realized these can be called more than once even if I only call them once. I don't think it's worth us trying to alter the entire threading model for this use case.

### Testing Details :mag:
I added a test to verify the state is now being set correctly when using `startRequestsImmediately`. I can't really write other reliable tests that will produce the race consistently. I do however have them in my networking test suite on top of AF and I can reproduce the race about 1 in 4 tries.

### DIFF
--- a/Source/Session.swift
+++ b/Source/Session.swift
@@ -471,12 +471,25 @@ open class Session {
     }
 
     func resumeOrSuspendTask(_ task: URLSessionTask, ifNecessaryForRequest request: Request) {
-        if startRequestsImmediately || request.isResumed {
+        if startRequestsImmediately && request.isInitialized {
+            // Resumes the request and ensures the state is updated to .resumed
+            request.resume()
+            return
+        }
+
+        // There's no way right now to avoid calling didResume or didSuspend twice in this case.
+        // The race is created by creating a request and calling resume (or suspend) on it right afterwards.
+        // Depending on conditions, you may end up resuming or suspending the task twice. The only way to guarantee
+        // this would only resume the task once would be to add thread-safety around the `resume`, `suspend`,
+        // and `cancel` APIs on Request. That thread-safety would also have to lock the actual calls to the
+        // task as well.
+
+        if request.isResumed && task.state != .running {
             task.resume()
             request.didResume()
         }
 
-        if request.isSuspended {
+        if request.isSuspended && task.state != .suspended {
             task.suspend()
             request.didSuspend()
         }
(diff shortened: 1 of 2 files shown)


Source/Session.swift
```diff
@@ -471,12 +471,25 @@ open class Session {
...
+        }
+
+        // There's no way right now to avoid calling didResume or didSuspend twice in this case.
+        // The race is created by creating a request and calling resume (or suspend) on it right afterwards.
+        // Depending on conditions, you may end up resuming or suspending the task twice. The only way to guarantee
+        // this would only resume the task once would be to add thread-safety around the `resume`, `suspend`,
+        // and `cancel` APIs on Request. That thread-safety would also have to lock the actual calls to the
+        // task as well.
+
+        if request.isResumed && task.state != .running {
```
jshier: Resuming an already `.running` task has no negative side effects, so are we just preventing the call to `request.didResume()` here? Same for `.suspended`? In that case, perhaps we should move the `did*` calls that trigger the event monitor into `Request` itself and have separate methods and events for the task actions. Something like `didResumeTask(_ task: URLSessionTask)` should work. That way the `Request` will only receive `didResume` whenever its state is actually updated (which should only be once no matter how many times `resume()` is called I think) but will receive `didResumeTask()` every time a task is resumed on its behalf. Same with the other state transitions.

cnoon: Yep, I think that makes a lot of sense. I think the `canTransitionToState` logic also needs to move to a testAndSet block instead of getting the value, comparing it, then setting it. It's currently not thread-safe and I've already been able to easily produce contention in a simple set of tests.

jshier: If you have tests that hit the thread sanitizer that's great. Do you think that this PR should include the lifetime reconsiderations that were mentioned in #2778? I think separating the `Request` and task lifetimes is a start, but like you said, we probably want to revisit whether `resume()`, `cancel()`, and `suspend()` are wholly async to keep them on the same internal queue, or whether there are parts we want synchronous but threadsafe. We'll probably want to make `canTransitionToState` safe in any event. I'll give it more thought.

cnoon:
Closing this PR out to defer to the more robust solution build in #2796.

PR #2783, opened 2019-04-04
Bugfix - Session Delegate Precondition

cnoon:
This PR removes the URL session delegate precondition check.

### Goals :soccer:
Users should be able to create `Session` instances alongside performance monitoring frameworks such as NewRelic that are swizzling our URLSessionDelegate APIs by proxying. 

While the session delegate precondition check is a good idea, enforcing the check causes Alamofire to be unusable with other frameworks that are swizzling out URL session delegate APIs. In cases where say Alamofire and NewRelic are being used within the same application, the precondition check will always fail.

Screen Shot 2019-04-02 at 5 34 42 PM

> The above example demonstrates what you get when you `po` the `session.delegate` and `delegate` inside the `precondition` removed in this PR. The `session.delegate` is an `NRMAURLSessionTaskDelegate` (_courtesy of NewRelic proxy swizzling_) and the `delegate` is an `Alamofire.SessionDelegate`.

I'm open to other ideas here as well. We could add a flag to disable the precondition that's `false` by default, but I thought it might be overkill to expose this precondition controls through the public API.

### Testing Details :mag:
N/A.

### DIFF
--- a/Source/Session.swift
+++ b/Source/Session.swift
@@ -54,8 +54,6 @@ open class Session {
                 redirectHandler: RedirectHandler? = nil,
                 cachedResponseHandler: CachedResponseHandler? = nil,
                 eventMonitors: [EventMonitor] = []) {
-        precondition(session.delegate === delegate,
-                     "SessionManager(session:) initializer must be passed the delegate that has been assigned to the URLSession as the SessionDataProvider.")
         precondition(session.delegateQueue.underlyingQueue === rootQueue,
                      "SessionManager(session:) intializer must be passed the DispatchQueue used as the delegateQueue's underlyingQueue as rootQueue.")
 


tariq235:
My 2 year old code started crashing on Alamofire 5. 
I am using AlamofireObjectMapper "6.0.0" which install Alamofire at "5.0.0-beta.2"

Screen Shot 2019-04-11 at 1 00 41 PM

hgado11:
> My 2 year old code started crashing on Alamofire 5.
> I am using AlamofireObjectMapper "6.0.0" which install Alamofire at "5.0.0-beta.2"
> 
> Screen Shot 2019-04-11 at 1 00 41 PM

Did you fix it?

saikishoreagrapalem:
Hi @cnoon 

i'm creating the framework which is like monitoring framework such as Newrelic and dynatrace. Could you please tell me how to monitor using URLSessionDelegate APIs

Thanks in advance

PR #2787, opened 2019-04-04
Update Readme with current release version

arslanbilal:


jshier:
Updating the `README` in this way on a regular basis is unnecessary, as the dependency managers all have specific syntax for ensuring you're using the latest prerelease version. Thanks anyway! Cheers! 🍻

Issue #2793, opened 2019-04-10
ServerTrustPolicy pinPublicKeys type check takes 500+ ms


radianttap:
In this code:

```swift
if certificateChainEvaluationPassed {
    outerLoop: for serverPublicKey in ServerTrustPolicy.publicKeys(for: serverTrust) as [AnyObject] {
        for pinnedPublicKey in pinnedPublicKeys as [AnyObject] {
            if serverPublicKey.isEqual(pinnedPublicKey) {
                serverTrustIsValid = true
                break outerLoop
            }
        }
    }
}
```

this particular line consistently takes over 500ms just for type-checking in Swift:

```swift
if serverPublicKey.isEqual(pinnedPublicKey) {
```

Casting to `AnyObject` is done purely in order to use `isEqual` call. Not sure are there more type-safe way to check for this equality? Those objects are `SecKey`, which is an empty class and has no `isEqual` implementation. Not sure is simple equality check between two SecKey instances enough to work properly:

```swift
if serverPublicKey == pinnedPublicKey {
```

radianttap:
If changed to this:

```swift
if certificateChainEvaluationPassed {
	outerLoop: for serverPublicKey in ServerTrustPolicy.publicKeys(for: serverTrust) {
		for pinnedPublicKey in pinnedPublicKeys {
			if serverPublicKey == pinnedPublicKey {
				serverTrustIsValid = true
				break outerLoop
			}
		}
	}
}
```

all tests in ServerTrustPolicyPinPublicKeysTestCase are passing.

PR #2794, opened 2019-04-10
Casting to AnyObject makes life hard for Swift type checker.

radianttap:
It takes about 500ms for the isEqual line, thus slowing down autocomplete and build times.

Using proper types types resolves this issue.

### Issue Link :link:
#2793

### Testing Details :mag:
Relevant test suite (ServerTrustPolicyPinPublicKeysTestCase) runs successfully.

### DIFF
--- a/Source/ServerTrustPolicy.swift
+++ b/Source/ServerTrustPolicy.swift
@@ -219,9 +219,9 @@ public enum ServerTrustPolicy {
             }
 
             if certificateChainEvaluationPassed {
-                outerLoop: for serverPublicKey in ServerTrustPolicy.publicKeys(for: serverTrust) as [AnyObject] {
-                    for pinnedPublicKey in pinnedPublicKeys as [AnyObject] {
-                        if serverPublicKey.isEqual(pinnedPublicKey) {
+                outerLoop: for serverPublicKey in ServerTrustPolicy.publicKeys(for: serverTrust) {
+                    for pinnedPublicKey in pinnedPublicKeys {
+                        if serverPublicKey == pinnedPublicKey {
                             serverTrustIsValid = true
                             break outerLoop
                         }


jshier:
Thanks for the PR!

This is interesting. Perhaps this implementation works now, as I definitely tried it when refactoring from Alamofire 4, but that was back with Swift 4 or 4.2. As long as we have test coverage on both the positive and negative side of the key pinning logic, I'm happy with this change.

@cnoon Any thoughts? This was an awful hack from long ago and if it's unnecessary now I think we'll be better off.

jshier:
@radianttap Just saw this is on the `hotfix` branch. That branch is only for Alamofire 4, which this change would not be part of, if only because it won't work in older Swift versions. Could you point the PR to master?

radianttap:
I did not follow the Alamofire development for a while, and it seems things in v5 are changing a lot. I need to do a separate PR for the master, right and keep this one for v4?

cnoon:
I would love to see this change make it into AF5. I don't even remember having to force that cast back when I wrote the original version, but if the Swift 5 compiler allows the comparison without having to force the bridging, then that would great! @radianttap if you could re-open a new PR against `master` that would be ideal. I'm not sure this is worth fixing in the AF4.x branch. I'm also not sure that it will be backwards compatible with older compilers.

cnoon:
@jshier I think I'd advocate for getting this change into AF5, and not trying to make it on AF4.x. Is that your take as well?

jshier:
Yes. Like you said, I'm pretty sure there will be compatibility issues with older compilers anyway.

radianttap:
Thanks for the discussion guys. I'm not sure what is the lowest supported Swift version for v4 but this change should compile just fine at least in Swift 4.2.

However, for master branch I don't think any change is needed, since the[ corresponding line](https://github.com/Alamofire/Alamofire/blob/6e3dec3a658be15c85b7966d53373f6371b17ae6/Source/ServerTrustEvaluation.swift#L292) is casting into AnyHashable, not AnyObject. And that does not incur any performance issues I can see.
I ran the tests with and without that casting and all seems fine. Thus if you want to get rid of the casting to AnyHashable, I'll make a PR - let me know.

BTW, I'm testing by adding `-Xfrontend -warn-long-expression-type-checking=50` as Debug custom flags for swift compiler (Other swift flags).

cnoon:
I think it would be great to have a PR against `master` that removes the `AnyHashable` casting @radianttap if you have time. Thanks for all your work on this! 🍻

PR #2796, opened 2019-04-11
Refactor Request Event Management

### LINKED ISSUE #2759: Cancel Callback not Received + EventMonitor Cancel Calls is Received Twice each time

## What did you do?
created a Download request with AF.download  and called cancel on the request
Receive Response Handler callback (to grab resumeData)
request.response ( .. )
## What happened instead?
Did not receive callback (although progress & success is being received) so that the handler is fine and not lost and Cancel callback is received twice on EventMonitor Side, I checked that on DownloadRequest, eventMonitor?.requestDidCancel(self) is being called only once.
## Alamofire Environment

**Alamofire version:*5*
**Xcode version:*10 (latest)*
**Swift version:*4.2*
**Platform(s) running Alamofire:*iOS 12.1*
**macOS version running Xcode:*10.14.3*

jshier:
### Issue Link :link:
This PR is an alternate take on #2779 and should fix #2759. 

### Goals :soccer:
This PR refactors how `Request`'s lifetime events are managed. 

This PR achieves greater safety and accuracy with a few changes:
- State transitions inside of `Request` have been made fully thread-safe through convenience API on `Protector`.
- `Session`'s implementation of `didResumeRequest`, `didCancelRequest`, `didCancelRequest(_:byProducingResumeData:)`, and `didSuspendRequest` have been refactored to separate the `Request` and `URLSessionTask` lifetime management, ensuring the `Request` methods are only called once.
- `Session`'s `resumeOrSuspend` method has been refactored and renamed to `updateStatesForTask(_:request:)`, which now handles all state combinations between `request.state` and `stateRequestImmediately` to ensure only the appropriate actions are taken in certain circumstances.

### Testing Details :mag:
Tests have been added to ensure that only the expect events happen in particular scenarios, no matter how many times `resume`, `suspend`, or `cancel` are called; that the events happen only the correct number of times; and that unexpected lifetime events don't happen.


Source/Session.swift
```diff
@@ -465,18 +465,25 @@ open class Session {
...
-        if request.isSuspended {
+            request.didResumeTask(task)
+        case (_, .suspended):
             task.suspend()
-            request.didSuspend()
+            request.didSuspendTask(task)
+        case (_, .cancelled):
+            task.cancel()
+            request.didCancelTask(task)
         }
```
cnoon: I think there is still a race here, but I don't think it ends up having any undesirable behavior from what I can tell. Stick with me here for a sec.

The `Request.resume`, `Request.suspend`, and `Request.cancel` APIs do not pipe through the `rootQueue` in any way. The transition logic runs inside the protector, but not inside the `rootQueue`. Therefore, the `Protector` seems to be the only thread-safety mechanism.

If this is the case (certainly possible it's not), then it's possible to have the following happen:

- `updateStatesForTask` is called on rootQueue and `request.state` if pulled from `Protecter`
- Another thread calls `Request.cancel()` and changes the state to cancelled
- The `updateStatesForTask` continues with a now out-of-date state (let's say it is initialized)
- The `request.resume()` method is called
- The attempt to transition to the `resumed` state will fail silently

If that's all correct, then there's still a race, but I don't think it has any side effects given all the `resume`, `suspend`, and `cancel` Request APIs all attempt to transition back through the protector and will be stopped if they can't transition.

I wanted to point this out 1) for my own understanding and 2) in case this was not your intent for this logic. Anyways, I think this is solid from what I can tell.

jshier: You're correct: there's a race on `request.state`. Luckily, as you noticed, the result doesn't matter, as both code paths represent correct behavior. Obviously, changing the state before `updateStatesForTask(_:request:)` is called leads directly to the outcome we want. In your scenario, the change to `request.state` is immediate, but `rootQueue` is already executing inside `updateStatesForTask`'s `switch`. So the created `task` will have the action called, and `request` will receive the appropriate task callback. However, when the `Request` action was run, after `state` is changed, the `delegate` was called and its actions were enqueued on `rootQueue`. Once the queue runs that closure we just have to guarantee that the correct action are performed.

In summary, since we know the correct actions are always enqueued after `updateStatesForTasks`, we always behave correctly.

I've added a cancellation test for an automatically resumed `Request` being cancelled from many queues concurrently in https://github.com/Alamofire/Alamofire/pull/2796/commits/f354b4bfacbe1ffeaad99d5f0aa42277a2e7ec4c. We can add more, but this should give us coverage of the worst case scenario.

cnoon: Yep, good callout on the lazy task creation. That will certainly require some internal refactoring of this type of logic, but should still be possible without breaking public API changes.

Issue #2797, opened 2019-04-12
Version string '5.0.0-beta.4' in info.plist is causing TestFlight upload failure


animeshd:
> ℹ Please fill out this template when filing an issue.
> All lines beginning with an ℹ symbol instruct you with what info we expect. You can delete those lines once you've filled in the info.
>
> Per our [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md), we use GitHub for
> bugs and feature requests, not general support. Other issues should be opened on Stack Overflow with the tag `alamofire`.
>
> Please remove this line and everything above it before submitting.

* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

TestFlight upload of my app which imports Alamofire via carthage
`github "Alamofire/Alamofire" "5.0.0-beta.4"`

## What did you expect to happen?

Upload success and app show up on TestFlight as a new version.

## What happened instead?

Got error
`[11:33:55]: [iTMSTransporter] [2019-04-11 11:33:54 PDT] <main> ERROR: ERROR ITMS-90060: "This bundle is invalid. The value for key CFBundleShortVersionString '5.0.0-beta.4' in the Info.plist file must be a period-separated list of at most three non-negative integers. Please find more information about CFBundleShortVersionString at https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring"`

## Alamofire Environment

**Alamofire version:** 5.0.0-beta.4
**Xcode version:** 10.2
**Swift version:** 5.0
**Platform(s) running Alamofire:**
**macOS version running Xcode:** 10.14.4

## Demo Project

ℹ Please link to or upload a project we can download that reproduces the issue.

animeshd:
The problematic value is picked up from [here] https://github.com/Alamofire/Alamofire/blob/e9c742d8791aa3c028c65abdf9966a0cb4bf3e26/Source/Info.plist#L18

TestFlight upload was working fine for release named `5.0.0.beta.1`

cnoon:
#2799 should resolve this issue. I'm going to close this issue out for now. Please direct all further comments to #2799.

PR #2799, opened 2019-04-12
Bugfix - Info.plist TestFlight Validation

cnoon:
This PR replaces the `-` with `.` in the CFBundleShortVersionString value in the Info.plist to pass TestFlight validation.

### Issue Link :link:
#2797 

### Goals :soccer:
To ensure Alamofire can pass TestFlight validation when pulled in as a submodule.

This is much more complex than it seems. We typically use the following naming conventions for Git tags, Podspec versions, the Info.plist versions.

- MAJOR.MINOR.PATCH-{alpha|beta|rc}.{prerelease-version}

This is the most common SemVer friendly format to use for [pre-release versions](https://semver.org/#spec-item-9). It is required by Carthage and SPM. However, as pointed out in #2797, using a `-` in the CFBundleShortVersionString will result in a validation error when uploading to AppStore Connect. After some investigation, it appears that CocoaPods handles this situation by stripping the pre-release version altogether and just using `MAJOR.MINOR.PATCH`.

Because of all of this, we've decided to update the `CFBundleShortVersionString` to use a `.` instead of a `-`. This will fix the ASC validation issue, and also allow `5.0.0.beta.5` to show up in the user agent string instead of just `5.0.0` with submodules, Carthage, and SPM. Unfortunately, this will still show up as `5.0.0` when using CocoaPods, but we figured it was better to have inconsistency here to allow the pre-release version to show up in most places.

### Testing Details :mag:
N/A

### DIFF
--- a/Source/Info.plist
+++ b/Source/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>5.0.0-beta.4</string>
+	<string>5.0.0.beta.4</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>


sumac13:
I don't believe this actually solves the issue when using Carthage

Still getting this error when uploading

`ERROR ITMS-90060: "This bundle is invalid. The value for key CFBundleShortVersionString '5.0.0.beta.5' in the Info.plist file must be a period-separated list of at most three non-negative integers. Please find more information about CFBundleShortVersionString at https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring"
`
`
The call to the iTMSTransporter completed with a non-zero exit status: 1. This indicates a failure.
`

cnoon:
Thank you for pointing this out @sumac13. Looks like we were mistaken with Apple allowing pre-release versions at all. I'll have a PR put together shortly that removes the pre-release info from the version in the Info.plist files.

cnoon:
Addressed in #2808.

tirrorex:
Bumping on issue #2797 @cnoon someone using beta 5 still has the issue (see the second comment in this issue) : 
https://github.com/Alamofire/Alamofire/issues/2853

PR #2802, opened 2019-04-12
Audit - HTTP Headers Refactor

cnoon:
This PR refactors the `HTTPHeader` and `HTTPHeaders` types to `Header` and `Headers`.

### Goals :soccer:
The goal of this PR is to drop the unnecessary `HTTP` prefix as well as simplify the property extensions on system types. Properties with `http` in the name are the Apple variants, where ones without are the Alamofire variants.

Pretty straightforward, just refactored throughout the codebase.

### Testing Details :mag:
N/A


jshier:
I've updated the PR for only the `httpHeaders` to `headers` property rename. The type name remains the same.

PR #2804, opened 2019-04-12
Audit - Request ACL

cnoon:
Update the ACL on all `Request` classes to be `public` instead of `open`.

### Goals :soccer:
Remove the ability to subclass the `Request` types.

AF5 supports extensibility and customization through composition instead of inheritance. Because of this, we've decided to limit subclassing to Alamofire types only (DataRequest, DownloadRequest, UploadRequest). We are limiting the ability to subclass these types since we do not want to expose the complexities of the internal request system through public APIs. Without those, you wouldn't be able to do much anyways by subclassing. The addition of the event monitoring system in AF5 as well removes the need to subclass purely for proxying API calls.

### Testing Details :mag:
N/A

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -26,7 +26,7 @@ import Foundation
 
 /// `Request` is the common superclass of all Alamofire request types and provides common state, delegate, and callback
 /// handling.
-open class Request {
+public class Request {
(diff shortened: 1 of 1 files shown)


cnoon:
Chatted offline in Slack that `final` doesn't make sense since you can't `unfinal` with `@testable`, and it is really useful to be able to subclass for mocking certain behaviors in test suites.

PR #2807, opened 2019-04-14
Bugfix: DownloadRequest event duplication and missing events.

jshier:
### Issue Link :link:
This PR covers a variety of areas, but should also fix the `DownloadRequest` part of #2759.

### Goals :soccer:
This PR fixes a few issues:
- `DownloadRequest` overrides `cancel()` but didn't get the previous refactor. This PR applies the previous strategy to the overridden `cancel()` to fix duplicate event calls.
- `DownloadRequest` was missing event calls for response serialization. These have been added.
- Bizarrely, `DownloadRequest` gets metrics and task completion in a different order than `DataRequest`/`UploadRequest`, which meant that metrics were never actually being gathered, as the task / request map was broken as soon as the task completed but before metrics were gathered. Additional state and API has been added to `RequestTaskMap` to track both events and only break the association when both have been received.
- Duplicate task events could be received when tasks had already transitioned to a new state but the action's work was still enqueued. Logic has been added to ensure a task can be transitioned to a state before doing so. This should ensure that all task transitions happen only once.

### Testing Details :mag:
This PR adds tests for all request types to ensure all lifetime events are properly being called, no more, no less.

### DIFF
--- a/Source/RequestTaskMap.swift
+++ b/Source/RequestTaskMap.swift
@@ -115,7 +115,7 @@ struct RequestTaskMap {
             fatalError("RequestTaskMap consistency error: no events corresponding to task found.")
         }
 
-        switch events {
+        switch (events.completed, events.metricsGathered) {
         case (_, true): fatalError("RequestTaskMap consistency error: duplicate metricsGatheredForTask call.")
         case (false, false): taskEvents[task] = (completed: false, metricsGathered: true)
         case (true, false): self[task] = nil
@@ -127,7 +127,7 @@ struct RequestTaskMap {
             fatalError("RequestTaskMap consistency error: no events corresponding to task found.")
         }
 
-        switch events {
+        switch (events.completed, events.metricsGathered) {
         case (true, _): fatalError("RequestTaskMap consistency error: duplicate completionReceivedForTask call.")
         case (false, false): taskEvents[task] = (completed: true, metricsGathered: false)
         case (false, true): self[task] = nil
--- a/Source/RequestTaskMap.swift
+++ b/Source/RequestTaskMap.swift
@@ -28,14 +28,18 @@ import Foundation
 struct RequestTaskMap {
     private var tasksToRequests: [URLSessionTask: Request]
     private var requestsToTasks: [Request: URLSessionTask]
+    private var taskEvents: [URLSessionTask: (completed: Bool, metricsGathered: Bool)]
 
     var requests: [Request] {
         return Array(tasksToRequests.values)
     }
 
-    init(tasksToRequests: [URLSessionTask: Request] = [:], requestsToTasks: [Request: URLSessionTask] = [:]) {
+    init(tasksToRequests: [URLSessionTask: Request] = [:],
+         requestsToTasks: [Request: URLSessionTask] = [:],
+         taskEvents: [URLSessionTask: (completed: Bool, metricsGathered: Bool)] = [:]) {
         self.tasksToRequests = tasksToRequests
         self.requestsToTasks = requestsToTasks
+        self.taskEvents = taskEvents
     }
 
     subscript(_ request: Request) -> URLSessionTask? {
@@ -48,12 +52,14 @@ struct RequestTaskMap {
 
                 requestsToTasks.removeValue(forKey: request)
                 tasksToRequests.removeValue(forKey: task)
+                taskEvents.removeValue(forKey: task)
 
                 return
             }
 
             requestsToTasks[request] = newValue
             tasksToRequests[newValue] = request
+            taskEvents[newValue] = (completed: false, metricsGathered: false)
         }
     }
 
@@ -67,12 +73,14 @@ struct RequestTaskMap {
 
                 tasksToRequests.removeValue(forKey: task)
                 requestsToTasks.removeValue(forKey: request)
+                taskEvents.removeValue(forKey: task)
 
                 return
             }
 
             tasksToRequests[task] = newValue
             requestsToTasks[newValue] = task
+            taskEvents[task] = (completed: false, metricsGathered: false)
         }
     }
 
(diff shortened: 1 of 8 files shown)


Source/RequestTaskMap.swift
```diff
@@ -28,14 +28,18 @@ import Foundation
 struct RequestTaskMap {
     private var tasksToRequests: [URLSessionTask: Request]
     private var requestsToTasks: [Request: URLSessionTask]
+    private var taskEvents: [URLSessionTask: (completed: Bool, metricsGathered: Bool)]
```
cnoon: I think I'd recommend making a `TaskEvent` struct with `completed` and `metricsGathered` properties to make the switches easier to follow. Right now you're switching on `true` and `false` values without enough context.

Source/Session.swift
```diff
@@ -556,7 +556,8 @@ extension Session: RequestDelegate {
         rootQueue.async {
             request.didCancel()
 
-            guard let task = self.requestTaskMap[request] else {
+            // Cancellation only has an effect on running or suspended tasks.
+            guard let task = self.requestTaskMap[request], [.running, .suspended].contains(task.state) else {
```
cnoon: I feel like we're trying to solve the problem with bandaids rather than addressing the core issue. The reason all these edge cases have been created is that `resume`, `suspend`, and `cancel` are not locking onto the same queue as the internal updates from the `Session`. The state transition is locked, but the execution of the matching operation isn't locked, and that produces the race. Why don't we just eliminate the race altogether?

```swift
@discardableResult
public func resume() -> Self {
    underlyingQueue.async {
        guard self.protectedMutableState.attemptToTransitionTo(.resumed) else { return }
        self.delegate?.resumeRequest(self)
    }

    return self
}
```

By switching `resume` to be called on the `underlyingQueue` (same as `rootQueue` in `Session`), you are now locking the state transition with the execution of the operation. This should solve the actual race and will allow you to remove these checks in the delegate as well as remove the async dispatch here at the beginning.

Source/RequestTaskMap.swift
jshier: Ah, that would make the calculations easier.

Source/Session.swift
jshier: Making the entire process asynchronous opens us up to host of other other issues. Namely:
- User expectations are no longer met. When a `URLSessionTask` has `cancel()` called, the `state` of that task is updated immediately, despite the fact that the delegate method indicating the actual cancellation will come later. It seems reasonable for users to expect the same behavior from Alamofire, especially since we don't offer any sort of async completion handler for these actions.
- In addition to user expectations not being met, internal expectations may no longer hold, as we could no longer rely on actions like cancellation being queryable immediately, meaning the library may do a lot of unnecessary work, like issuing a request retry, when it didn't need to. So the race is still there, we've just moved its effects somewhere else internally.

Ultimately, this is no longer a thread-safety issue but an order of operations issue. Even though we're using a serial queue internally, there's no way for us to guarantee the order of operations that have dependent effects when enqueueing things asynchronously. i.e. There's no way for us to insert a cancellation action before an already pending retry attempt. So since that problem isn't solvable (at least not without a huge amount of work) we aren't buying ourselves anything by putting the state transition on the same queue.

So while such a change might make dealing with task creation a bit easier, it would complicate the cancellation side of things quite a bit, and the state updates would no longer be immediately available to users.

Source/RequestTaskMap.swift
jshier: Actually, looking into it, it doesn't really make the calculations any easier. What exactly did you have in mind? AFAICT, I'll need to switch on the booleans either way.

Source/Session.swift
cnoon: Seems like `underlyingQueue.sync` would solve the issue.

Source/RequestTaskMap.swift
cnoon: It would just make the switch more clear:

```swift
switch events {
case (true, _): fatalError("RequestTaskMap consistency error: duplicate completionReceivedForTask call.")
case (false, false): taskEvents[task] = (completed: true, metricsGathered: false)
case (false, true): self[task] = nil
}
```

vs.

```swift
switch (events.completed, events.metricsGathered) {
case (true, _): fatalError("RequestTaskMap consistency error: duplicate completionReceivedForTask call.")
case (false, false): taskEvents[task] = (completed: true, metricsGathered: false)
case (false, true): self[task] = nil
}
```

Right now, I have no idea what's on the left and what's on the right when looking at the switch.

Source/Session.swift
jshier: Not really. `sync` just waits for the closure to completed. That means we'd still have cancellation enqueued after internal actions which query for that state. In addition, we'd block the calling queue, which may not be a trivial amount of time considering the `rootQueue` is also the `URLSession`'s `delegateQueue`, as well as the `underlyingQueue` for every `Request` created by that `Session` instance. It would also mean we couldn't call any of this API internally ourselves, as we'd deadlock.

I'll give a solution here more thought.

Source/RequestTaskMap.swift
jshier: Okay, I can do that.

Source/Session.swift
cnoon: It opens up a bigger issue of the queuing system design. This should be a change that we could make, but I'm not sure we can without hanging up this call in certain situations (i.e. root queue is parsing a big JSON blob). By using the same underlying queue across the board, I can't see how we could prevent this from happening. This is why we've used separate queues in the past.

Source/RequestTaskMap.swift
```diff
@@ -83,10 +91,46 @@ struct RequestTaskMap {
...
         return tasksToRequests.isEmpty
     }
+
+    var isEventsEmpty: Bool {
+        precondition(taskEvents.isEmpty == isEmpty, "RequestTaskMap.isEventsEmpty invalid, isEmpty: \(isEmpty) != taskEvents.isEmpty: \(taskEvents.isEmpty)")
+
+        return taskEvents.isEmpty
+    }
+
+    mutating func disassociateIfNecessaryAfterGatheringMetricsForTask(_ task: URLSessionTask) {
```
cnoon: I wonder if reusing the state provider APIs would make sense here:

```swift
disassociateIfNecessaryAfterGatheringMetricsForTask -> didGatherMetricsForTask
disassociateIfNecessaryAfterCompletingTask -> didCompleteTask
```

jshier: Those were actually my first names for these methods but I wanted to indicate there was a side effect of possibly breaking the association between task and request.

cnoon: Gotcha. I think I'd roll with these then.

Source/Session.swift
cnoon: Based on offline Slack conversation, let's move forward with this as is right now.

PR #2809, opened 2019-04-17
[WIP] Feature - Request State Queue

cnoon:
This PR is a work-in-progress to demonstrate a possible way to address the request state synchronization challenges between `Session` and `Request`.

### Goals :soccer:
To ensure `Request` state updates through the `resume`, `suspend` and `cancel` APIs cannot produce inconsistent behavior where state changes cause an operation to execute that is no longer valid.

The core change here is to pivot away from relying on the async behavior of the `rootQueue` to manage the thread-safety of the `Request` state changes.

### Testing Details :mag:
Test suite is passing without any thread-sanitizer issues after switching `RequestTaskMap` over to a `class` instead of `struct` which makes sense anyways since it lives alongside the `Session`.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -102,7 +102,7 @@ public class Request {
     }
 
     /// Protected `MutableState` value that provides threadsafe access to state values.
-    let protectedMutableState: Protector<MutableState> = Protector(MutableState()) // NOTE: CN - probably shouldn't make this internal
+    fileprivate let protectedMutableState: Protector<MutableState> = Protector(MutableState())
 
     /// `State` of the `Request`.
     public fileprivate(set) var state: State {
@@ -467,6 +467,21 @@ public class Request {
     /// Called when creating a `URLSessionTask` for this `Request`. Subclasses must override.
     func task(for request: URLRequest, using session: URLSession) -> URLSessionTask {
         fatalError("Subclasses must override.")
+    }
+
+    // MARK: Request State
+
+    func attemptToTransitionTo(_ state: Request.State) -> Bool {
+        var transitioned = false
+
+        protectedMutableState.write { mutableState in
+            guard mutableState.state.canTransitionTo(state) else { return }
+
+            mutableState.state = state
+            transitioned = true
+        }
+
+        return transitioned
     }
 
     // MARK: - Public API
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -102,7 +102,7 @@ public class Request {
     }
 
     /// Protected `MutableState` value that provides threadsafe access to state values.
-    fileprivate let protectedMutableState: Protector<MutableState> = Protector(MutableState())
+    let protectedMutableState: Protector<MutableState> = Protector(MutableState()) // NOTE: CN - probably shouldn't make this internal
 
     /// `State` of the `Request`.
     public fileprivate(set) var state: State {
@@ -480,10 +480,7 @@ public class Request {
     /// - Returns: The `Request`.
     @discardableResult
     public func cancel() -> Self {
-        guard protectedMutableState.attemptToTransitionTo(.cancelled) else { return self }
-
-        delegate?.cancelRequest(self)
-
+        delegate?.attemptToTransition(self, to: .cancelled)
         return self
     }
 
@@ -492,10 +489,7 @@ public class Request {
     /// - Returns: The `Request`.
     @discardableResult
     public func suspend() -> Self {
-        guard protectedMutableState.attemptToTransitionTo(.suspended) else { return self }
-
-        delegate?.suspendRequest(self)
-
+        delegate?.attemptToTransition(self, to: .suspended)
         return self
     }
 
@@ -505,10 +499,7 @@ public class Request {
     /// - Returns: The `Request`.
     @discardableResult
     public func resume() -> Self {
-        guard protectedMutableState.attemptToTransitionTo(.resumed) else { return self }
-
-        delegate?.resumeRequest(self)
-
+        delegate?.attemptToTransition(self, to: .resumed)
         return self
     }
 
@@ -725,10 +716,9 @@ public protocol RequestDelegate: AnyObject {
     func retryResult(for request: Request, dueTo error: Error, completion: @escaping (RetryResult) -> Void)
     func retryRequest(_ request: Request, withDelay timeDelay: TimeInterval?)
 
-    func cancelRequest(_ request: Request)
+    func attemptToTransition(_ request: Request, to state: Request.State)
+
     func cancelDownloadRequest(_ request: DownloadRequest, byProducingResumeData: @escaping (Data?) -> Void)
-    func suspendRequest(_ request: Request)
-    func resumeRequest(_ request: Request)
 }
 
 // MARK: - Subclasses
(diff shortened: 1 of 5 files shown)


Source/Session.swift
```diff
@@ -552,23 +585,48 @@ extension Session: RequestDelegate {
...
-            guard let task = self.requestTaskMap[request], [.running, .suspended].contains(task.state) else {
-                request.finish()
-                return
-            }
+            switch state {
+            case .resumed:
+                request.didResume()
+
+                // Tasks can only be resumed if they're suspended.
+                guard let task = self.requestTaskMap[request], task.state == .suspended else { return }
```
cnoon: Shouldn't need the state checks anymore.

Source/Session.swift
```diff
@@ -469,21 +489,33 @@ open class Session {
...
-            task.resume()
-            request.didResumeTask(task)
-        case (_, .suspended):
-            task.suspend()
-            request.didSuspendTask(task)
-        case (_, .cancelled):
-            task.cancel()
-            request.didCancelTask(task)
+        requestStateQueue.sync {
+            switch (startRequestsImmediately, request.state) {
```
cnoon: While not super obvious, this is the core issue that is resolved in this PR. Previously, after the `request.state` is read, it could be modified when running the `switch`. This produces all sorts of tricky edge cases that need to be handled. Fundamentally, we need to be able to rely on that state not being swapped out from under us to ensure the operation we're running is still the correct one to run. This ensures that's the case.

Source/Session.swift
```diff
@@ -552,23 +585,48 @@ extension Session: RequestDelegate {
         }
     }
 
-    public func cancelRequest(_ request: Request) {
-        rootQueue.async {
-            request.didCancel()
+    public func attemptToTransition(_ request: Request, to state: Request.State) {
+        requestStateQueue.sync {
+            guard request.protectedMutableState.attemptToTransitionTo(state) else { return }
```
cnoon: All of this logic can be redesigned to be MUCH cleaner by combining state transition logic with the `requestStateQueue` itself in some specialized APIs. This is just a first attempt to show how it could actually work, then we can clean it up if we like it.

Source/Session.swift
cnoon: Removed in d65b5e56 since they are no longer necessary.

Source/Session.swift
cnoon: Refactored in 613f6314 into a `finishRequestIfCancelled(_:)` API. It cleans up the logic a TON, but I'm not in love with the name. It was the best I could come up with.

cnoon:
This design currently keeps the `requestStateQueue` completely internal to `Session`. I think I like this approach the best, but we could also consider passing the `requestStateQueue` to each request as well. That would allow us to run the `resume` logic inside the queue directly in `Request`, instead of having to jump out to `Session` through the delegate, hop onto the queue, then attempt to transition the state back on `Request`. Since all this logic is internal, it's not a huge deal either way.

Another thing this could potentially allow though would be for us to also be able to put the `Request.state` property itself behind the queue. Therefore, you wouldn't even be able to grab the state until any other state operations complete which is exactly what you want. These state modifications are super quick and just lump the actual state change with the calls on the task keeping everything perfectly in sync. However, if we did want to lock the `Request.state` access inside the `requestStateQueue` as well, we'd need to probably switch over to using a recursive lock instead of `sync` on a DispatchQueue. Otherwise we'll be in a case where we'll have to call into state already inside a `requestStateQueue.sync` which would deadlock.

Anyways, lots of options here.

Source/Session.swift
```diff
@@ -469,21 +482,33 @@ open class Session {
...
-            request.didCancelTask(task)
+        requestStateQueue.sync {
+            switch (startRequestsImmediately, request.state) {
+            case (true, .initialized):
+                guard request.attemptToTransitionTo(.resumed) else { return }
+
+                request.didResume()
+
+                task.resume()
+                request.didResumeTask(task)
```
cnoon: Should we hop back to `rootQueue` for these `didResume` and `didResumeTask` calls as well? The only things we actually need to call on `requestStateQueue` are the transition and `task.resume()`.

Source/RequestTaskMap.swift
```diff
@@ -128,9 +131,12 @@ struct RequestTaskMap {
...
-        case (true, _): fatalError("RequestTaskMap consistency error: duplicate completionReceivedForTask call.")
-        case (false, false): taskEvents[task] = (completed: true, metricsGathered: false)
-        case (false, true): self[task] = nil
+        case (true, _):
+            fatalError("RequestTaskMap consistency error: duplicate completionReceivedForTask call.")
+        case (false, false):
+            taskEvents[task] = (completed: true, metricsGathered: false)
+        case (false, true):
+            self[task] = nil
         }
```
cnoon: Just seemed easier to read on newlines.

Source/Request.swift
```diff
@@ -469,6 +469,21 @@ public class Request {
...
 
+    // MARK: Request State
+
+    func attemptToTransitionTo(_ state: Request.State) -> Bool {
+        var transitioned = false
+
+        protectedMutableState.write { mutableState in
+            guard mutableState.state.canTransitionTo(state) else { return }
+
+            mutableState.state = state
```
cnoon: We should remove the ability to modify `state` this way in a future PR. Instead, `canTransitionTo` should be a `transitionTo` API that only allows transitioning to a valid state.

Source/Session.swift
```diff
@@ -446,10 +448,21 @@ open class Session {
         }
     }
 
+    private func finishRequestIfCancelled(_ request: Request) -> Bool {
```
cnoon: I could really use some help with the name of this API.

Source/Session.swift
```diff
@@ -552,69 +577,60 @@ extension Session: RequestDelegate {
...
-            // Cancellation only has an effect on running or suspended tasks.
-            guard let task = self.requestTaskMap[request], [.running, .suspended].contains(task.state) else {
-                request.finish()
-                return
-            }
+            switch state {
+            case .resumed:
+                request.didResume()
+
+                guard let task = request.task else { return }
```
cnoon: By using the `request.task` instead of the `requestTaskMap`, we avoid having to lock the `RequestTaskMap` since it could be accessed from different threads.

Tests/SessionTests.swift
```diff
@@ -449,36 +449,38 @@ class SessionTestCase: BaseTestCase {
...
+//
+//        // Then
+//        XCTAssertNotNil(response, "response should not be nil")
+//        XCTAssertTrue(request.isCancelled)
+//        XCTAssertTrue((request.task == nil) || (request.task?.state == .canceling || request.task?.state == .completed))
+//
+//        guard let error = request.error?.asAFError, case .explicitlyCancelled = error else {
+//            XCTFail("Request should have an .explicitlyCancelled error.")
+//            return
+//        }
```
cnoon: This test is fixed in #2810.

cnoon:
Based on a very in-depth Slack conversation, @jshier and I have both decided to go with #2814 instead of the approach used in this PR. Both accomplish the same goal which is to lock the request state change with the task update call. The main difference is that #2814 accomplishes this without having to add an additional locking mechanism. It leverages the unfair lock already inside the protector.

PR #2810, opened 2019-04-17
Feature - Terminal Response Serializers

cnoon:
This PR fixes an issue that should be an extreme edge case of misuse where developers are adding a response serializer to an already completed request. Previously, the completion would never be called. It's now called with a specific response serialization error.

### Goals :soccer:
Any response serializer added to a completed request should still have the completion called with a failure.

Thankfully, the implementation here is pretty straightforward. We are now tracking a `.finished` state on `Request`. When a response serializer is appended to a `Request`, we now check if the `Request` is `.finished`. If it is, we update the error on the `Request`, and kick off another round of response serialization processing. This will clear out the response serializer and the process can continue as many times as the client desires.

Overall, no one should ever hit this use case. We only found this due to a test creating a request, cancelling it, resuming it, then adding a response serializer to it.

```swift
let request = session.request(urlRequest)
    .cancel()
    .resume()
    .response { resp in
        response = resp
        expectation.fulfill()
    }
```

This is not a case that we intend to support because it is explicitly misusing the public API. However, we wanted to ensure that the response serializer completion was still called with a clear error that it was added after the `Request` has already completed.

#### Previous Behavior and Retry Logic

This use case is supported in AF 4.x with the way the internal Request queue handles response serializer additions. In AF5, we've rebuilt the threading system, and also the response serialization system to support retry as well. We cannot support both retry from response serializers as well as executing addition response serializers after the request completes due to really weird edge cases that would crop up with retry logic.

For example, let's say I am using retry, and append two response serializers correctly. I execute the request, it goes through a retry, and the two completions are called. Then a user adds a 3rd response serializer to the `Request`. If we just executed it like normal, it "could" trigger a retry, but how would we do that? We've already called the completions on the first two response serializers. It wouldn't make sense to execute them again. It also wouldn't make sense to retry a `Request` based on the third response serializer's retry result without the other two response serializers knowing about it. 

Because of these weird conditions, we've decided to have this simply result in an error. There are no conditions where you should be resuming your requests before adding response serializers through chaining when `startRequestsImmediately` is `false`. When it's `true`, You need to add your response serializers immediately after to ensure they are picked up before a cached response is pulled from the `URLCache`.

### Testing Details :mag:
Added a failing test that demonstrated the issue and the fix is now passing.

### DIFF
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -139,6 +139,8 @@ public enum AFError: Error {
         case decodingFailed(error: Error)
         /// Generic serialization failed for an empty response that wasn't type `Empty` but instead the associated type.
         case invalidEmptyResponse(type: String)
+        /// A response serializer was added to the request after the request was already finished.
+        case responseSerializerAddedAfterRequestFinished
     }
 
     /// Underlying reason a server trust evaluation error occured.
@@ -554,6 +556,8 @@ extension AFError.ResponseSerializationFailureReason {
             return "Empty response could not be serialized to type: \(type). Use Empty as the expected type for such responses."
         case .decodingFailed(let error):
             return "Response could not be decoded because of error:\n\(error.localizedDescription)"
+        case .responseSerializerAddedAfterRequestFinished:
+            return "Response serializer was added to the request after it had already finished."
         }
     }
 }
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -38,17 +38,24 @@ public class Request {
     /// - cancelled: Set when `cancel()` is called. Any tasks created for the `Request` will have `cancel()` called on
     ///              them. Unlike `resumed` or `suspended`, once in the `cancelled` state, the `Request` can no longer
     ///              transition to any other state.
+    /// - finished:  Set when all response serialization completion closures have been cleared on the `Request` and
+    ///              queued on their respective queues.
     public enum State {
-        case initialized, resumed, suspended, cancelled
+        case initialized, resumed, suspended, cancelled, finished
 
         /// Determines whether `self` can be transitioned to `state`.
         func canTransitionTo(_ state: State) -> Bool {
             switch (self, state) {
-            case (.initialized, _): return true
-            case (_, .initialized), (.cancelled, _): return false
-            case (.resumed, .cancelled), (.suspended, .cancelled),
-                 (.resumed, .suspended), (.suspended, .resumed): return true
-            case (.suspended, .suspended), (.resumed, .resumed): return false
+            case (.initialized, _):
+                return true
+            case (_, .initialized), (.cancelled, _), (.finished, _):
+                return false
+            case (.resumed, .cancelled), (.suspended, .cancelled), (.resumed, .suspended), (.suspended, .resumed):
+                return true
+            case (.suspended, .suspended), (.resumed, .resumed):
+                return false
+            case (_, .finished):
+                return true
             }
         }
     }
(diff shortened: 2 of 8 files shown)


Source/AFError.swift
```diff
@@ -139,6 +139,8 @@ public enum AFError: Error {
         case decodingFailed(error: Error)
         /// Generic serialization failed for an empty response that wasn't type `Empty` but instead the associated type.
         case invalidEmptyResponse(type: String)
+        /// The response serializer was added to the request after the request was already completed.
```
jshier: I think this should say "A response serializer...", and "completed" should probably be "finished."

Source/AFError.swift
```diff
@@ -139,6 +139,8 @@ public enum AFError: Error {
...
+        case responseSerializerAddedAfterRequestFinished
```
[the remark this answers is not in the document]
jshier: Sounds good to me.

Source/Request.swift
```diff
@@ -38,17 +38,26 @@ public class Request {
...
 
         /// Determines whether `self` can be transitioned to `state`.
         func canTransitionTo(_ state: State) -> Bool {
             switch (self, state) {
-            case (.initialized, _): return true
-            case (_, .initialized), (.cancelled, _): return false
-            case (.resumed, .cancelled), (.suspended, .cancelled),
-                 (.resumed, .suspended), (.suspended, .resumed): return true
-            case (.suspended, .suspended), (.resumed, .resumed): return false
+            case (.initialized, _):
```
jshier: I think the state machine need to be updated to ensure (or make explicit) we can only transition to `finished` from non-`cancelled` states.

Source/Request.swift
```diff
@@ -387,7 +398,14 @@ public class Request {
 
     /// Appends the response serialization closure to the `Request`.
     func appendResponseSerializer(_ closure: @escaping () -> Void) {
-        protectedMutableState.write { $0.responseSerializers.append(closure) }
+        protectedMutableState.write { mutableState in
+            mutableState.responseSerializers.append(closure)
+
+            if mutableState.state == .finished {
```
jshier: Probably want to check for `.cancelled` separately so we run the serializers either way.

Source/Request.swift
```diff
@@ -420,6 +438,8 @@ public class Request {
                 // An example of how this can happen is by calling cancel inside a response completion closure.
                 mutableState.responseSerializers.removeAll()
                 mutableState.responseSerializerCompletions.removeAll()
+
+                mutableState.state = .finished
```
jshier: We only want to transition to `.finished` if we weren't cancelled, so this should probably use the state transition API.

Source/Request.swift
cnoon: I agree. I actually put that code in and it's covered by this case already:

```swift
case (_, .initialized), (.cancelled, _):
    return false
```

cnoon: To make it more clear, I updated that case to this:

```swift
case (_, .initialized), (.cancelled, _), (.finished, _):
    return false
```

Source/Request.swift
cnoon: Yep, for sure. I had to add a new `responseSerializerProcessingFinished` flag in mutable state to track the completion. We can't just trigger response serializer processing when the request state is cancelled, because the regular request process will eventually get around to calling process once the queues catch up with the state change.

Source/Request.swift
cnoon: Agreed.

Source/AFError.swift
cnoon: Resolved in 656f331d.

Source/Request.swift
cnoon: Resolved in 656f331d.

Source/Request.swift
cnoon: Check added in 656f331d.

PR #2818, opened 2019-05-03
Fix an issue where empty Encodable Parameters resulted in a non-nil query string.

DavidBarry:
### Goals :soccer:
This fixes an issue where a `Parameters` instance passed to `URLEncodedFormParameterEncoder` that produced an empty query string would result in that empty string being set to the `percentEncodedQuery` instead of `nil`. This would result in a URL of `https://httpbin.org/get?` instead of `https://httpbin.org/get` and differs from the behavior of the existing `URLEncoding`.

This adds a check for an empty string when setting the `percentEncodedQuery` and sets `nil` instead in that case.

### Testing Details :mag:
A new test was added to cover this case using an empty dictionary as the parameter type. This test mirrors one that exists for `URLEncoding` for this situation.


jshier:
Yep, looks great! Thanks @DavidBarry!

PR #2836, opened 2019-06-09
Update handling of os_unfair_lock to manage the buffer.

jshier:
### Goals :soccer:
In talking to Apple engineers at WWDC, I was told that passing a single lock instance by reference may lead to a crash rate of up to 1%, due to the compiler managing the lifetime of the instance in unexpected ways. This PR switches to keeping the memory buffer directly. We'll need to do the same thing if we add support for `pthread_mutex` back as well.

Rather than passing a single `os_unfair_lock` by reference to lock and unlock, manage the `os_unfair_lock_t` instance directly by manually allocating and initializing the memory, as well as cleaning up afterward.

### Testing Details :mag:
No tests were updated. I haven't seen this, so this is a preventative fix.

### DIFF
--- a/Source/Protector.swift
+++ b/Source/Protector.swift
@@ -28,14 +28,24 @@ import Foundation
 
 /// An `os_unfair_lock` wrapper.
 final class UnfairLock {
-    private var unfairLock = os_unfair_lock()
+    private let unfairLock: os_unfair_lock_t
+    
+    init() {
+        unfairLock = .allocate(capacity: 1)
+        unfairLock.initialize(to: os_unfair_lock())
+    }
+    
+    deinit {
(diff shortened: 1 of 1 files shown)


jshier:
@cnoon In the midst of building a Combine operator for me (which is now on the `initial-combine-support` branch, he needed a locking mechanism. Once I told him about our `os_unfair_lock` wrapper, he immediately wanted to check it for this bug. Given that he's also the `Operation` maintainer, locking is kind of his thing. So it was a fortuitous but unrelated discovery.

cnoon:
Wow, that's awesome! Super glad you happened to be chatting with that particular person! That would have been horrible for us to have to try and track down.

Issue #2841, opened 2019-06-14
Cancel download without producing resume data in tmp folder


ullstrm:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

We are using DownloadRequest and are calling its `cancel`-method when the user explicitly cancels the download. This method is calling the `cancel(byProducingResumeData` on `URLSessionDownloadTask` which will move the partially downloaded file to the tmp-folder.

## What did you expect to happen?

I would like to have a way of cancelling this download request without producing the resume-data.

## What happened instead?

I need to import Alamofire with @testable and call the `URLSessionDownloadTask`'s `cancel` explicitly for this to not happend.

## Alamofire Environment

**Alamofire version: 4.8.2
**Xcode version: 10.2.1
**Swift version: 5
**Platform(s) running Alamofire: iOS
**macOS version running Xcode: 10.14.5

jshier:
You're right, there's no way to cancel a `DownloadRequest` without producing resume data. This was done under the assumption that anyone cancelling a download would want to resume it in the future, if possible. Unfortunately it's not a behavior we can change for Alamofire 4, as other users may be depending on it. For testing, I would suggest clearing your temp directory between test runs if you want to cleanup this file.

For Alamofire 5, we could make this behavioral change, both to match `URLSessionDownloadTask`'s behavior, and because the current cancellation behavior is a bit awkward to use without a callback when resume data is available. Users have to access it in response handlers, which is non-obvious. I have a callback version in my `inline-docs-update` branch, but both versions still produce resume data. @cnoon Any thoughts on this? I can easily make `cancel()` not produce resume data and have a `cancel(byProducingResumeData:)` with an optional closure, but that changes a fundamental behavior Alamofire has had for 5 years.

cnoon:
Seems like something we should probably support IMO. I think your suggestion @jshier is probably the best approach. Let users call the `Request.cancel()` API if that's what they actually want, then add a resume data variant on DownloadRequest that has the exactly same logic currently, it just calls the completion when finished generating the resume data.

ullstrm:
What are the odds that a function i.e. `cancel(createResumeData: Bool = true)` will find its way into `Alamofire 4.8.2`? Would this be a candidate for a PR? By adding that parameter, no unexpected behaviour will appear for users as long as the parameter defaults to `true`

[...]

PR #2851, opened 2019-06-20
Add option to not create resume data on cancel for DownloadRequest

ullstrm:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/2841

### Goals :soccer:
Add option to not create resume data on cancel for DownloadRequest. I've noticed that in Alamofire 5 this has now become the default behaviour of the cancel for DownloadRequest. It would be great to at least add this possibility to Alamofire 4.

An assumption is made that there is no problem to just leave the resumeData-property of the downloadDelegate as-is.

### Testing Details :mag:
I've added a test that verifies that this function is not adding any resume data.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -497,8 +497,19 @@ open class DownloadRequest: Request {
     // MARK: State
 
     /// Cancels the request.
-    open override func cancel() {
-        downloadDelegate.downloadTask.cancel { self.downloadDelegate.resumeData = $0 }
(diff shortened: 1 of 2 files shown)


BinaryDennis:
This PR looks good to me 👍
Would be nice if it got merged soon

danielsaidi:
Yes, we need this in the main repo!

jshier:
Thanks for the PR!

PR #2856, opened 2019-06-26
progressQueue and responseQueue

RomanPodymov:
### Goals :soccer:
Created constants for the default queues using in functions with progress handler and in responses. For instance it will useful [here](https://github.com/tristanhimmelman/AlamofireObjectMapper/blob/6.1.0/AlamofireObjectMapper/AlamofireObjectMapper.swift), we will be able to declare `public func responseObject<T: BaseMappable>(queue: DispatchQueue = .main, keyPath: String? = nil, mapToObject object: T? = nil, context: MapContext? = nil, completionHandler: @escaping (DataResponse<T>) -> Void) -> Self` as `public func responseObject<T: BaseMappable>(queue: DispatchQueue = Alamofire.responseQueue, keyPath: String? = nil, mapToObject object: T? = nil, context: MapContext? = nil, completionHandler: @escaping (DataResponse<T>) -> Void) -> Self`.

Added `public let progressQueue = DispatchQueue.main` and `public let responseQueue = DispatchQueue.main`, use them in the corresponding functions instead of `DispatchQueue.main`.

### Testing Details :mag:
No new tests added as this is a very light refactoring

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -605,11 +605,11 @@ public class Request {
     /// Only the last closure provided is used.
     ///
     /// - Parameters:
-    ///   - queue:   The `DispatchQueue` to execute the closure on. Defaults to `.main`.
+    ///   - queue:   The `DispatchQueue` to execute the closure on. Defaults to `progressQueue`.
     ///   - closure: The code to be executed periodically as data is read from the server.
     /// - Returns:   The `Request`.
     @discardableResult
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! However, the use of `.main` is already a reusable value. If you wish to track your own queues like this, go ahead, but this doesn't fit with the library. Thanks anyway!

Issue #2844, opened 2019-06-15
Migrate to native Swift Result enum


philtre:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

A native `Result` enum has been added to the Swift Standard Library in Xcode 10.2. 
Also, because `Error` can now conform to itself, we can use it as a concrete type in generics (eg. `Result<String, Error>`).

The native Swift `Result` differs from the Alamofire implementation in two significant ways:
1. It defines a generic constraint on the failure type
2. the `flatMap` and `flatMapError` methods do not throw but rather return a new `Result`

I propose refactoring Alamofire's implementation of `Result` so that it is implemented as a set of extensions on the native Swift implementation.

I'll be submitting a pull request shortly.

jshier:
This has already been done as part of our work for Alamofire 5. We've removed all public `Result` extensions and left the few we need internally visible to the library.

philtre:
I guess my question would be whether there is a need for the `AFResult` typealias when the code using the two extensions - `flatMap` and `flatMapError` - can be easily changed to use the native non-throwing methods.

Moreover, the `DataResponse` and `DownloadResponse` classes now have `flatMap` and `flatMapError` methods with a different signature than the native `Result` equivalents.

Would it be sensible to add a failure type constraint on those two classes (eg. `DataResponse<Success, Failure: Error>`) and then simply forward the `map` and `flatMap` closures to the object's `result` property?

jshier:
`AFResult` is used heavily in our tests, and is also what we expose in our public APIs that use `Result`.

As for the mapping APIs, which the APIs don't match `Result`'s, I don't think they need to. `Result` was intentionally given a restricted API set by the Swift core team, given the considerations of life in the standard library, and I don't think we need to limit all `Result` using APIs to that case.

As to making the `Response` types doubly generic, that may have some value. Would you be willing to be together a PR making a change and exploring the impact on users? My main concern is that it adds complexity where most users don't need it, so it would be good to see some motivating examples.

@cnoon Any thoughts here?

cnoon:
I think I could go either way with the doubly generic response types here @jshier. I get the motivation, but I agree with your point that this is most likely additional complexity for the benefit of very few. It's also a pretty big breaking change that might be a bit painful for clients to adopt.

[...]

PR #2864, opened 2019-07-09
Remove AFResult and add a failure constraint to Response

philtre:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/2844

### Goals :soccer:
Since the introduction of a native `Result` enum in Swift and protocol self-conformance, there is very little need for a custom `AFResult` enum. 
Indeed, recent Alamofire commits have removed the custom `AFResult` implementation in favor of a typealias which hides the failure constraint.
However, the native `Result` provides `flatMap` and `flatMapError` implementations that differ from those in Alamofire. 
The Alamofire implementation accepts throwing closures: 

```swift
func flatMap<NewSuccess>(_ transform: (Success) throwing -> NewSuccess) -> Result<NewSuccess, Error> { ... }
```

... while the native implementation accepts closures returning a new `Result`:

```swift
func flatMap<NewSuccess>(_ transform: (Success) -> Result<NewSuccess, Failure>) -> Result<NewSuccess, Failure> { ... }
```

The latter is IMO much more versatile because `Result` already has a throwing initializer. One can easily achieve Alamofire's implementation:
```swift
let dataResult: Result<Data, Error> = ...
let myObjectResult: Result<Int, Error> = dataResult.flatMap { data in 
    .init{ try MyObject(from: data) } 
}
```
... as well as other cool things like defaulting to a value in case of error:
```swift
let defaultObjectResult = myObjectResult.flatMapError{ error in 
    if error is NotFoundError {
       .success(MyObject.default) 
    }
    return .failure(MyError(error))
}
```

The above also applies to `DataResponse` and `DownloadResponse`. These two classes also have `flatMap` and `flatMapError` functions that differ from their native counterparts.
In order to bring them closer to what is now standard behavior, one would have to add a `Failure` constraint to these generics.

Let's say your API layer returns both a `Result` or a `DataResponse`:

```swift

enum APIError: Error {...}

class API {
    func fetchTextResponse(completion: (DataResponse<String>) -> Void) {
       AF.request(...).responseString{ dataResponse in 
          completion(dataResponse)
        }
    }
    func fetchTextResult(completion: (Result<String, APIError>) -> Void) {
        fetchTextResponse { dataResponse in 
            let result: Result<String, APIError> = dataResponse.result.flatMapError { error in
                if dataResponse.response?.statusCode == 404 {
                    return .success("Not Found")
                }
                return APIError.serverError(error)
            }
            completion(result)
        }
    }
}

func handleAPIError(_ error: APIError) {...}

func onResult(_ result: Result<String, APIError>) {
    switch result {
        case .success(let value):
            print(value)
        case .failure(let error):
            handleAPIError(error)
    }
}

API().fetchTextResult{ result in 
   onResult(result)
}
API().fetchTextResponse{ dataResponse in 
    let apiResult: Result<String, APIError> = dataResponse.result.flatMapError { error in
            if dataResponse.response?.statusCode == 404 {
                return .success("Not Found")
            }
            return APIError.serverError(error)
    }
    onResult(apiResult)
}
```

This is obviously not ideal and because `DataResponse` and `DownloadResponse` don't have a generic failure constraint, the only way to avoid flat-mapping the error outside of the API class is to create a custom doubly-generic wrapper for `DataResponse` and `DownloadResponse`.

The above example would look like this with doubly generic responses:

```swift

enum APIError: Error {...}

class API {
    func fetchTextResponse(completion: (DataResponse<String, APIError>) -> Void) {
       AF.request(...).responseString{ dataResponse in 
            let apiResponse = dataResponse.flatMapError { error in
                if dataResponse.response?.statusCode == 404 {
                    return .success("Not Found")
                }
                return APIError.serverError(error)
            }
            completion(apiResponse)
        }
    }
    func fetchTextResult(completion: (Result<String, APIError>) -> Void) {
        fetchTextResponse { apiResponse in 
            completion(apiResponse.result)
        }
    }
}

func handleAPIError(_ error: APIError) {...}

func onResult(_ result: Result<String, APIError>) {
    switch result {
        case .success(let value):
            print(value)
        case .failure(let error):
            handleAPIError(error)
    }
}

API().fetchTextResult{ result in 
    onResult(result)
}
API().fetchTextResponse{ response in 
    onResult(response.result)
}
```

Since `DataResponse` and `DownloadResponse` are very commonly used in code, this change could potentially be a big annoyance to existing users.
To avoid this, I kept these as typealieases to two new structs:

```swift
public struct FailingDataResponse<Success, Failure: Error> {...}
public struct FailingDownloadResponse<Success, Failure: Error> {...}
public typealias DataResponse<Success> = FailingDataResponse<Success, Error>
public typealias DownloadResponse<Success> = FailingDownloadResponse<Success, Error>
```

### Testing Details :mag:
I modified some of the tests to use the new `flatMap` and `flatMapError` functions.

I also changed the `import Alamofire` statements to `@testable import Alamofire` so that internal properties are available for testing. 
I considered adding `Result+Alamofire.swift` to the test targets instead, but found it to be too confusing.


jshier:
Thanks for the PR! There's a lot to process here, so rather than reviewing this all at once I think it would be best to break it into several PRs to review individually. Here's what I was thinking:

**Removing `AFResult`**
- We cannot fully remove `AFResult`. It is a migration tool between Alamofire 4's `Result<T>` type and Swift's `Result<T, E>` type in Swift 5.
- We need to keep the `typealias` and its integration into our public APIs so that users can transition just by using a find / replace.
- Removal of all other internal and testing uses of the type should be fine.
- There should be no changes the internal API we expose, since it's just for our use.
- There should be no additional `@testable` imports. We avoid these as much as possible to ensure we're properly testing the public API of Alamofire and not internal components.

**`*Response` Functional API Changes**
- There is some room for change here, but matching the `Result` API is not appropriate, especially the `flatMap` change. I plan to add a `tryMap` to `Result` since the core team isn't a fan of a throwing `flatMap` or `rethrows` on `Result`, but there's no reason for `*Response` to be bound by that.
- Ultimately, I don't think `flatMap` makes any sense on `*Response`, as we're not flattening any `*Response`-like structure into the value.
- What we should do is make the `map` function `rethrows` so it can handle both throwing and non-throwing transforms, like we see on `Array`.
- `mapError` can have the same change made.
- The `flatMap` variants can be removed.

**Doubly Generic `*Response`**
- This decision will come last, as I'm still interested in seeing how it affects user code both old and new.
- The renaming isn't going to work, as the types should be named according to Swift's best practices. This is one reason I'd like to see this change separately, so we can see how impactful it is without any attempt at mitigation.
- If we do make the types doubly generic, we should also consider wrapper all errors in `AFError`, so that the final types is always `*Response<T, AFError>`, which can then be easily mapped to other error types. This is a bigger change which can come afterward, but we probably don't want to ship the doubly generic versions unless we also do this, so we'll see.

@cnoon Any feedback on any of these points or the PR?

So, I think separating the changes into PRs like this will allow us to consider them separately, even if they're ultimately dependent on each other. @philtre, if you feel like you can make these changes and work with us on the process, let us know so we can turn these PR reviews around quickly, as we'd like to do a release (beta 7 or rc 1) soon.

philtre:
@jshier I'm more than happy to split this PR into smaller ones. 

I do want to discuss the functional changes to `*Response` a bit more, though. 

The way I look at a `*Response` structure is that it is a "Wrapper around a HTTP response with post-processing of the raw response body into a `Result` property". 
So, the mapping functions are just additional post-processing of the `result` property (none of them modify the underlying HTTP response in any way). 
Structurally, it would make a lot of sense for the `*Response` API to match the `Result` API because it's literally just forwarding those mapping closures to the `result` property and re-wrapping the resulting `Result` value.
This would obviously mean that the `*Response` structures would have to be doubly generic, just like the wrapped `Result`.

I do like the idea of a doubly generic `*Response` having `AFError` as the default error type.

---

As far as making mapping functions `rethrow`, I just can't think of a use case for it. If I want to transform the result of an async operation, I don't want it to throw. If anything, I want an error thrown by the transformation closure to be wrapped into a `.failure` and returned without throwing.

jshier:
> So, the mapping functions are just additional post-processing of the result property (none of them modify the underlying HTTP response in any way).
Structurally, it would make a lot of sense for the *Response API to match the Result API because it's literally just forwarding those mapping closures to the result property and re-wrapping the resulting Result value.

That the `value` and `error` transforms are done by transforming the `result` value is an implementation detail and has no impact on the naming considerations of higher level `*Response` APIs. As far as users are concerned, those appear to be separate APIs. Additionally, the `*Response` doesn't have the same limitations on API surface that `Result` does, so containing ourselves to API built with vastly different concerns is not appropriate.

I agree that making the type doubly generic will likely unlock the full power of these transforms, but I believe they can be separated so that the API changes are made first, and then the double change, as that will have a much larger impact to the library, especially if we want to wrap all errors in `AFError`.

> As far as making mapping functions rethrow, I just can't think of a use case for it. If I want to transform the result of an async operation, I don't want it to throw. If anything, I want an error thrown by the transformation closure to be wrapped into a .failure and returned without throwing.

You're right, we don't want `rethrows`, just to combine the current `map` and `flatMap` implementations into one that takes a throwing closure. Current Swift best practice seems to say that `flatMap` should only exist to actually flatten similar types together, not "flatten" errors into other values.

jshier:
Thinking about it more, it seems like combining `map` and `flatMap` would be a bad idea given we lose the ability to have a specific error type in that case. At most perhaps renaming `flatMap` to `tryMap` might be the only change to make.

@philtre Are you still interested in breaking up this PR along the lines described?

philtre:
@jshier Yes, I have three PRs ready:

1. Remove all references to AFResult https://github.com/Alamofire/Alamofire/pull/2891
2. `flatMap` to `tryMap` https://github.com/Alamofire/Alamofire/pull/2892
3. Doubly generic `*Response` https://github.com/Alamofire/Alamofire/pull/2893

I also attempted to wrap all errors in `AFError`, which was one of the discussed use cases for doubly generic `*Response`. It actually wouldn't be a big problem. There's only a few places where thrown errors would have to be wrapped into `AFError`.

philtre:
Pretty sure we can close this now.

PR #2887, opened 2019-08-02
throw error while JSONEncoding without POST method

adyliu:
### Issue Link :link:

- https://github.com/Moya/Moya/issues/1533
- https://github.com/Moya/Moya/issues/1645
- https://stackoverflow.com/questions/52626403/alamofire-request-changes-method-name-on-its-own
- https://stackoverflow.com/questions/55172430/the-requested-resource-does-not-support-http-method-t-or-st-with-alamofire
- https://stackoverflow.com/questions/48654888/it-shows-405-server-error-at-first-time-to-get-result-from-api-server-and-it-wo?rq=1

The current behaviour, where `NSURLSession` sends the Content-Length header but not the message body, is hard to justify.

For example:

```
import Alamofire

let completionHandler: (DataResponse<String>) -> Void = { (resp) in
    switch resp.result {
    case .success(let txt):
        print("\(txt)")
    case .failure(let err):
        print("ERROR => \(err)")
    }
}

Alamofire.Session.default.request("http://httpbin.org/get",
    method: .get,
    parameters: [:],
    encoding: JSONEncoding.default).responseString(completionHandler: completionHandler)
```

The server will hang up and wait the client to send the request body, but `NSURLSession` will never send them.

```
	GET /get HTTP/1.1
	Host: httpbin.org
	Content-Type: application/json
	Connection: keep-alive
	Accept: */*
	User-Agent: demo3/1.0 (com.apple.dt.playground.stub.iOS_Simulator.demo3-0AAFA08A-42ED-4E2F-AF5A-AA0E3670BD87; build:1; iOS 12.4.0) Alamofire/5.0.0
	Accept-Language: en;q=1.0
	Content-Length: 2
	Accept-Encoding: br;q=1.0, gzip;q=0.9, deflate;q=0.8
```

The java servlet containers will read the next two bytes if exists, so the next http request will left
```
    T /some-url
    ST /some-url
```

The `T` means `GET`, and `ST` means 'POST`.

The developer should change parameters to `parameters: nil` or change encoding to `encoding: URLEncoding.default`.

So we should tell the developer : 
it's not suitable that using JSONEncoding in the GET request.

### DIFF
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -172,6 +172,8 @@ public enum AFError: Error {
     case parameterEncoderFailed(reason: ParameterEncoderFailureReason)
     /// Multipart form encoding failed.
     case multipartEncodingFailed(reason: MultipartEncodingFailureReason)
+    /// `POST` method required with `httpBody`
+    case postMethodRequiredWithHttpBody
     /// `RequestAdapter` failed threw an error during adaptation.
     case requestAdaptationFailed(error: Error)
     /// Response validation failed.
@@ -444,6 +446,8 @@ extension AFError: LocalizedError {
             return reason.localizedDescription
         case .multipartEncodingFailed(let reason):
             return reason.localizedDescription
+        case .postMethodRequiredWithHttpBody:
+            return "`POST` method required with Http Body"
         case .requestAdaptationFailed(let error):
             return "Request adaption failed with error: \(error.localizedDescription)"
         case .responseValidationFailed(let reason):
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -266,7 +266,7 @@ public struct JSONEncoding: ParameterEncoding {
         guard let parameters = parameters else { return urlRequest }
         // swift NSURLSession sends the 'Content-Length' header but not the message body for other methods; `httpBody` will hang up the remote session or make the packet confused in the server
         guard "POST" == ( urlRequest.httpMethod?.uppercased() ?? "") else {
-            throw AFError.ParameterEncoderFailureReason.missingRequiredComponent(.httpMethod(rawValue: "POST"))
+            throw AFError.parameterEncoderFailed(reason: .missingRequiredComponent(.httpMethod(rawValue: "POST")))
         }
 
         do {
@@ -299,7 +299,7 @@ public struct JSONEncoding: ParameterEncoding {
         guard let jsonObject = jsonObject else { return urlRequest }
         // swift NSURLSession sends the 'Content-Length' header but not the message body for other methods; `httpBody` will hang up the remote session or make the packet confused in the server
         guard "POST" == ( urlRequest.httpMethod?.uppercased() ?? "") else {
-            throw AFError.ParameterEncoderFailureReason.missingRequiredComponent(.httpMethod(rawValue: "POST"))
+            throw AFError.parameterEncoderFailed(reason: .missingRequiredComponent(.httpMethod(rawValue: "POST")))
         }
         
         do {
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -264,10 +264,6 @@ public struct JSONEncoding: ParameterEncoding {
         var urlRequest = try urlRequest.asURLRequest()
 
         guard let parameters = parameters else { return urlRequest }
-        // swift NSURLSession sends the 'Content-Length' header but not the message body for other methods; `httpBody` will hang up the remote session or make the packet confused in the server
-        guard "POST" == ( urlRequest.httpMethod?.uppercased() ?? "") else {
-            throw AFError.parameterEncoderFailed(reason: .missingRequiredComponent(.httpMethod(rawValue: "POST")))
-        }
 
         do {
             let data = try JSONSerialization.data(withJSONObject: parameters, options: options)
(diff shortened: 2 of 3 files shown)


jshier:
Thanks for the PR! This is something that we’ve thought about before but declined to implement due to the possibility of users needing this flexibility. However, now that Apple has added a specific error for this case (GET requests with bodies), it might be a good idea for Alamofire to do the same thing. 

So rather than limiting this to `JSONParameterEncoder` I think it would be a good idea to put a validation check right in the lifetime methods that create `URLRequests` in general. This logic is part of `Session’, and we could put the validation logic in a `URLRequest` extension. That way an error would be generated no matter the encoder used.

This would require a new `AFError` case and some tests, in addition to the other logic changes. Would you be able to make these changes?

@cnoon, any additional thoughts here?

cnoon:
So there's an error now thrown by `URLSession` if you set the `httpBody` on a `GET` request @jshier? While I completely agree that in theory that would make sense, I think that's most likely too strict due to odd backend API design of legacy systems. There's nothing that prohibits backend devs from requiring `httpBody` data for `GET` requests. While I completely understand that we shouldn't be doing that, it doesn't mean we won't have to do that in certain situations.

Aside from the `httpBody` error comment, I think that AF should not be setting a `Content-Length` like that if there's no content. That certainly seems like a bug. If that's because you're setting an encoding type, but no parameters, then it seems like our API isn't strict enough, or should be smarter internally.

jshier:
> So there's an error now thrown by URLSession if you set the httpBody on a GET request @jshier? 

This is a new error returned by `URLSession` for all 2019 Apple OSes (see the [release notes](https://developer.apple.com/documentation/macos_release_notes/macos_catalina_10_15_beta_5_release_notes):

> All URLSessionTask instances with a GET HTTP method which contain a body will now produce the error NSURLErrorDataLengthExceedsMaximum. (46025234)

This will likely come to Foundation open source soon as well. Given that fact, Alamofire preventing the creation and resumption of a task when we detect that case seems like a good idea to me and makes behavior consistent across all OS versions Alamofire support.

> Aside from the httpBody error comment, I think that AF should not be setting a Content-Length like that if there's no content. That certainly seems like a bug. If that's because you're setting an encoding type, but no parameters, then it seems like our API isn't strict enough, or should be smarter internally.

Alamofire never explicitly sets the `Content-Length` header, it's always been handled by `URLSession`. Unfortunately `URLSession` has never properly set it for `GET` requests with body data, so this isn't a behavior change for us at all. In fact, if we wanted Alamofire to support `GET` requests with body data properly, we should be setting the header ourselves. However, Apple's changes to the 2019 OSes force us down a single path here, so having a single error seems like our best option.

jshier:
@adyliu Would you be able to update your PR along the lines I described above?

adyliu:
> @adyliu Would you be able to update your PR along the lines I described above?

I updated the code.

Our developers used the error `JSONEncoding` in some base `classes` and buried the **quirky** bug which costed me lots of time to find it.

> Apple will fix the bug at OS 10.15. Maybe we should tell the developer to error before that.

These apis should not throw errors because so many users use them in the code.
```
Alamofire.AF.request()
```
and
```
Session.request()
```


```
    open func request<Parameters: Encodable>(_ convertible: URLConvertible,
                                             method: HTTPMethod = .get,
                                             encoder: ParameterEncoder = URLEncodedFormParameterEncoder.default,
                                             headers: HTTPHeaders? = nil,
                                             interceptor: RequestInterceptor? = nil) -> DataRequest {
        let convertible = RequestEncodableConvertible(url: convertible,
                                                      method: method,
                                                      parameters: parameters,
                                                      encoder: encoder,
                                                      headers: headers)

        return request(convertible, interceptor: interceptor)
    }
```

Throwing the error in the `Encoder` is more comfortable.

```
    struct RequestEncodableConvertible<Parameters: Encodable>: URLRequestConvertible {
        let url: URLConvertible
        let method: HTTPMethod
        let parameters: Parameters?
        let encoder: ParameterEncoder
        let headers: HTTPHeaders?

        func asURLRequest() throws -> URLRequest {
            let request = try URLRequest(url: url, method: method, headers: headers)

            return try parameters.map { try encoder.encode($0, into: request) } ?? request
        }
    }
```

jshier:
@adyliu No, I meant moving the error from the encoder to a validation step for the `URLRequest`s produced when making requests. If you look at `performSetupOperations(for request: Request, convertible: URLRequestConvertible)`, you should be able to add validation steps after the initial request and after the adapted request are created and the errors should be handled automatically.

adyliu:
@jshier I move the codes to `Session.performSetupOperations` as your advice.

jshier:
@adyliu Sorry, I guess I should've been more specific. What we need here is to match the validation that `URLSession` now performs. Which means that:
1. We can wrap the validation rule in a method on `URLRequest` (we already have an extension file for it) and have it throw a new `AFError` case with its own reason type:
```swift
extension URLRequest {
    func validate() throws {
        if method == .get && httpBody != nil {
            throw AFError.invalidURLRequest(reason: ...) 
        }
    }
}
```
2. This validation should be called when we create `URLRequest`s, so in `performSetupOperations(for request: Request, convertible: URLRequestConvertible)` after we create the initial `URLRequest` and then after we adapt it.
```swift
    func performSetupOperations(for request: Request, convertible: URLRequestConvertible) {
        do {
            let initialRequest = try convertible.asURLRequest()
            try initialRequest.validate()
            rootQueue.async { request.didCreateInitialURLRequest(initialRequest) }

            guard !request.isCancelled else { return }

            if let adapter = adapter(for: request) {
                adapter.adapt(initialRequest, for: self) { result in
                    do {
                        let adaptedRequest = try result.get()
                        try adaptedRequest.validate()
```

I believe this should be enough to get the validation we need while match `URLSession`'s behavior.

jshier:
@adyliu I've put up #2905 to add the validation I specified. Thanks for the PR and for prompting this!

PR #2890, opened 2019-08-06
Add cancelAllRequests API

jshier:
### Issue Link :link:
While there's no specific request, this is a question that comes up every so often, so it makes sense to add an API for it.

### Goals :soccer:
This PR adds a `cancelAllRequests` method to `Session`, which cancels all currently in flight `Request`s. It's a more limited version of a similar API on `URLSession`, in that it does't block additional `Request`s from being made while cancellation is occurring.

This PR adds `var activeRequests: Set<Request>` to `Session`, and `Request`s are added when created and removed when they run `cleanup()` after running the response serializers.

This PR also adds a workaround for the apparent fact that `URLSessionTask`s which are never resumed will never return `URLSessionTaskMetrics`. Given this is an important event for Alamofire to see, we now always call `resume()` on tasks before calling `cancel()`. This appears to guarantee metrics will always be generated for the task. However, on the 2019 OSes, as of beta 5, this is not 100% reliable in Alamofire, resulting in about 1 failure per 25,000 cancellations, at least according to my tests. Previous OS versions are 100% reliable, so I've reported this as a bug (FB6789976).

### Testing Details :mag:
I've added tests for cancelling 100 auto resumed and non resumed `Request`s, as well as a retry scenario. Let me know if any other scenarios are needed.

### DIFF
--- a/Source/Session.swift
+++ b/Source/Session.swift
@@ -202,11 +202,13 @@ open class Session {
     ///         `Request`s may not cancel immediately due internal work, and may not cancel at all if they are close to
     ///         completion when cancelled.
     ///
-    /// - Parameter completion: Closure to be called when all `Request`s have been cancelled.
-    public func cancelAllRequests(completion: (() -> Void)? = nil) {
+    /// - Parameters:
+    ///   - queue:      `DispatchQueue` on which the completion handler is run. `.main` by default.
+    ///   - completion: Closure to be called when all `Request`s have been cancelled.
+    public func cancelAllRequests(completingOnQueue queue: DispatchQueue = .main, completion: (() -> Void)? = nil) {
         rootQueue.async {
             self.activeRequests.forEach { $0.cancel() }
-            completion?()
+            queue.async { completion?() }
         }
     }
 
@@ -776,9 +778,9 @@ open class Session {
     /// - Parameter request: The `Request` to perform.
     func perform(_ request: Request) {
         switch request {
-        case let r as DataRequest: self.perform(r)
-        case let r as UploadRequest: self.perform(r)
-        case let r as DownloadRequest: self.perform(r)
+        case let r as DataRequest: perform(r)
+        case let r as UploadRequest: perform(r)
+        case let r as DownloadRequest: perform(r)
         default: fatalError("Attempted to perform unsupported Request subclass: \(type(of: request))")
         }
     }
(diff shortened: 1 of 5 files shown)


Source/Session.swift
```diff
@@ -192,6 +194,22 @@ open class Session {
...
+    ///
+    /// - Note: This is an asynchronous operation and does not block the creation of future `Request`s. Cancelled
+    ///         `Request`s may not cancel immediately due internal work, and may not cancel at all if they are close to
+    ///         completion when cancelled.
+    ///
+    /// - Parameter completion: Closure to be called when all `Request`s have been cancelled.
+    public func cancelAllRequests(completion: (() -> Void)? = nil) {
+        rootQueue.async {
+            self.activeRequests.forEach { $0.cancel() }
+            completion?()
```
cnoon: Should we hop back off `rootQueue` here to avoid clients hanging it up? If yes, should we provide a `queue` parameter that defaults maybe to `.main`?

Source/Session.swift
```diff
@@ -750,11 +768,17 @@ open class Session {
...
+    ///
+    /// - Parameter request: The `Request` to perform.
     func perform(_ request: Request) {
         switch request {
-        case let r as DataRequest: perform(r)
-        case let r as UploadRequest: perform(r)
-        case let r as DownloadRequest: perform(r)
+        case let r as DataRequest: self.perform(r)
+        case let r as UploadRequest: self.perform(r)
+        case let r as DownloadRequest: self.perform(r)
```
cnoon: Do we actually need `self.` now?

Source/Session.swift
```diff
@@ -64,6 +64,8 @@ open class Session {
 
     /// Internal map between `Request`s and any `URLSessionTasks` that may be in flight for them.
     var requestTaskMap = RequestTaskMap()
+    /// Set of currently active `Request`s.
+    var activeRequests: Set<Request> = []
```
cnoon: I like it!

Source/Session.swift
jshier: I was trying to match `URLSession` behavior here, but this is a better idea. Added in https://github.com/Alamofire/Alamofire/pull/2890/commits/39aaa58af41d5d423c60d21613c64a1f5b75d0bd

Source/Session.swift
jshier: Nope, removed in https://github.com/Alamofire/Alamofire/pull/2890/commits/39aaa58af41d5d423c60d21613c64a1f5b75d0bd.

jshier:
@cnoon This is ready for another look.

PR #2891, opened 2019-08-07
Replace all internal references to AFResult<T> with Result<T, Error>

philtre:
### Goals :soccer:
Remove all internal references to `AFResult` but keep the `AFResult` typealias for migration purposes.

This is an attempt to break apart https://github.com/Alamofire/Alamofire/pull/2864


Source/AFResult.swift
```diff
@@ -25,83 +25,3 @@
 /// `Result` that always has an `Error` `Failure` type.
 public typealias AFResult<T> = Result<T, Error>
```
jshier: There's really no reason to have just this `typealias` in its own file. It can be combined with the `Result` extensions.

Source/ResponseSerialization.swift
```diff
@@ -189,7 +189,7 @@ extension DataRequest {
         appendResponseSerializer {
             // Start work that should be on the serialization queue.
             let start = CFAbsoluteTimeGetCurrent()
-            let result = AFResult { try responseSerializer.serialize(request: self.request,
+            let result = Result<Serializer.SerializedObject, Error> { try responseSerializer.serialize(request: self.request,
```
jshier: Please adjust the parameter alignment, and perhaps put the `try` on its own line to it isn't as wide.

Source/AFResult.swift
philtre: I've moved the typealias to `Result+Alamofire.swift`

Source/ResponseSerialization.swift
philtre: Done.

PR #2892, opened 2019-08-07
Rename internal `flatMap` to `tryMap`

philtre:
### Goals :soccer:

1. Rename internal `Result` mapping functions: 
  - `flatMap` to `tryMap`,
  - `flatMapError` to `tryMapError`

2. Rename internal `Result` convenience properties:
  - `success` to `value`
  - `failure` to `error`

3. Use autoclosure properties in convenience `Result` initializer

This PR builds on top of https://github.com/Alamofire/Alamofire/pull/2891
and is an attempt to break up https://github.com/Alamofire/Alamofire/pull/2864

### DIFF
--- a/Source/Result+Alamofire.swift
+++ b/Source/Result+Alamofire.swift
@@ -57,10 +57,10 @@ extension Result {
 
     /// Evaluates the specified closure when the `Result` is a success, passing the unwrapped value as a parameter.
     ///
-    /// Use the `flatMap` method with a closure that may throw an error. For example:
+    /// Use the `tryMap` method with a closure that may throw an error. For example:
     ///
     ///     let possibleData: Result<Data, Error> = .success(Data(...))
-    ///     let possibleObject = possibleData.flatMap {
+    ///     let possibleObject = possibleData.tryMap {
     ///         try JSONSerialization.jsonObject(with: $0)
     ///     }
     ///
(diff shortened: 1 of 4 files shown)


Source/Result+Alamofire.swift
```diff
@@ -0,0 +1,106 @@
...
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+// MARK: - Internal APIs
+
+extension Result {
+    /// Returns the associated value if the result is a success, `nil` otherwise.
+    var value: Success? {
```
jshier: These property names should not be changed. In fact, they were changed *to* `success` / `failure` to get away from our previous names in order to prevent [a confusing error](https://stackoverflow.com/questions/55484714/value-is-inaccessible-due-to-internal-protection-level/55485139#55485139) seen when using `value` / `error` with older Alamofire tutorials.

Source/Result+Alamofire.swift
```diff
@@ -0,0 +1,106 @@
...
+        guard case .failure(let error) = self else { return nil }
+        return error
+    }
+
+    /// Initializes a `Result` from value or error. Returns `.failure` if the error is non-nil, `.success` otherwise.
+    ///
+    /// - Parameters:
+    ///   - value: A value.
+    ///   - error: An `Error`.
+    init(value: @autoclosure () -> Success, error: @autoclosure () -> Failure?) {
```
jshier: What was the purpose of this change? I can't find a use site in the current diff that shows improvement with the use of autoclosures..

Source/Result+Alamofire.swift
philtre: I was unaware of this problem. I will revert this change.

Source/Result+Alamofire.swift
philtre: If `error` is non-nil, the `value` closure isn't executed at all. I can also revert this change if you disagree that there is any benefit in this.

jshier:
@philtre Can you merge `master` or rebase to fix the conflicts and cleanup the diff here?

philtre:
Branch rebased.

PR #2901, opened 2019-08-12
Unfreeze HTTPMethod

jshier:
### Goals :soccer:
This PR converts `HTTPMethod` from an `enum` into a `struct`. This should allow users to keep using `HTTPMethod` when they have to make calls using methods we don't cover, without having to resort to raw `Strings`s.

`HTTPMethod` is now a `struct` conforming to `RawRepresentable`, `Equatable`, and `Hashable`. No internal changes were necessary except for the use of a no longer optional initializer. All `String`s can be made valid `HTTPMethod`s now, standardizing on the `uppercased()` value of the `String` as the `rawValue`.

### Testing Details :mag:
No tests added, as there's no custom behavior here. Can certainly be added if there's anything that might be necessary.

### DIFF
--- a/Source/HTTPMethod.swift
+++ b/Source/HTTPMethod.swift
@@ -22,7 +22,8 @@
 //  THE SOFTWARE.
 //
 
-/// HTTP method definitions.
+/// Type representing HTTP methods. Raw `String` value is stored and compared case-sensitively, so
+/// `HTTPMethod.get != HTTPMethod(rawValue: "get")`.
 ///
 /// See https://tools.ietf.org/html/rfc7231#section-4.3
 public struct HTTPMethod: RawRepresentable, Equatable, Hashable {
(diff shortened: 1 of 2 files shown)


Source/HTTPMethod.swift
```diff
@@ -25,23 +25,32 @@
...
-    case trace   = "TRACE"
+    public static let trace = HTTPMethod(rawValue: "TRACE")
+
+    public let rawValue: String
+    
+    /// Creates a new instance with the `uppercased()` version of the specified `rawValue`.
+    /// 
+    /// - Parameter rawValue: The raw `String` value, stored `uppercased`.
+    public init(rawValue: String) {
+        self.rawValue = rawValue.uppercased()
```
cnoon: Is there any reason @jshier that we'd not want to uppercase the `rawValue`?

jshier: Only if we want to support case-sensitive methods (e.g. `get` vs. `GET` vs. `GeT`), which I don't think we do. I'll take another look at the standard though.

jshier: Actually, it looks like [HTTP methods are considered case-sensitive](https://tools.ietf.org/html/rfc7230#section-3.1.1). I'm not sure how much that's used in practice though. To be save it does seem like I should remove the `uppercased()` and make a not about the sensitivity.

jshier: Updated in https://github.com/Alamofire/Alamofire/pull/2901/commits/83732f1549244cdebbab05998ac86583ad5001d1.

Issue #2683, opened 2019-01-09
Strip XSSI common header from JSON


grosch:
This is an enhancement request.  A common XSSI protection is to prepend this string:

>)]}',\n

to the front of the HTTP body when the content-type is application/json.  Angular's HttpClient library for example knows to automatically strip that out from the message body.  Is it possible for Alamofire to do that as well?

jshier:
This certainly isn't something we'd want to do automatically or by default. I suppose it's possible we could offer a configuration option in the JSON response serializers to do it, but it's so non-standard that I question the value of including it in the library. It's relatively trivial to implement yourself.

jshier:
This should be resolved by #2903. Specifically, you can pass `{ ($0.prefix(6) == Data(")]}',\n".utf8)) ? $0.dropFirst(6) : $0 }` as the `dataPreprocessor` parameter to strip the XSSI string.

jshier:
This has been changed to use a `DataPreprocessor` protocol, so now you can use the `GoogleXSSIPreprocessor` to strip this data.

PR #2903, opened 2019-08-12
Add DataPreprocessor API

jshier:
### Issue Link :link:
This is a generalized solution to #2683.

### Goals :soccer:
This PR adds the `DataPreprocessor` API to all response serializers, to allow easy access to raw `Data` before serialization, to do things like stripping XSSI strings.

`DataPreprocessor` is just a closure that each serializer can have set. By default it just passes `Data` through, so it should have minimal impact for anyone not using the API. It also supports throwing, so errors can be returned as well.

### Testing Details :mag:
Additional tests were added to the various `DataResponseSerializer` cases, as all serialization flows through those methods.

### DIFF
--- a/Source/ResponseSerialization.swift
+++ b/Source/ResponseSerialization.swift
@@ -64,8 +64,6 @@ public protocol DownloadResponseSerializerProtocol {
 
 /// A serializer that can handle both data and download responses.
 public protocol ResponseSerializer: DataResponseSerializerProtocol & DownloadResponseSerializerProtocol {
-    /// Closure type used to prepare `Data` for serialization.
-    typealias DataPreprocessor = (Data) throws -> Data
     /// `DataPreprocessor` closure used to prepare incoming `Data` for serialization.
     var dataPreprocessor: DataPreprocessor { get }
     /// `HTTPMethod`s for which empty response bodies are considered appropriate.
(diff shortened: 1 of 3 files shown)


jshier:
Design question: is the closure good enough, or should this be full abstracted into a protocol?

Source/ResponseSerialization.swift
```diff
@@ -64,18 +64,25 @@ public protocol DownloadResponseSerializerProtocol {
 
 /// A serializer that can handle both data and download responses.
 public protocol ResponseSerializer: DataResponseSerializerProtocol & DownloadResponseSerializerProtocol {
+    /// Closure type used to prepare `Data` for serialization.
+    typealias DataPreprocessor = (Data) throws -> Data
```
cnoon: I really like this change overall. It really gives you a lot of flexibility if you need it. I think for users that actually need this functionality though might be better served by this being a protocol rather than a closure. I think most people would want to define an explicit type to very easily test rather than having to implement it as a closure. I just think it could be some fairly complex logic that people may want to reuse more easily. They could pass functions off as the closure, but that doesn't seem ideal.

In the XSSI issue, it seems like it would be nice to be able to create an `XSSIDataPreprocessor: DataPreprocessor` that you could extensively unit test and leverage throughout your code. So I think I lean towards having `DataPreprocessor` be a protocol rather than a typealias'd closure. Either way, both solutions are solid.

Source/ResponseSerialization.swift
```diff
@@ -398,33 +405,37 @@ extension DataRequest {
 /// request returning `nil` or no data is considered an error. However, if the response is has a status code valid for
 /// empty responses (`204`, `205`), then an empty `Data` value is returned.
 public final class DataResponseSerializer: ResponseSerializer {
-    /// HTTP response codes for which empty responses are allowed.
+    public let dataPreprocessor: DataPreprocessor
```
cnoon: Does `dataPreprocessor` have a place in the top-level response chainable APIs, or is it such an advanced feature that it only warrants customization at the response serializer initializers?

Also, would you expect people to create their own chainable extensions when using these?

> Not that this matters so much for this PR, but just curious in general.

jshier:
@cnoon I've refactored `DataPreprocessor` into a protocol, take a look! I think it's okay, but would be far better but for a Swift limitation around being able to use static members of protocols.

Source/ResponseSerialization.swift
```diff
@@ -64,18 +64,42 @@ public protocol DownloadResponseSerializerProtocol {
...
+struct Passthrough: DataPreprocessor {
+    func preprocess(_ data: Data) throws -> Data { return data }
+}
+
+/// `DataPreprocessor` that trims Google's typical `)]}',\n` XSSI JSON header.
+struct XSSI: DataPreprocessor {
+    func preprocess(_ data: Data) throws -> Data {
+        return (data.prefix(6) == Data(")]}',\n".utf8)) ? data.dropFirst(6) : data
+    }
+}
```
cnoon: I'm not sure this `XSSI` type is actually being used at all right now. Also, would it make sense to expose `Passthrough` as a `public` type? Would it be convenient for clients, or would you just expect those building custom response serializers to use `ResponseSerializer.defaultDataPreprocessor`?

cnoon: If either of these do get made `public`, I think I'd recommend tacking on a suffix like `Preprocessor` or `DataPreprocessor` so it's a bit more clear what they are.

Source/ResponseSerialization.swift
jshier: That could be the next step for this API, but I don't think it will be used much, so I don't think it's valuable as a chainable API. Customizing the existing serializers should be enough for what users need here.

Source/ResponseSerialization.swift
jshier: I've made them both public, updated the naming, and added some tests.

Source/ResponseSerialization.swift
jshier: I updated to a protocol in https://github.com/Alamofire/Alamofire/pull/2903/commits/6d1a440ceb7d3bd8ae6aed4cc923d17bf202c0fa.

PR #2915, opened 2019-08-21
Refactor Reachability

jshier:
# Goals :soccer:
This PR refactors our `SCNetworkReachability` wrapper to simplify usage, adopt more modern styling, and fix a bug.

There are a few changes in here:
- Moved `ConnectionType` inside `NetworkReachabilityStatus`, since the types are related.
- Added `default` static instance (using the zero address) so users don't have to create an instance in the common case.
- Renamed usages of `WWAN` to `Cellular`, since that's what it actually means.
- Rewrote `DispatchQueue` handling to separate the queue used for reachability notifications and listener callbacks. Notifications are now always on a background queue and the listener queue is now customizable when listening is started, while defaulting to `.main`.
- Lifetime management has been updated: `startListening` now calls `stopListening` to clear any state before setting up a new `SCNetworkReachabilityContext`.
- The zero address initializer has been simplified to no longer require memory rebounding.
- Various status computations have been moved to initializers or convenience properties, rather than methods on the manager itself.
- Added support for the cellular checks on tvOS because they're technically supported, so why not.

### Testing Details :mag:
Added some lifecycle tests where the manager is started, stopped, and started again, and updated other tests that calculate the reachability status.

### DIFF
--- a/Source/NetworkReachabilityManager.swift
+++ b/Source/NetworkReachabilityManager.swift
@@ -27,7 +27,7 @@
 import Foundation
 import SystemConfiguration
 
-/// The `NetworkReachabilityManager` class listens for reachability changes of hosts and addresses for both WWAN and
+/// The `NetworkReachabilityManager` class listens for reachability changes of hosts and addresses for both cellular and
 /// WiFi network interfaces.
 ///
 /// Reachability can be used to determine background information about why a network operation failed, or to retry
@@ -48,9 +48,7 @@ open class NetworkReachabilityManager {
 
             var networkStatus: NetworkReachabilityStatus = .reachable(.ethernetOrWiFi)
 
-            #if os(iOS)
-                if flags.isCellular { networkStatus = .reachable(.cellular) }
-            #endif
+            if flags.isCellular { networkStatus = .reachable(.cellular) }
 
             self = networkStatus
         }
@@ -68,12 +66,19 @@ open class NetworkReachabilityManager {
     /// network reachability status.
     public typealias Listener = (NetworkReachabilityStatus) -> Void
 
+    /// Default `NetworkReachabilityManager` for the zero address and a `listenerQueue` of `.main`.
+    public static let `default` = NetworkReachabilityManager()
+
     // MARK: - Properties
 
     /// Whether the network is currently reachable.
     open var isReachable: Bool { return isReachableOnCellular || isReachableOnEthernetOrWiFi }
 
     /// Whether the network is currently reachable over the WWAN interface.
+    ///
+    /// - Note: Using this property to decide whether to make a high or low bandwidth request is not recommended.
+    ///         Instead, set the `allowsCellularAccess` on any `URLRequest`s being issued.
+    ///
     open var isReachableOnCellular: Bool { return status == .reachable(.cellular) }
 
     /// Whether the network is currently reachable over Ethernet or WiFi interface.
@@ -94,9 +99,7 @@ open class NetworkReachabilityManager {
 
     /// The current network reachability status.
     open var status: NetworkReachabilityStatus {
-        guard let flags = flags else { return .unknown }
-
-        return NetworkReachabilityStatus(flags)
+        return flags.map(NetworkReachabilityStatus.init) ?? .unknown
     }
 
     /// A closure executed when the network reachability status changes.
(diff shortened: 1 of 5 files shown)


Source/NetworkReachabilityManager.swift
```diff
@@ -42,65 +42,86 @@ open class NetworkReachabilityManager {
...
-    }
-
-    /// The dispatch queue to execute the `listener` closure on.
-    open var listenerQueue: DispatchQueue = DispatchQueue.main
+    open var isReachableOnEthernetOrWiFi: Bool { return status == .reachable(.ethernetOrWiFi) }
 
-    /// A closure executed when the network reachability status changes.
-    open var listener: Listener?
+    /// `DispatchQueue` on which reachability will update.
+    public let reachabilityQueue = DispatchQueue(label: "org.alamofire.reachabilityQueue")
```
cnoon: Does the `label` become an issue if you create more than 1 manager? I'm sure I've done this before without realizing. Would two separate DispatchQueues created with the same label actually be two different instances that just happened to share the same label? Would that only affect the queue visualization in Xcode? Or are there other implications?

Source/NetworkReachabilityManager.swift
```diff
@@ -109,24 +130,17 @@ open class NetworkReachabilityManager {
...
+        var zero = sockaddr()
+        zero.sa_len = UInt8(MemoryLayout<sockaddr>.size)
+        zero.sa_family = sa_family_t(AF_INET)
 
-        guard let reachability = withUnsafePointer(to: &address, { pointer in
-            return pointer.withMemoryRebound(to: sockaddr.self, capacity: MemoryLayout<sockaddr>.size) {
-                return SCNetworkReachabilityCreateWithAddress(nil, $0)
-            }
-        }) else { return nil }
+        guard let reachability = SCNetworkReachabilityCreateWithAddress(nil, &zero) else { return nil }
```
cnoon: Much better 👌🏻

Source/NetworkReachabilityManager.swift
```diff
@@ -137,86 +151,100 @@ open class NetworkReachabilityManager {
...
-        listener?(networkReachabilityStatusForFlags(flags))
-    }
+        let newStatus = NetworkReachabilityStatus(flags)
 
-    // MARK: - Internal - Network Reachability Status
+        guard previousStatus != newStatus else { return }
 
-    func networkReachabilityStatusForFlags(_ flags: SCNetworkReachabilityFlags) -> NetworkReachabilityStatus {
-        guard isNetworkReachable(with: flags) else { return .notReachable }
+        previousStatus = newStatus
```
cnoon: Do we also need to lock status read access inside the reachability queue to be fully thread-safe?

Source/NetworkReachabilityManager.swift
jshier: It's just a debugging aid, so it would only affect visualization in the debugger and Xcode. If we want to ensure all the queues we use are uniquely labeled, I think we should do a separate PR, as there are a few queues in the `Session` that would need to be made unique. I'm not sure how valuable it is, since you can always tell queues apart by their address.

Source/NetworkReachabilityManager.swift
jshier: 👍

Source/NetworkReachabilityManager.swift
jshier: I've made `previousStatus` protected in https://github.com/Alamofire/Alamofire/pull/2915/commits/005376d2c1714e377f9e6064205b91eefaee34a3, which I think is enough to protect our usage. I'm not sure about the thread-safety of `reachability` instance itself, but our usage should be fine. Let me know what you think.

Source/NetworkReachabilityManager.swift
cnoon: Okay, thanks for confirming. That was my guess as to how it worked. I don't know how often users would have two sessions or managers. It would also be difficult to know which one was which. Probably not that useful.

Source/NetworkReachabilityManager.swift
cnoon: Seems good to me as long as `SCNetworkReachabilityGetFlags` is thread-safe since we're regenerating status each time we call it.

jshier: We don't ever call this API directly, only users will, by calling one of the `is` properties, or getting the `status` or `flags` directly. Our usage of `flags` comes from the notification callback on the background queue, which only accesses the `previousStatus` value when calling the `listener`

PR #2921, opened 2019-08-28
Add AF*Response, AFResult types, update tests to use AFError

jshier:
### Goals :soccer:
This PR adds `AFDataResponse` and `AFDownloadResponse`, which don't require an explicit `Failure` type, just defaulting to `AFError`. Additionally, `AFResult` has also been updated to expect an `AFError` `Failure` type as well, as many users have already transitioned to Swift 5's `Result`. These types are used to simplify some of the response APIs.

This PR also refactors a lot of tests to take advantage of the fact that `AFError` is now used directly in most cases and to be more explicit around different errors wrapped in `AFError`.

### Testing Details :mag:
No new tests, just refactors.

### DIFF
--- a/Source/NetworkReachabilityManager.swift
+++ b/Source/NetworkReachabilityManager.swift
@@ -99,17 +99,21 @@ open class NetworkReachabilityManager {
         return flags.map(NetworkReachabilityStatus.init) ?? .unknown
     }
 
-    /// A closure executed when the network reachability status changes.
-    private var listener: Listener?
-
-    /// `DispatchQueue` on which listeners will be called.
-    private var listenerQueue: DispatchQueue?
+    /// Mutable state storage.
+    struct MutableState {
+        /// A closure executed when the network reachability status changes.
+        var listener: Listener?
+        /// `DispatchQueue` on which listeners will be called.
+        var listenerQueue: DispatchQueue?
+        /// Previously calculated status.
+        var previousStatus: NetworkReachabilityStatus?
+    }
 
     /// `SCNetworkReachability` instance providing notifications.
     private let reachability: SCNetworkReachability
 
-    /// Protected storage for the previous status.
-    private let previousStatus = Protector<NetworkReachabilityStatus?>(nil)
+    /// Protected storage for mutable state.
+    private let mutableState = Protector(MutableState())
 
     // MARK: - Initialization
 
@@ -163,8 +167,10 @@ open class NetworkReachabilityManager {
                              onUpdatePerforming listener: @escaping Listener) -> Bool {
         stopListening()
 
-        listenerQueue = queue
-        self.listener = listener
+        mutableState.write { state in
+            state.listenerQueue = queue
+            state.listener = listener
+        }
 
         var context = SCNetworkReachabilityContext(version: 0,
                                                    info: Unmanaged.passRetained(self).toOpaque(),
@@ -195,9 +201,11 @@ open class NetworkReachabilityManager {
     open func stopListening() {
         SCNetworkReachabilitySetCallback(reachability, nil, nil)
         SCNetworkReachabilitySetDispatchQueue(reachability, nil)
-        previousStatus.write { $0 = nil }
-        listenerQueue = nil
-        listener = nil
+        mutableState.write { state in
+            state.listener = nil
+            state.listenerQueue = nil
+            state.previousStatus = nil
+        }
     }
 
     // MARK: - Internal - Listener Notification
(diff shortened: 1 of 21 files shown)


Source/NetworkReachabilityManager.swift
```diff
@@ -195,9 +201,11 @@ open class NetworkReachabilityManager {
     open func stopListening() {
         SCNetworkReachabilitySetCallback(reachability, nil, nil)
         SCNetworkReachabilitySetDispatchQueue(reachability, nil)
```
cnoon: Newline?

Tests/DownloadTests.swift
```diff
@@ -309,44 +305,35 @@ class DownloadResponseTestCase: BaseTestCase {
...
+        try "random_data".write(to: fileURL, atomically: true, encoding: .utf8)
 
-            // When
-            AF.download("https://httpbin.org/get", to: { _, _ in (fileURL, []) })
-                .response { resp in
-                    response = resp
-                    expectation.fulfill()
-                }
+        let expectation = self.expectation(description: "Download should complete but fail to move file")
+        var response: DownloadResponse<URL?, AFError>?
```
cnoon: You could probably collapse these as well if you wanted.

Tests/ParameterEncoderTests.swift
```diff
@@ -226,7 +226,7 @@ final class URLEncodedFormEncoderTests: BaseTestCase {
         let result = Result<String, Error> { try encoder.encode(parameters) }
 
         // Then
-        XCTAssertEqual(result.value, "a=a")
+        XCTAssertEqual(result.success, "a=a")
```
cnoon: This is a bit of a surprising change to me. Has `error` also been refactored to `failure`? I also think I feel these are less intuitive than the previous. `value` and `error`. Could you speak to this change a bit?

Tests/ResponseTests.swift
```diff
@@ -70,6 +70,8 @@ class ResponseTestCase: BaseTestCase {
         XCTAssertNil(response?.response)
         XCTAssertNil(response?.data)
         XCTAssertNotNil(response?.error)
+        XCTAssertEqual(response?.error?.isSessionTaskError, true)
+        XCTAssertEqual((response?.error?.underlyingError as? URLError)?.code, .cannotFindHost)
```
cnoon: Weird mismatch now between `response.error` and `result.failure`. I don't think I really like the `value` -> `success` and `error` to `failure` change. Those are the cases, not the values inside the cases. Seems like an odd choice to use the same name as a convenience property. Seems more like a Bool to me than a value.

Tests/TLSEvaluationTests.swift
```diff
@@ -143,11 +142,11 @@ class TLSEvaluationExpiredLeafCertificateTestCase: BaseTestCase {
 
         // Then
         XCTAssertNotNil(error, "error should not be nil")
-
-        if let error = error?.asAFError {
-            XCTAssertTrue(error.isServerTrustEvaluationError, "should be .certificatePinningFailed")
+        XCTAssertEqual(error?.isServerTrustEvaluationError, true)
```
cnoon: Newline? Same comment throughout this file.

Tests/ParameterEncoderTests.swift
jshier: As we discussed, these changes just make the testing `Result` extensions match the internal ones. The internal extensions have been renamed both to match likely naming from possibly language features, as well as preventing users from seeing the "has internal access level" error on the old properties when updating code.

Tests/ResponseTests.swift
jshier: Covered in Slack discussion.

Tests/DownloadTests.swift
jshier: Collapse what?

Tests/TLSEvaluationTests.swift
jshier: Added.

Source/NetworkReachabilityManager.swift
jshier: These are all statements to reset some state, so a separation doesn't seem necessary.

jshier:
@cnoon Updated.

Tests/DownloadTests.swift
cnoon: Sorry, `DownloadResponse` -> `AFDownloadResponse`.

Source/NetworkReachabilityManager.swift
cnoon: Ah no worries, I generally always put newlines in front of conditionals.

Tests/DownloadTests.swift
jshier: We could do this for all uses of the `Response` types. For now, we can keep them consistently the base type. If we want to use the AF types, we can make the changes all at once I think.

Issue #2925, opened 2019-08-30
ITMS-90806: CFBundleIdentifier collision


nikolas-theodosis:
As of yesterday, when trying to archive and upload a new binary in App Store, I get the following error:

_ITMS-90806: CFBundleIdentifier collision - Each bundle must have a unique bundle identifier. The bundle identifier 'org.cocoapods.Alamofire' is used in the bundles '[Alamofire.framework, Alamofire.framework]'_

A related issue was raised for other libraries [here](https://forums.developer.apple.com/thread/122048) suggesting that the library has the same bundle identifier for iOS and watchOS and one of them should be changed. 

## Alamofire Environment

**Alamofire version:** 4.8.2
**Xcode version:** 10.2
**Swift version:** 5.0
**Platform(s) running Alamofire:** iOS, watchOS

jshier:
This is a CocoaPods issue they're [already tracking](https://github.com/CocoaPods/CocoaPods/issues/9135). Apple playing games with bundle verification again.

JonMo:
@jshier, Not sure it is a CocoaPods issue only. We are using Carthage and having the same issue. It would just need to change the bundle identifier of the project to this `org.alamofire.Alamofire.${PLATFORM_NAME}`.
Thanks!

jshier:
Ah, true. We should probably change this on all platforms. I wonder how SPM will handle this.

jshier:
SPM won't need to handle this since it only supports static products.

cnoon:
I'm curious how we've been doing this at Nike for the past year. We've been shipping Alamofire on iOS and watchOS on NRC the whole time through CocoaPods. I'll do a bit of digging...

jshier:
Looks like a new warning, so unless you've done a release over the last few days, you wouldn't've seen it yet.

cnoon:
Ah that makes more sense. I think tacking on the platform name then as a suffix makes the most sense.

jshier:
This has been fixed for v4 and v5 and will go out in 4.9.0 and 5.0.0-rc.1.

PR #2928, opened 2019-09-03
Make bundle id per platform

JonMo:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/2925

### Goals :soccer:
Fixing an issue when submitting a binary to the App Store:
ITMS-90806: CFBundleIdentifier collision - Each bundle must have a unique bundle identifier. The bundle identifier 'org.cocoapods.Alamofire' is used in the bundles '[Alamofire.framework, Alamofire.framework]'

Add platform name to Project bundle identifier

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -1744,7 +1744,7 @@
 					"-framework",
 					CFNetwork,
 				);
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR! This looks good to me, we just need to make sure there're no unintended consequences. 

@cnoon Any comment?

jshier:
@JonMo Also, could you open another version of this PR against the `hotfix` branch so we can release it for v4 as well?

PR #2949, opened 2019-09-23
Fixed crash in Demo app

ozgur:
### Issue Link :link:

In demo, when you click on cell "Reachability Status", app crashes due to implicitly unwrapped property being **nil**.

Screen Shot 2019-09-23 at 15 31 37

### Testing Details :mag:

- Launch demo
- Tap on "Reachability Status"


jshier:
Your changes are equivalent to the original code.

PR #2953, opened 2019-09-27
feat (Response) : update debugDescription

rain2540:
add parameters info for debugDescription of DataResponse and DownloadResponse

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -134,6 +134,8 @@ extension DataResponse: CustomStringConvertible, CustomDebugStringConvertible {
         var output: [String] = []
 
         output.append(request != nil ? "[Request]: \(request!.httpMethod ?? "GET") \(request!)" : "[Request]: nil")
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR! It's unlikely we'll be doing another release of Alamofire 4, but I'll keep this PR around until 5 is fully released.

jshier:
@rain2540 Looks like we'll be doing a 4.9.1 release after all. If you can update your PR to use the `debugDescription` implementation from Alamofire 5, which includes the body data, we can make this PR part of that release.

rain2540:
I've updated my PR to use the debugDescription implementation from Alamofire 5, which includes the body data.

jshier:
Turns out your changes are incompatible with Alamofire 4. Please resubmit your PR with a building version. You can use the Alamofire 5 implementation as an example, but you can't copy the code directly.

rain2540:
@jshier I've resubmitted my PR #2976

Issue #2957, opened 2019-10-02
Alamofire 5: URL parameter ordering


DanijelHuis:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

Hi,

If I remember correctly, Alamofire 4 used to order URL parameters alphabetically. I am testing Alamofire 5 and it seems it orders them randomly now, which breaks URLCache.

Example:
If I run following code multiple times I get different results:
```
let parameters = ["a1" : "a1_value", "b1" : "b1_value"]
AF.request("https://httpbin.org/get", parameters: parameters).responseString()
```
...
"https://httpbin.org/get?a1=a1_value&b1=b1_value"
"https://httpbin.org/get?b1=b1_value&a1=a1_value"

Is this intentional? I tried providing string and array instead of dictionary but that doesn't work. I know I can put parameters manually in URL but this was more convenient.
Thank you!

## Alamofire Environment

**Alamofire version:** 5.0.0-rc2
**Xcode version:** Version 11.0 (11A420a)
**Swift version:** 5
**Platform(s) running Alamofire:** iOS
**macOS version running Xcode:** 10.14.6 (18G87)

jshier:
There is a new version of parameter encoding in Alamofire 5, `ParameterEncoder`, that encodes any `Encodable` type into requests. For URL form encoding, as you've done here, it uses our new `URLEncodedFormParameterEncoder` and its underlying `URLEncodedFormEncoder`. That type currently uses a `Dictionary` to store the encoded parameters, meaning it can have random order, due to how `Dictionary` works.

You can still trigger the old behavior by passing an `encoding` parameter, which will go down the old code path and have ordered behavior. 

I didn't realize `URLCache` couldn't handle parameters in different orders, so this will need to get fixed before release. Any help is appreciated.

DanijelHuis:
I am not sure if this will fix all possible cases, but it works for few examples I've written at the end.

in URLEncodedFormEncoder.swift, replace this:

```
    func serialize(_ object: [String: URLEncodedFormComponent]) -> String {
        var output: [String] = []
        for (key, component) in object {
            let value = serialize(component, forKey: key)
            output.append(value)
        }

        return output.joinedWithAmpersands()
    }
```
with this
```
    func serialize(_ object: [String: URLEncodedFormComponent]) -> String {
        var output: [String] = []
[...]

PR #2961, opened 2019-10-07
Stable encoded parameter order for URLEncodedFormEncoder

jshier:
### Issue Link :link:
#2957 

### Goals :soccer:
This PR refactors the implementation of `URLEncodedFormEncoder` to make the parameter encoding consistent between runs.

This PR swaps the previous `Dictionary` storage in `URLEncodedFormComponent`'s `object` case for a `[(key: String, value: URLEncodedFormComponent)]`, allowing for a constant order at runtime. Additionally, an `alphabetizeKeyValuePairs` parameter was added to `URLEncodedFormEncoder` which controls the sorting of encoded components as they're serialized into the final `String`.

This PR also turns off `SWIFT_DETERMINISTIC_HASHING` to ensure we're testing with randomized `Dictionary` ordering. Due to this PR it's no longer necessary to have it turned on.

### Testing Details :mag:
Tests were updated for new expected ordering and to expect exact matches.


Source/URLEncodedFormEncoder.swift
```diff
@@ -317,20 +324,24 @@ public final class URLEncodedFormEncoder {
...
     /// - Parameters:
-    ///   - arrayEncoding:     The `ArrayEncoding` to use. `.brackets` by default.
-    ///   - boolEncoding:      The `BoolEncoding` to use. `.numeric` by default.
-    ///   - dataEncoding:      The `DataEncoding` to use. `.base64` by default.
-    ///   - dateEncoding:      The `DateEncoding` to use. `.deferredToDate` by default.
-    ///   - keyEncoding:       The `KeyEncoding` to use. `.useDefaultKeys` by default.
-    ///   - spaceEncoding:     The `SpaceEncoding` to use. `.percentEscaped` by default.
-    ///   - allowedCharacters: The `CharacterSet` of allowed (non-escaped) characters. `.afURLQueryAllowed` by default.
-    public init(arrayEncoding: ArrayEncoding = .brackets,
+    ///   - alphabetizeKeyValuePairs: Whether or to sort the encoded key value pairs. `true` by default.
```
jshier:
```suggestion
    ///   - alphabetizeKeyValuePairs: Whether or not to sort the encoded key value pairs. `true` by default.
```

Source/URLEncodedFormEncoder.swift
```diff
@@ -294,11 +294,18 @@ public final class URLEncodedFormEncoder {
...
+                return "URLEncodedFormEncoder requires keyed root object. Received \(object) instead."
             }
         }
     }
 
+    /// Whether or not to sort the encoded key value pairs.
+    ///
+    /// - Note: This setting ensures a consistent ordering for all encodings of the same parameters. When set to `false`,
+    ///         encoded `Dictionary` values may have a different encoded order each time they're encoded due to
+    ///       ` Dictionary`'s random storage order.
```
cnoon: I think this `Note` is a bit outdated since you've switched over to a tuple array. Isn't it now going to be either the order the parameters were provided or alphabetically sorted?

jshier: No, this note refers to `Dictionary`s being encoded, which will still have a variable order between app launches. Let me know if there’s a better way to say that here.

PR #2965, opened 2019-10-14
Move Automatic Resume Behavior to After the First Response Handler is Added

jshier:
### Issue Link :link:
No specific issue, but this change makes using Alamofire APIs that "chain" onto `Request`s fully safe, rather than there being an implicit race between adding those behaviors and the request completing.

### Goals :soccer:
This PR moves the automatic resume behavior from within `Session`'s `Request` set up logic, which called `resume` right after a `URLSessionTask` is created, to after the appending of a response serializer. 

This PR adds an additional property to the internal `RequestDelegate` protocol to allow `Request`s to query whether `startRequestsImmediately` is set. Additionally, logic has been added in `appendResponseSerializer` to, while accessing `mutableState`, check whether the `Request` can transition to the `resumed` state, and, if so, enqueues a check for `startsImmediately` on the `underlyingQueue`. If that value is `true`, `resume()` will be called, which already handles duplicate `resume` calls.

### Testing Details :mag:
No additional tests were added, but a few were updated to ensure they call response handlers and resume the request.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -65,7 +65,7 @@ public class Request {
 
     // MARK: - Initial State
 
-    /// `UUID` prividing a unique identifier for the `Request`, used in the `Hashable` and `Equatable` conformances.
+    /// `UUID` providing a unique identifier for the `Request`, used in the `Hashable` and `Equatable` conformances.
     public let id: UUID
     /// The serial queue for all internal async actions.
     public let underlyingQueue: DispatchQueue
@@ -115,7 +115,7 @@ public class Request {
         var error: AFError?
     }
 
-    /// Protected `MutableState` value that provides threadsafe access to state values.
+    /// Protected `MutableState` value that provides thread-safe access to state values.
     fileprivate let protectedMutableState: Protector<MutableState> = Protector(MutableState())
 
     /// `State` of the `Request`.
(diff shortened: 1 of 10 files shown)


Source/Request.swift
```diff
@@ -478,6 +480,10 @@ public class Request {
             if mutableState.responseSerializerProcessingFinished {
                 underlyingQueue.async { self.processNextResponseSerializer() }
             }
+            
+            if mutableState.state.canTransitionTo(.resumed) {
+                underlyingQueue.async { if self.delegate?.startImmediately == true { self.resume() } }
+            }
```
cnoon: Genius! Too bad we didn't think of this months ago. Great work!

Source/Session.swift
```diff
@@ -922,26 +922,21 @@ open class Session {
...
                 rootQueue.async { request.didSuspendTask(task) }
-            case (_, .cancelled):
+            case .cancelled:
                 // Resume to ensure metrics are gathered.
                 task.resume()
                 task.cancel()
                 rootQueue.async { request.didCancelTask(task) }
-            case (_, .finished):
-                // Do nothing
-                break
```
cnoon: Great updates here. Thankfully the huge amount of tests we have for all these scenarios can confirm these changes are valid. In theory though, you've eliminated the entire race by simply waiting to resume the request until the response serializer has been added. This also fixes the race condition for hitting the URLCache faster than the response serializer is added.

PR #2989, opened 2019-10-27
Update README.md

Set78:


jshier:
This change doesn't seem to clarify anything, but thanks anyway!

Issue #2923, opened 2019-08-29
The Client Certificate Validation Support on 5.0.0-beta.*


liuwin7:
## What did you do?

ℹ The client certificate validation doesn't work.

## What did you expect to happen?

ℹ The client certificate validation can be supported.

## What happened instead?

ℹ Just a comment on the case `NSURLAuthenticationMethodClientCertificate`
 ```swift
// case NSURLAuthenticationMethodClientCertificate:
// Alamofire doesn't currently support client certificate validation.
```  
Is it permanent or temporary ?

## Alamofire Environment

**Alamofire version:** 5.0.0-beta.7
**Xcode version:** 10.3 (10G8)
**Swift version:** 5.0
**Platform(s) running Alamofire:** ios
**macOS version running Xcode:** 10.14.6 (18G87)

jshier:
No permanent, we just haven't seen much demand for it. If you have a design for the feature in mind, let us know, or better yet, open a PR with the implementation.

liuwin7:
What if the server wants the client certificate ?

PR #2993, opened 2019-11-05
The Client Certificate Support

liuwin7:
Add the `NSURLAuthenticationMethodClientCertificate` to the case, just like the `NSURLAuthenticationMethodNTLM`, `NSURLAuthenticationMethodNegotiate`.

### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/2923

### Goals :soccer:
The client certificate is supported in the last major version, and the performing is just use a function to get the credential which is attached to the target. So, I think the new version is no need to drop this feature.

Just add a case `NSURLAuthenticationMethodClientCertificate` to the `NSURLAuthenticationMethodHTTPBasic`, `NSURLAuthenticationMethodHTTPDigest` and etc.
So, when a `NSURLAuthenticationMethodClientCertificate` authentication challenge coming, the `SessionDelegate` will get the credential from the task or the credential storage.

### Testing Details :mag:
I setup a client credential to the credential storage for my custom protection space, and the TLS can be establish successfully.

### DIFF
--- a/Source/SessionDelegate.swift
+++ b/Source/SessionDelegate.swift
@@ -79,10 +79,8 @@ extension SessionDelegate: URLSessionTaskDelegate {
         switch challenge.protectionSpace.authenticationMethod {
         case NSURLAuthenticationMethodServerTrust:
             evaluation = attemptServerTrustAuthentication(with: challenge)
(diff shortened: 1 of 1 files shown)


liuwin7:
@jshier

liuwin7:
it's my pleasure.

PR #3003, opened 2019-11-19
Response test host error fix

tronthomas:
### Issue Link :link:
For certain unit tests, an expected host connection failure can occur for different reasons depending on how the test is run. Previously the test were only checking for one host connection failure condition which would cause these test to fail on certain systems

### Goals :soccer:
Allow unit tests which expect host connection failures to succeed on all systems.

Modified tests so they can accept from multiple reasons for why the host connection may have failed

### Testing Details :mag:
Tested locally by running the unit tests, full verification will need to happen in the continuous integration phase

### DIFF
--- a/Tests/ResponseTests.swift
+++ b/Tests/ResponseTests.swift
@@ -71,7 +71,7 @@ final class ResponseTestCase: BaseTestCase {
         XCTAssertNil(response?.data)
         XCTAssertNotNil(response?.error)
         XCTAssertEqual(response?.error?.isSessionTaskError, true)
-        XCTAssertEqual(response?.hasHostURLError, true)
+        XCTAssertEqual(response?.error?.isHostURLError, true)
         XCTAssertNotNil(response?.metrics)
     }
 }
@@ -124,7 +124,7 @@ final class ResponseDataTestCase: BaseTestCase {
         XCTAssertNil(response?.data)
         XCTAssertEqual(response?.result.isFailure, true)
         XCTAssertEqual(response?.error?.isSessionTaskError, true)
-        XCTAssertEqual(response?.hasHostURLError, true)
+        XCTAssertEqual(response?.error?.isHostURLError, true)
         XCTAssertNotNil(response?.metrics)
     }
 }
@@ -177,7 +177,7 @@ final class ResponseStringTestCase: BaseTestCase {
         XCTAssertNil(response?.data)
         XCTAssertEqual(response?.result.isFailure, true)
         XCTAssertEqual(response?.error?.isSessionTaskError, true)
-        XCTAssertEqual(response?.hasHostURLError, true)
+        XCTAssertEqual(response?.error?.isHostURLError, true)
         XCTAssertNotNil(response?.metrics)
     }
 }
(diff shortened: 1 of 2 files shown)


Tests/ResponseTests.swift
```diff
@@ -672,8 +672,27 @@ final class ResponseTryMapErrorTestCase: BaseTestCase {
...
 
         XCTAssertEqual(underlyingError.asAFError?.isSessionTaskError, true)
-        XCTAssertEqual((underlyingError.asAFError?.underlyingError as? URLError)?.code, .cannotFindHost)
-
+        XCTAssertEqual(underlyingError.asAFError?.isHostURLError, true)
         XCTAssertNotNil(response?.metrics)
     }
 }
+
+private extension DataResponse where Failure == AFError {
```
jshier: We prefer extensions on the `AFError` type itself, not the response type, otherwise we need separate extensions for every response.

Tests/ResponseTests.swift
```diff
@@ -672,8 +672,27 @@ final class ResponseTryMapErrorTestCase: BaseTestCase {
...
+    var hasHostURLError: Bool? {
+        return error?.isHostURLError
+    }
+}
+
+private extension AFError {
```
jshier: You can put this in the `AFError+AlamofireTests.swift` file alongside our other `is*` convenience properties.

Additionally, you can simplify the property down to: 
```swift
return [.cannotConnectToHost, .cannotFindHost].contains((underlyingError as? URLError)?.code)
```

Tests/ResponseTests.swift
```diff
@@ -672,8 +672,27 @@ final class ResponseTryMapErrorTestCase: BaseTestCase {
...
+    var isHostURLError: Bool {
+        return urlHostErrorCodes.contains(urlErrorCode)
+    }
+    
+    var urlErrorCode : URLError.Code {
```
jshier: I don't think a separate property is necessary here, and it's not really appropriate to fall back to the `.unknown` case rather than using `nil`.

Tests/ResponseTests.swift
```diff
@@ -672,8 +672,27 @@ final class ResponseTryMapErrorTestCase: BaseTestCase {
...
+        return (underlyingError as? URLError)?.code ?? .unknown
+    }
+}
+
+private var urlHostErrorCodes : Set<URLError.Code> {
```
jshier: You can inline this check pretty easily, so no need to have a separate declaration.

Tests/ResponseTests.swift
```diff
@@ -518,7 +518,7 @@ final class ResponseTryMapTestCase: BaseTestCase {
         XCTAssertNil(response?.data)
         XCTAssertEqual(response?.result.isFailure, true)
         XCTAssertEqual(response?.error?.asAFError?.isSessionTaskError, true)
-        XCTAssertEqual((response?.error?.asAFError?.underlyingError as? URLError)?.code, .cannotFindHost)
+        XCTAssertTrue(urlHostErrorCodes.contains(response!.error!.asAFError!.urlErrorCode))
```
jshier: This should be structured similar to the previous line, where we compare the value optionally.

tronthomas:
I do not really understand what has caused the Test SPM Integration failure. The details link offers no useful information. From what little I've been able to research, SPM seems to refer to Swift Package Manager, only I have no idea what the actual issue with the package manager is or how to fix it. How can I find more information on the cause of the failure and how to correct it?

PR #3015, opened 2019-12-10
add index style serialization array nested dictionary

lyj98:
### Issue Link :link:

#1839 #2431 
### Goals :soccer:

There is no real specification for how to URL encode collection parameters like arrays and
dictionaries.

for: `["foo": ["a", 1, true, ["bar": 2], ["qux": 3],  ["quy": ["quz": 3]]]]`

1. default `.brackets` encode is:

` foo[]=a&foo[]=1&foo[]=1&foo[][bar]=2&foo[][qux]=3&foo[][quy]`[quz]=3

2. `.noBrackets` encode is: 

`foo=a&foo=1&foo=1&foo[bar]=2&foo[qux]=3&foo[quy][quz]=3`

3. [jQuery](https://api.jquery.com/jQuery.param/#example-1) >=1.4 and Node.js package  [query-string](https://www.npmjs.com/package/query-string) provide index style. 

`foo[0]=a&foo[1]=1&foo[2]=1&foo[3][bar]=2&foo[4][qux]=3&foo[5][quy][quz]=3`

A new `ArrayEncoding` enumeration is added to `URLEncoding`. The cases are `.brackets` and `.noBrackets` and `.index `
The `arrayEncoding` passed to `init` is used to control the array key serialization in `URLEncoding.queryComponents`.

### Testing Details :mag:

The new tests are named like the existing tests, with an appended suffix:

- `testURLParameterEncodeArrayNestedDictionaryValueParameterWithIndex`

tests pass.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -85,14 +85,19 @@ public struct URLEncoding: ParameterEncoding {
         case brackets
         /// No brackets are appended. The key is encoded as is.
         case noBrackets
+        
+        case index
 
-        func encode(key: String) -> String {
+        func encode(key: String, index: Int) -> String {
             switch self {
             case .brackets:
                 return "\(key)[]"
             case .noBrackets:
                 return key
+            case .index:
+                return "\(key)[\(index)]"
             }
+            
         }
     }
 
@@ -193,8 +198,8 @@ public struct URLEncoding: ParameterEncoding {
                 components += queryComponents(fromKey: "\(key)[\(nestedKey)]", value: value)
             }
         } else if let array = value as? [Any] {
-            for value in array {
-                components += queryComponents(fromKey: arrayEncoding.encode(key: key), value: value)
+            for (index, value) in array.enumerated() {
+                components += queryComponents(fromKey: arrayEncoding.encode(key: key, index: index), value: value)
             }
         } else if let value = value as? NSNumber {
             if value.isBool {
--- a/Tests/ParameterEncodingTests.swift
+++ b/Tests/ParameterEncodingTests.swift
@@ -205,6 +205,22 @@ class URLParameterEncodingTestCase: ParameterEncodingTestCase {
             XCTFail("Test encountered unexpected error: \(error)")
         }
     }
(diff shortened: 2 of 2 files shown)


Source/ParameterEncoding.swift
```diff
@@ -85,14 +85,19 @@ public struct URLEncoding: ParameterEncoding {
         case brackets
         /// No brackets are appended. The key is encoded as is.
         case noBrackets
+        
+        case index
```
jshier: Please add inline documentation, similar to the documentation for the other cases.

jshier:
You can also merge master to pick up more reliable tests.

Source/ParameterEncoding.swift
```diff
@@ -85,13 +85,17 @@ public struct URLEncoding: ParameterEncoding {
         case brackets
         /// No brackets are appended. The key is encoded as is.
         case noBrackets
+        /// Index in square brackets ("key[index]")
```
jshier: This should be phrased similarly to the other cases. Something like "Brackets containing the item index appended. This matches the jQuery and Node.js behavior."

Source/ParameterEncoding.swift
```diff
@@ -85,13 +85,17 @@ public struct URLEncoding: ParameterEncoding {
...
+        case index
```
jshier: Perhaps this case would be clearer as `indexInBrackets`?

Tests/ParameterEncoderTests.swift
```diff
@@ -485,6 +485,54 @@ final class URLEncodedFormEncoderTests: BaseTestCase {
         // Then
         XCTAssertFalse(result.isSuccess)
     }
+    
+    func testThatEncodableStructCanBeEncodedWithIndex() {
```
jshier: This isn't testing a struct but a superclass.

jshier: Additionally, please add test cases for the other encodable types, like the struct, and the encodable subclass.

dhoom56 (approved):

PR #3044, opened 2020-01-26
Map DataResponse success result by a key path

RomanPodymov:
Hello.
Thank you for Alamofire.

### Goals :soccer:
Found the following example for `func map<NewSuccess>(_ transform: (Success) -> NewSuccess) -> DataResponse<NewSuccess, Failure>` in the documentation for this function:

```
let possibleData: DataResponse<Data> = ...
let possibleInt = possibleData.map { $0.count }
```

As for me, there should be also `let possibleInt = possibleData.map(\.count)`.

Added a new function `func map<NewSuccess>(_ keyPath: KeyPath<Success, NewSuccess>) -> DataResponse<NewSuccess, Failure>` to `DataResponse`. This function is similar to `func map<NewSuccess>(_ transform: (Success) -> NewSuccess) -> DataResponse<NewSuccess, Failure>`, the only difference is the mapping function.

### Testing Details :mag:
Added `ResponseMapTestCase.testThatMapTransformsSuccessValueByKeyPath`.

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -143,6 +143,7 @@ extension DataResponse {
                                                  result: result.map(transform))
     }
 
+    #if !swift(>=5.2)
     /// Similar to `func map<NewSuccess>(_ transform: (Success) -> NewSuccess) -> DataResponse<NewSuccess, Failure>`,
     /// but uses a key path instead of a closure
     ///
@@ -163,6 +164,7 @@ extension DataResponse {
                                                  serializationDuration: serializationDuration,
                                                  result: result.map { $0[keyPath: keyPath] } )
     }
+    #endif
 
     /// Evaluates the given closure when the result of this `DataResponse` is a success, passing the unwrapped result
     /// value as a parameter.
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -143,7 +143,6 @@ extension DataResponse {
                                                  result: result.map(transform))
     }
 
-    #if !swift(>=5.2)
     /// Similar to `func map<NewSuccess>(_ transform: (Success) -> NewSuccess) -> DataResponse<NewSuccess, Failure>`,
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! This is very useful functionality but I don't think your PR is necessary, as we're going to get this functionality for free, automatically, as part of [Swift 5.2](https://github.com/apple/swift-evolution/blob/master/proposals/0249-key-path-literal-function-expressions.md). This would allow us to use key paths *everywhere*, not just in specific APIs, and is a much better solution going forward. 

I'll leave this PR open until I can confirm that this will function the way we want using the 5.2 snapshots (they're currently not building Alamofire's tests) but ultimately I don't think it will be necessary.

RomanPodymov:
Hello @jshier 
I think not all Alamofire users will switch to Swift 5.2 right after it's release. So I can wrap the code from this pull request into `#if !swift(>=5.2) #endif` to make this feature available for all users.

jshier:
No, we wouldn't want to make this API conditionally available (which would be `swift(<5.2)`), due to the maintenance burden that would impose. Users on 5.2 will just get it for free when they update.

jshier:
Swift 5.2 is included in the new Xcode 11.4 beta release yesterday, so I've finally be able to test the new automatic `KeyPath` to function conversion. It works like you'd expect (though there's no autocomplete). I modified one of our tests to see what it looks like:
```swift
AF.request("https://httpbin.org/post", method: .post, parameters: parameters, encoder: JSONParameterEncoder.default)
    .responseDecodable(of: HTTPBinResponse.self) { response in
        receivedResponse = response
        originResponse = response.map(\.origin)
        expect.fulfill()
    }
```

So thanks for your PR, but I think the best course now is to let users who adopt Swift 5.2 get the feature for free, without any work from Alamofire.

PR #3058, opened 2020-02-10
add Alamofire namespace to DataResponse and DownloadResponse debugDes…

embassem:
Update DataResponse debugDescription

### Goals :soccer:
Update log message of DataResponse and DownloadResponse to include `[Alamofire]` namespace at the beginning of the line, that will help search and filter the log message, like other libraries do such as Firebase SDKs

### DIFF
--- a/Source/Response.swift
+++ b/Source/Response.swift
@@ -98,8 +98,8 @@ extension DataResponse: CustomStringConvertible, CustomDebugStringConvertible {
             let sortedHeaders = response.headers.sorted()
 
             return """
-            [Status Code]: \(response.statusCode)
-            [Headers]:
+            [Alamofire] [Status Code]: \(response.statusCode)
+            [Alamofire] [Headers]:
             \(sortedHeaders)
             """
         } ?? "nil"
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR!

`debugDescription` isn't a logging description, it's what's seen in the debugger when printing the variable, so I'm not sure it's appropriate to attempt to namespace it. This would be something better handling in an actual logger, if Alamofire actually shipped one (I can see the value, if we can constrain the feature set). For example, this would be entirely unnecessary if logging through `OSLog`, as `OSLog` already has the notion of namespaces internally, allowing users to print what they want and capturing the namespace as part of the data recorded. You can see this when viewing logs in the Console app. 

So ultimately I don't think this is a good fit for the library and should be part of whatever logging solution consumers come up with, or a logger that's provided by Alamofire. 

@cnoon Any thoughts?

embassem:
what I meant is in order to log request to console  and not use a 3rd party for that we may  do it like that 
```
.responseObject { (response: DataResponse<Response<T>,AFError>) in
                print(response)
//...
        }
```
or when want more details, can use `debugPrint(response)`
but the log in the console will be lost among all other logs.

jshier:
I understand the desire here, as well as your implementation, but I don't think it's appropriate to add extraneous information to the debug description like this. If you need a log marker you can add them yourself. Your implementation from the PR can be done fairly simply in code:
```swift
response.debugDescription.split(separator: "\n").map { "[Alamofire]: \($0)" }.joined(separator: "\n")
```
Whereas, if this change were adopted, anyone using a dedicated logging system that can automatically separate logging sources would get a bunch of text they didn't want, obscuring what they're trying to log.

In essence, what you want is easier to add programmatically than it is to remove, so it probably shouldn't be added.

Issue #3072, opened 2020-02-22
Add DataStreamRequest


jshier:
`DataStreamRequest` is a replacement for Alamofire 4's `stream` closure API. It encapsulates a `URLSessionDataTask` just like `DataRequest`, but instead of accumulating the `Data` in memory, it passes it through a closure API. This issues tracks its development.

Features:
- [x] Streaming closure API
- Support existing Alamofire `Request` APIs
  - [x] Retry
  - [x] Validation of response codes.
  - [x] Validation of content types.
- [x] Abstracted streaming serialization support.
- [x] Produce an `InputStream` of incoming `Data`.
- [x] Clean up stream APIs.
- [x] Inline documentation.
- [x] Add additional `EventMonitor` events for streaming.
- [x] Usage and Advanced Usage documentation.

jshier:
`DataStreamRequest` has been merged.

PR #3074, opened 2020-02-23
Add DataStreamRequest

jshier:
### Issue Link :link:
Completes #3072

### Goals :soccer:
This PR implements `DataStreamRequest`, a replacement for Alamofire 4's `stream` handler.

`DataStreamRequest` is a new `Request` subclass that uses a new serialization method to stream incoming `Data` to an arbitrary number of closures.

### Testing Details :mag:
Tests have been added.


Source/Request.swift
```diff
@@ -1034,6 +1034,232 @@ public class DataRequest: Request {
...
+                self.responseSerializerDidComplete {
+                    queue.async {
+                        do {
+                            let completion = Completion(request: self.request,
+                                                        response: self.response,
+                                                        metrics: self.metrics,
+                                                        error: self.error)
+                            try stream(.init(event: .complete(completion), token: .init(self)))
+                        } catch {
+                            // What to do with error?
```
cnoon: Missing a solution here?

Source/Request.swift
```diff
@@ -1034,6 +1034,232 @@ public class DataRequest: Request {
...
+        guard case let .stream(result) = event, case let .failure(error) = result else { return nil }
+
+        return error
+    }
+
+    public var completion: DataStreamRequest.Completion? {
+        guard case let .complete(completion) = event else { return nil }
+
+        return completion
+    }
```
cnoon: Were these intended to be `public`?

Source/SessionDelegate.swift
```diff
@@ -49,12 +49,14 @@ open class SessionDelegate: NSObject {
...
+//        }
+//
+//        return request
 
-        guard let request = provider.request(for: task) as? R else {
-            fatalError("Returned Request is not of expected type: \(R.self).")
-        }
-
-        return request
+        return provider.request(for: task) as? R
```
cnoon: Is this a straight replacement? If yes, can we remove the commented out code?

Source/Validation.swift
```diff
@@ -184,6 +191,34 @@ extension DataRequest {
...
+    public func validate<S: Sequence>(contentType acceptableContentTypes: @escaping @autoclosure () -> S) -> Self where S.Iterator.Element == String {
+        return validate { [unowned self] _, response in
+            self.validate(contentType: acceptableContentTypes(), response: response)
+        }
+    }
+
+    public func validate() -> Self {
+        validate(statusCode: acceptableStatusCodes).validate(contentType: self.acceptableContentTypes)
+    }
+}
```
cnoon: 👍🏻

Source/Request.swift
jshier: Yeah, I'm not sure what to do with an error thrown during the completion event. My inclination is to intentionally ignore it since it doesn't make much sense.

Source/Request.swift
jshier: Yes, just a bit of convenience so users don't have to switch all the time, especially in the `Never` error cases.

Source/SessionDelegate.swift
jshier: Yep, needed to be able to check for multiple types. Removed in https://github.com/Alamofire/Alamofire/pull/3074/commits/a12ab983b06efeccf77d0ec86d02e85f4a99b92a.

Tests/DataStreamTests.swift
```diff
@@ -0,0 +1,749 @@
...
+                    switch result {
+                    case let .success(data):
+                        secondAccumulatedData.append(data)
+                    }
+                case let .complete(completion):
+                    secondCompleteOnMain = Thread.isMainThread
+                    secondResponse = completion.response
+                    secondExpectation.fulfill()
+                }
+            }
```
cnoon: Pretty cool that this is supported. I'm not sure what an actual use case would be, but I'm sure someone will do something really cool with it.

Tests/DataStreamTests.swift
```diff
@@ -0,0 +1,749 @@
...
+
+        // When
+        let stream = AF.streamRequest(URLRequest.makeHTTPBinRequest(path: "xml", headers: [.contentType("application/xml")]))
+            .responseStream { stream in
+                switch stream.event {
+                case .complete:
+                    expect.fulfill()
+                default: break
+                }
+            }.asInputStream()
```
cnoon: Will we crash if you don't put `responseStream` before `asInputStream`? All your docs do it in this order, but I'm not sure it's called out explicitly.

Tests/DataStreamTests.swift
```diff
@@ -0,0 +1,749 @@
...
+        let expect = expectation(description: "stream complete")
+
+        // When
+        AF.streamRequest(URLRequest.makeHTTPBinRequest(path: "bytes/50"), automaticallyCancelOnStreamError: true)
+            .responseStreamDecodable(of: HTTPBinResponse.self) { stream in
+                switch stream.event {
+                case let .complete(completion):
+                    complete = completion
+                    response = completion.response
+                    expect.fulfill()
```
cnoon: What exactly trips the error in this test? Not obvious where it comes from. Just curious.

Tests/DataStreamTests.swift
```diff
@@ -0,0 +1,749 @@
...
+        // Given
+        let expectedSize = 1000
+        var error: AFError?
+        let expect = expectation(description: "stream should complete")
+
+        // When
+        AF.streamRequest("https://httpbin.org/bytes/\(expectedSize)").responseStream { stream in
+            switch stream.event {
+            case .stream:
+                stream.token.cancel()
```
cnoon: Is there really a need to expose the `token` if you're exposing the `cancel` API on `stream`?

Tests/DataStreamTests.swift
```diff
@@ -0,0 +1,749 @@
...
+        var completeOnMain = false
+        let redirected = expectation(description: "stream redirected")
+        let redirector = Redirector(behavior: .modify { _, _, _ in
+            redirected.fulfill()
+            return URLRequest.makeHTTPBinRequest(path: "stream/1")
+        })
+        let expect = expectation(description: "stream complete")
+
+        // When
+        AF.streamRequest(URLRequest.makeHTTPBinRequest(path: "status/301"))
```
cnoon: Nice.

Source/ResponseSerialization.swift
```diff
@@ -798,3 +798,151 @@ extension DownloadRequest {
...
+    ///   - dataPreprocessor: `DataPreprocessor` used to process incoming `Data` before it's passed through the `decoder`.
+    public init(decoder: DataDecoder = JSONDecoder(), dataPreprocessor: DataPreprocessor = PassthroughPreprocessor()) {
+        self.decoder = decoder
+        self.dataPreprocessor = dataPreprocessor
+    }
+
+    public func serialize(_ data: Data) throws -> T {
+        let processedData = try dataPreprocessor.preprocess(data)
+        do {
+            return try decoder.decode(T.self, from: processedData)
```
cnoon: How do you know you have enough data to actually run the decode pass safely here? Not really sure how this actually works.

Source/Validation.swift
```diff
@@ -184,6 +191,34 @@ extension DataRequest {
     }
 }
 
+extension DataStreamRequest {
+    public typealias Validation = (_ request: URLRequest?, _ response: HTTPURLResponse) -> ValidationResult
+
+    public func validate<S: Sequence>(statusCode acceptableStatusCodes: S) -> Self where S.Iterator.Element == Int {
```
cnoon: Missing `@discardableResult`.

Source/Validation.swift
```diff
@@ -184,6 +191,34 @@ extension DataRequest {
...
+    ///
+    /// - returns: The request.
+    @discardableResult
+    public func validate<S: Sequence>(contentType acceptableContentTypes: @escaping @autoclosure () -> S) -> Self where S.Iterator.Element == String {
+        return validate { [unowned self] _, response in
+            self.validate(contentType: acceptableContentTypes(), response: response)
+        }
+    }
+
+    public func validate() -> Self {
```
cnoon: Missing `@discardableResult`.

Source/Request.swift
```diff
@@ -1082,6 +1082,236 @@ public class DataRequest: Request {
...
+    public func asInputStream(bufferSize: Int = 1024) -> InputStream {
+        var inputStream: InputStream?
+        $streamMutableState.write { state in
+            Foundation.Stream.getBoundStreams(withBufferSize: bufferSize,
+                                              inputStream: &inputStream,
+                                              outputStream: &state.outputStream)
+            state.outputStream?.open()
+        }
+
+        return inputStream!
```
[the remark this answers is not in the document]
jshier: I was going to check `swift-corelibs-foundation` to see when it could be `nil` there but it's unimplemented, with this comment:

> // Discussion of this API is ongoing for its usage of AutoreleasingUnsafeMutablePointer

I'm guessing there some sort strangeness going on due to the Apple side of this API bridging to `CFStream` and the pointer shenanigans that has to happen there. Personally, I don't think there's any way this could fail, but I can't say it's guaranteed. I'll make the return value optional, it just makes it far more awkward to use.

Tests/DataStreamTests.swift
jshier: Returning 50 random bytes won't properly parse as an `HTTPBinResponse`, so the serializer produces an error.

Tests/DataStreamTests.swift
jshier: I imagined the `token` could be saved off separately so users could pass it around. I'm not sure how likely that is but I thought the flexibility might be useful.

Source/ResponseSerialization.swift
jshier: We don't. Unlike web sockets, this sort of streaming is completely unstructured and at the mercy of how `URLSession` receives the `Data`. I'm pretty sure `Data` which fits entirely within a single MTU frame will come through unbroken, and I've confirmed that to be the case with some streaming servers that stream small JSON payloads. But larger payloads likely won't work, but I didn't want to build a whole custom accumulator that would then feed into serializers, as the complexity of that would be very high. Plus, that scenario would be better covered by a web socket, which has automatic framing built in to make that sort of thing much easier as I understand it (though I'm working to confirm that behavior). 

Ultimately, if users find their `Data` is too big for the automatic handling, I think they could accumulate the `Data` themselves and trigger parsing using the serializers manually.

Source/Validation.swift
jshier: Addressed in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107.

Source/Validation.swift
jshier: Addressed in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107.

Tests/DataStreamTests.swift
jshier: It wouldn't crash, but it also wouldn't start the request. I've updated the logic in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107 to address that issue, please take a look.

Tests/DataStreamTests.swift
```diff
@@ -0,0 +1,749 @@
...
+        var response: HTTPURLResponse?
+        var streamOnMain = false
+        var completeOnMain = false
+        let expect = expectation(description: "stream should complete")
+
+        // When
+        AF.streamRequest(URLRequest.makeHTTPBinRequest(path: "bytes/\(expectedSize)")).responseStream { stream in
+            switch stream.event {
+            case let .stream(result):
+                streamOnMain = Thread.isMainThread
```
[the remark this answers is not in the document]
jshier: So far the test data isn't so big that multiple calls are needed. I could request more data, but since the handling of the stream side is unified, a single check should be all that's needed. I can add more if you think it's important.

Source/Session.swift
```diff
@@ -322,6 +322,86 @@ open class Session {
...
+
+    /// Creates a `DataStreamRequest` from the passed `URLRequestConvertible` value and `RequestInterceptor`.
+    ///
+    /// - Parameters:
+    ///   - convertible: `URLRequestConvertible` value to be used to create the `URLRequest`.
+    ///   - interceptor: `RequestInterceptor` value to be used by the returned `DataRequest`. `nil` by default.
+    ///
+    /// - Returns:       The created `DataStreamRequest`.
+    open func streamRequest(_ convertible: URLRequestConvertible,
+                            automaticallyCancelOnStreamError: Bool = false,
```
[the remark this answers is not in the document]
jshier: Addressed in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107.

Source/Session.swift
```diff
@@ -322,6 +322,86 @@ open class Session {
...
+    ///   - convertible: `URLConvertible` value to be used as the `URLRequest`'s `URL`.
+    ///   - method:      `HTTPMethod` for the `URLRequest`. `.get` by default.
+    ///   - headers:     `HTTPHeaders` value to be added to the `URLRequest`. `nil` by default.
+    ///   - interceptor: `RequestInterceptor` value to be used by the returned `DataRequest`. `nil` by default.
+    ///
+    /// - Returns:       The created `DataStream` request.
+    open func streamRequest(_ convertible: URLConvertible,
+                            method: HTTPMethod = .get,
+                            headers: HTTPHeaders? = nil,
+                            automaticallyCancelOnStreamError: Bool = false,
```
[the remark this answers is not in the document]
jshier: Addressed in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107.

Source/Session.swift
```diff
@@ -322,6 +322,86 @@ open class Session {
...
+    ///   - headers:     `HTTPHeaders` value to be added to the `URLRequest`. `nil` by default.
+    ///   - interceptor: `RequestInterceptor` value to be used by the returned `DataRequest`. `nil` by default.
+    ///
+    /// - Returns:       The created `DataStream` request.
+    open func streamRequest<Parameters: Encodable>(_ convertible: URLConvertible,
+                                                   method: HTTPMethod = .get,
+                                                   parameters: Parameters? = nil,
+                                                   encoder: ParameterEncoder = URLEncodedFormParameterEncoder.default,
+                                                   headers: HTTPHeaders? = nil,
+                                                   automaticallyCancelOnStreamError: Bool = false,
```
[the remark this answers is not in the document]
jshier: Addressed in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107.

Source/Request.swift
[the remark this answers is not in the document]
jshier: Addressed in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107.

Source/Request.swift
```diff
@@ -1082,6 +1082,236 @@ public class DataRequest: Request {
...
+    ///   - convertible:        `URLRequestConvertible` value used to create `URLRequest`s for this instance.
+    ///   - underlyingQueue:    `DispatchQueue` on which all internal `Request` work is performed.
+    ///   - serializationQueue: `DispatchQueue` on which all serialization work is performed. By default targets
+    ///                         `underlyingQueue`, but can be passed another queue from a `Session`.
+    ///   - eventMonitor:       `EventMonitor` called for event callbacks from internal `Request` actions.
+    ///   - interceptor:        `RequestInterceptor` used throughout the request lifecycle.
+    ///   - delegate:           `RequestDelegate` that provides an interface to actions not performed by the `Request`.
+    init(id: UUID = UUID(),
+         convertible: URLRequestConvertible,
+         automaticallyCancelOnStreamError: Bool,
```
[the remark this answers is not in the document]
jshier: Addressed in https://github.com/Alamofire/Alamofire/pull/3074/commits/52e4f454339fac1e5ad7009e631820f7c9aad107.

Source/Request.swift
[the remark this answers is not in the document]
cnoon: Fair enough. Thanks for digging in that deep. 👍🏻

Tests/DataStreamTests.swift
cnoon: Ah thanks. Was thinking too hard on it and missed the obvious.

Source/ResponseSerialization.swift
cnoon: Okay cool. That is what I "assumed" it was built for, but wanted to confirm. So this is much more of a "use at your own risk" type of feature.

Tests/DataStreamTests.swift
cnoon: Looks good. The optional input stream is a bit clunky, but seems like a better tradeoff. I think the `defer { resume() }` makes sense as well.

Tests/DataStreamTests.swift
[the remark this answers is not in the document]
cnoon: No that's fine. Just thought it may be warranted. Sounds like it's not.

PR #3081, opened 2020-02-24
Fix Swift PM Semantic Version

eseay:
### Goals :soccer:
Resolve a package manifest parsing error generated by the Swift PM integration snippet from the README.

When the existing package dependency description is added to the manifest, Xcode throws the following error:

```
Failed to parse the manifest file

Invalid semantic version string '5.0'
```

Added two characters to the README.

Per the Apple docs on [Version](https://developer.apple.com/documentation/swift_packages/version), the `patch` value is required.

### Testing Details :mag:
- Copy/paste the old dependency snippet into your `Package.swift` - ❌ 
- Copy/paste the new dependency snippet into your `Package.swift` - ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Once you have your Swift package set up, adding Alamofire as a dependency is as 
 
 ```swift
 dependencies: [
-    .package(url: "https://github.com/Alamofire/Alamofire.git", from: "5.0.0")
+    .package(url: "https://github.com/Alamofire/Alamofire.git", .upToNextMajor(from: "5.0.0"))
 ]
 ```
 
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Once you have your Swift package set up, adding Alamofire as a dependency is as 
 
 ```swift
 dependencies: [
-    .package(url: "https://github.com/Alamofire/Alamofire.git", from: "5.0")
+    .package(url: "https://github.com/Alamofire/Alamofire.git", from: "5.0.0")
 ]
 ```
 


jshier:
My concern here is ensuring that SPM users entering `5.0.0` like that actually get updates to the next major version, similar to Xcode 11's default SPM integration. Do you know if that's the case, or do we need to do something specific to enable that behavior in the snippet?

eseay:
@jshier From what I can tell, there are two dependency specifiers that are actually equivalent:
`.package(url: "https://github.com/Alamofire/Alamofire.git", .upToNextMajor(from: "5.0.0"))`
and
`.package(url: "https://github.com/Alamofire/Alamofire.git", from: "5.0.0")`

The former is actually my preferred means of specifying dependencies and matches the default specification from the Xcode UI. If you'd prefer to go this route, I'd be happy to update my PR accordingly. I chose to just keep the existing `from` implementation to minimize the change. 

In either case, I understand their functionality based on the ["Discussion" section of the docs](https://developer.apple.com/documentation/swift_packages/package/dependency/2878262-package). If I am interpreting your question correctly, I don't believe there is a way to specify the dependency such that some minor adjustment would be required to projects' `Package.swift` when Alamofire 6.0.0 eventually comes out.

jshier:
I think adding `.upToNextMajor` would be a good idea so users know the behavior they're getting.

eseay:
Alright, @jshier - I've updated the line according to our discussion. Let me know if there is anything else you'd like to discuss.

Issue #2971, opened 2019-10-15
Restful API in background mode not working!


yasirIqbal11:
* [ ] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

ℹ I want to call rest api in background mode in ios.   

class Networking {
    static let sharedInstance = Networking()
    public var sessionManager: Alamofire.SessionManager // most of your web service clients will call through sessionManager
    public var backgroundSessionManager: Alamofire.SessionManager // your web services you intend to keep running when the system backgrounds your app will use this
    
    private init() {
        self.sessionManager = Alamofire.SessionManager(configuration: URLSessionConfiguration.default)
    self.backgroundSessionManager = Alamofire.SessionManager(configuration: URLSessionConfiguration.background(withIdentifier: "com.nestedlogics.background"))
    self.backgroundSessionManager.startRequestsImmediately = true
    }
}

In AppDelegate.swift

func application(_ application: UIApplication, handleEventsForBackgroundURLSession identifier: String, completionHandler: @escaping () -> Void) {
        
        Networking.sharedInstance.backgroundSessionManager.backgroundCompletionHandler = completionHandler
    }

Also in capabilities I have enabled Background Fetch!

## What did you expect to happen?

ℹ  It should complete rest API request in background mode without any error.

## What happened instead?

ℹ The request callback give me failure: in result 
Error code -999 cancelled.

## Alamofire Environment

**Alamofire version:*4.8.1*
**Xcode version:*10.2.1*
**Swift version:*5*
**Platform(s) running Alamofire:*ios 13.1*
**macOS version running Xcode:*10.15 Beta*

## Demo Project

ℹ Please link to or upload a project we can download that reproduces the issue.

jshier:
Alamofire isn't able to survive the background transition, so if your app is killed while in the background the `SessionManager` will take its underlying `URLSession` down with it, cancelling any active `URLSessionTask`s that are in flight and producing the error you see. This is an unfortunately limitation of Alamofire's architecture we hope to remedy some time in the future.

You don't really need to use a background session configuration to ensure your ongoing requests are finished when your app transitions to the background. Apple has a variety of APIs to help with. I've found [this summary documentation](https://forums.developer.apple.com/thread/85066) to be the most help in implementing the background task assertion APIs.

yasirIqbal11:
Thanks!

PR #3094, opened 2020-03-02
Clarify dropping background support 5.0 migration

teameh:
First of all, congrats on the major release! Thanks a lot for the hard work to make it happen. 

I just migrated our app to v5. Everything went quite smooth except a few usages of background sessions. After some searching I've found [this answer](https://github.com/Alamofire/Alamofire/issues/2971#issuecomment-543847712) and [this PR](https://github.com/Alamofire/Alamofire/pull/2917) indicating that this is no longer possible.

I thought it would it be good to reflect this in the migration guide as well, so I've added a quote from the PR. Please feel free to change the text or write something completely different. 

Btw, how is this behaviour compared to v4? Was it never working, not reliable, or is this due to the new architecture?

### DIFF
--- a/Documentation/Alamofire 5.0 Migration Guide.md
+++ b/Documentation/Alamofire 5.0 Migration Guide.md
@@ -22,7 +22,7 @@ This guide is provided in order to ease the transition of existing applications 
 Most APIs have changed in Alamofire 5, so this list is not complete. While most top level `request` APIs remain the same, nearly every other type has changed in some way. For up to date examples, see our Usage and Advanced Usage documentation.
 
 - `SessionManager` has been renamed to `Session` and its APIs have completely changed.
-- Using a `URLSessionConfiguration` with a background identifier is not possible any more. We're explicitly ensuring Alamofire isn't used with background sessions, in order to prevent ongoing issues around support and surprise on the part of the user.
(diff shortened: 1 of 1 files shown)


Documentation/Alamofire 5.0 Migration Guide.md
```diff
@@ -22,6 +22,7 @@ This guide is provided in order to ease the transition of existing applications
 Most APIs have changed in Alamofire 5, so this list is not complete. While most top level `request` APIs remain the same, nearly every other type has changed in some way. For up to date examples, see our Usage and Advanced Usage documentation.
 
 - `SessionManager` has been renamed to `Session` and its APIs have completely changed.
+- Using a `URLSessionConfiguration` with a background identifier is not possible any more. We're explicitly ensuring Alamofire isn't used with background sessions, in order to prevent ongoing issues around support and surprise on the part of the user.
```
jshier:
```suggestion
- Background `URLSessionConfiguration`s are no longer supported. Alamofire was never designed to work in the background and its closure-based APIs cannot survive a background transition, leading to ongoing issues around background behavior. Explicit background support will be added through dedicated APIs at some point in the future.
```

teameh:
Yes that is perfect, thanks @jshier

Issue #1935, opened 2017-01-24
Alamofire on Linux Status


jshier:
We are making good progress towards getting Alamofire building on Linux (checkout the `linux` branch to see) but we've run into a big issue. Aside from a few constants and other minor changes that can be worked around, the Linux version of Swift is missing the `Security` framework from Apple's platforms entirely. So all of the functionality that Alamofire's `ServerTrustPolicy` depends on does not exist on the Linux platform. There are a few options here:

1. Refactor Alamofire to decouple `ServerTrustPolicy` from `SessionManger`. This will be a bit of work (meaning time) and will mean the the Linux version doesn't have any SSL verification beyond what `URLSession` provides.
1a. In addition to decoupling `ServerTrustPolicy` and `SessionManager`, abstract `ServerTrustPolicy` enough that Linux devs could implement it themselves using whatever dependency they wanted.
2. Adopt an external dependency for Linux to provide said features and abstract the `ServerTrustPolicy` to use either implementation depending on the running platform.
3. Similar to 2, but also replacing the implementation of `ServerTrustPolicy` on Apple platforms to use the dependency on all platforms.
4. Nothing. No port of Alamofire to Linux until the Swift team can provide a useful dependency that comes with the language or default Swift installs.

Personally I lean towards 1, as we've always worked to not have any external dependencies in Alamofire. However, this would be a major refactor and may not ship until Alamofire 5. 2 or 3 have the benefit of being a behind the scenes change but would likely be even more work than whatever changes we want for AF5 and so may take even longer.

However, we wanted to open this discussion up to public feedback to gather input and suggestions from the community before we choose a path and move forward.

codeOfRobin:
Quick question: Has Apple shown any indication that they'll be porting `Security.framework` to Linux?

jshier:
Not AFAIK, but they are working on server side libraries that may cover some of the same things. Whether those would be suitable for what we need I don't know and it's unlikely those libraries will be usable until after WWDC this year. I'm going to email the Swift users list and see if anyone wants to add anything.

codeOfRobin:
Ah, got it. FWIW, the Vapor (and Kitura guys) had to reimplement HTTP clients, either through libCURL or from scratch. And they used OpenSSL for the security stuff IIRC

https://github.com/IBM-Swift/Kitura-Request
https://vapor.github.io/documentation/http/client.html

jshier:
[...]

PR #3098, opened 2020-03-05
WIP Linux Support

SlaunchaMan:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/1935

### Goals :soccer:
Get Alamofire 5 building on Linux

There were some main things I had to do to get this to build. Some of them represent work that is yet to be done:

- [X] Import `FoundationNetworking` where needed.
- [X] Implement a lock using `pthread_mutex` functions.
- [X] Hide all of the `SecTrust` API usage behind `#ifdef` flags.
- [ ] Implement security methods on Linux, placing the remaining `SecTrust` dependencies behind an abstraction.
- [X] Use `#ifdef` to hide all `URLSessionTaskMetrics` references, [pending their addition to the open-source Foundation code](https://bugs.swift.org/browse/SR-11518).
- [x] Replace CoreFoundation code that handles encoding parsing.
- [ ] Replace Cocoa/Cocoa Touch code that does things like finding the executable name and version from the bundle.
- [x] Implement NSNumber type inspection as in [this comment](https://forums.swift.org/t/alamofire-on-linux-possible-but-not-release-ready/34553/6?u=slaunchaman).
- [ ] Replace `NSStream.getBoundStreams(withBufferSize:inputStream:outputStream:)` on Linux with similar function.

### Testing Details :mag:
I haven't started on anything test-related yet, as the test suite is not yet a part of the Swift package.


SlaunchaMan:
To try this out, run `docker build .` from your Alamofire directory.

Source/ParameterEncoding.swift
```diff
@@ -310,5 +318,7 @@ public struct JSONEncoding: ParameterEncoding {
 // MARK: -
 
 extension NSNumber {
+    #if !os(Linux)
     fileprivate var isBool: Bool { return CFBooleanGetTypeID() == CFGetTypeID(self) }
```
SlaunchaMan: I don't think emulating this will be possible on Linux, as [the Swift team](https://github.com/apple/swift-evolution/blob/master/proposals/0170-nsnumber_bridge.md#alternatives-considered) declined to keep `NSNumber` as strongly type-preserving.

Source/ResponseSerialization.swift
```diff
@@ -214,7 +226,7 @@ extension DataRequest {
         -> Self {
         appendResponseSerializer {
             // Start work that should be on the serialization queue.
-            let start = CFAbsoluteTimeGetCurrent()
+            let start = Date().timeIntervalSinceReferenceDate
```
SlaunchaMan: If we need more precision than this allows, we’ll have to find an alternative for `CFAbsoluteTimeGetCurrent` on Linux.

Source/ResponseSerialization.swift
```diff
@@ -514,6 +569,9 @@ public final class StringResponseSerializer: ResponseSerializer {
 
         data = try dataPreprocessor.preprocess(data)
 
+        #if os(Linux)
+        let actualEncoding = String.Encoding.isoLatin1
```
SlaunchaMan: Here we’ll need a Linux equivalent to `CFStringConvertIANACharSetNameToEncoding`.

SlaunchaMan: I found this: https://opensource.apple.com/source/CF/CF-476.18/CFStringUtilities.c.auto.html

Source/ParameterEncoding.swift
jshier: We should likely just ignore this older API on Linux entirely and only support `ParameterEncoder`s.

Source/ResponseSerialization.swift
jshier: I don't think we should import any outside code to support this on Linux. Instead, we could try to create a more Swift native cross platform solution.

Source/ResponseSerialization.swift
jshier: [This thread](https://forums.swift.org/t/recommended-way-to-measure-time-in-swift/33326) on the Swift Forums also provides some guidance.

Source/ResponseSerialization.swift
SlaunchaMan: Added a new initializer for `String.Encoding` that takes an IANA charset name. Right now it supports the same names as the code it replaces, but we could also add missing ones from [the list](https://www.iana.org/assignments/character-sets/character-sets.xhtml).

jshier:
@SlaunchaMan We've added test to the `Package.swift` file and updated our CI. Take a look if you want to try getting tests working. I think you'll need to add `--enable-test-discovery` to pick up the tests on Linux.

SlaunchaMan:
@jshier Tests now run on Linux, but for some reason the Docker container doesn't appear to be able to connect to a network. I’ll look into that some more but for now, enjoy the multitude of test failures!

SlaunchaMan:
I’ve filed [SR-12336](https://bugs.swift.org/browse/SR-12336) about this.

Source/Protector.swift
```diff
@@ -40,36 +70,46 @@ final class UnfairLock {
...
-    ///
-    /// - Returns:           The value the closure generated.
-    func around<T>(_ closure: () -> T) -> T {
-        lock(); defer { unlock() }
-        return closure()
+/// A `pthread_mutex_t` wrapper.
+final class MutexLock: Lock {
+    private var mutex: pthread_mutex_t
+    
+    init() {
```
jshier: As a side note, I think the handling of the mutex here needs to be directly allocated, so avoid similar issues seen with `os_unfair_lock`, as mentioned on [the Swift Forums](https://forums.swift.org/t/atomic-property-wrapper-for-standard-library/30468/18).

SlaunchaMan: Looks like `os_unfair_lock` is an UnsafeMutablePointer, but `pthread_mutex_t` is not, so I can’t call `.allocate(1)` like you do for the former.

weissi: random drive-by comment: @jshier is right, you will need to allocate the memory for `phread_mutex_t` on the heap manually. Swift doesn't guarantee that `&mutex` is always the same pointer so this lock implementation might randomly fail. Also, it fails to check the `pthread_*` return codes. See SwiftNIO's [lock implementation](https://github.com/apple/swift-nio/blob/master/Sources/NIOConcurrencyHelpers/lock.swift#L34) for something that should work.

SlaunchaMan: Thanks @weissi! I’ve updated the implementation to allocate manually and check for errors.

Source/ResponseSerialization.swift
SlaunchaMan: After reading that thread, I’ve replaced this call with `ProcessInfo.processInfo.systemUptime`, which seems to work. I doubt this code needs nanosecond precision.

SlaunchaMan:
Switching to the Swift nightly builds yielded some interesting results! Much more works, and now the only tests that end in failure are related to authentication or caching. Also, the mysterious crash now has a log:

<details>
<summary>Test execution log</summary>

```
Test Case 'SessionTestCase.testThatSessionCallsResponseSerializerCompletionsWhenAdapterThrowsErrorDuringRetry' started at 2020-03-21 03:39:48.283
Precondition failed: Attempt to add or subtract invalid fraction: file /home/buildnode/jenkins/workspace/oss-swift-package-linux-ubuntu-18_04/swift-corelibs-foundation/Sources/Foundation/ProgressFraction.swift, line 59
Current stack trace:
0    libswiftCore.so                    0x00007fa7e7d1bbe0 swift_reportError + 50
1    libswiftCore.so                    0x00007fa7e7d8d6b0 _swift_stdlib_reportFatalErrorInFile + 115
2    libswiftCore.so                    0x00007fa7e7a6b481 <unavailable> + 1389697
3    libswiftCore.so                    0x00007fa7e7a6b0c7 <unavailable> + 1388743
4    libswiftCore.so                    0x00007fa7e7a6b662 <unavailable> + 1390178
5    libswiftCore.so                    0x00007fa7e7a69a90 _assertionFailure(_:_:file:line:flags:) + 523
6    libFoundation.so                   0x00007fa7e867a818 <unavailable> + 8267800
7    libFoundation.so                   0x00007fa7e867bcf0 static _ProgressFraction.+ infix(_:_:) + 183
8    libFoundation.so                   0x00007fa7e8674090 Progress._overallFraction.getter + 231
9    libFoundation.so                   0x00007fa7e8670f80 Progress.totalUnitCount.setter + 944
10   AlamofirePackageTests.xctest       0x000056207ff3e1db <unavailable> + 713179
11   AlamofirePackageTests.xctest       0x000056207ff434cf <unavailable> + 734415
12   AlamofirePackageTests.xctest       0x000056207ff3c805 <unavailable> + 706565
13   AlamofirePackageTests.xctest       0x000056207ffa8878 <unavailable> + 1149048
14   AlamofirePackageTests.xctest       0x000056207ffa866e <unavailable> + 1148526
15   AlamofirePackageTests.xctest       0x000056207fefe650 <unavailable> + 452176
16   libdispatch.so                     0x00007fa7e6d16407 <unavailable> + 140295
17   libdispatch.so                     0x00007fa7e6d21c29 <unavailable> + 187433
18   libdispatch.so                     0x00007fa7e6d228ce <unavailable> + 190670
19   libdispatch.so                     0x00007fa7e6d2a522 <unavailable> + 222498
20   libpthread.so.0                    0x00007fa7e77006db <unavailable> + 30427
21   libc.so.6                          0x00007fa7e6682850 clone + 63
Exited with signal code 4
```

</details>

Looks like it has something to do with a `Progress`. I added `print()` statements to everywhere we set `totalUnitCount`, but I haven’t figured it out yet. Still, progress!

Source/Request.swift
```diff
@@ -557,10 +557,8 @@ public class Request {
     func reset() {
         error = nil
 
-        uploadProgress.totalUnitCount = 0
-        uploadProgress.completedUnitCount = 0
-        downloadProgress.totalUnitCount = 0
-        downloadProgress.completedUnitCount = 0
+        uploadProgress = Progress(totalUnitCount: 0)
```
jshier: Great if this fixes the crash, but it makes `Progress` useless, no?

SlaunchaMan: This is in the `reset()` function, so I’m not sure why the total and completed unit count would both need to be set to 0 instead of the progress object just recreated.

In any case yes, this appears to fix the crash.

jshier: Don't you have to use a single `Progress` instance for it to work? If I'm observing the `Progress` instance, or composing it in other `Progress` values, doesn't it need to stay the same instance to work? I've never actually used this API, but I can't see any other way for it to work.

jshier: In any case, this is another nice bug to report.

SlaunchaMan: Yep, I’ll try to build a small reproduction.

SlaunchaMan: Filed as [SR-12391](https://bugs.swift.org/browse/SR-12391).

Source/HTTPHeaders.swift
```diff
@@ -368,12 +368,20 @@ extension HTTPHeader {
...
+        
+        #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
         let executable = (info?[kCFBundleExecutableKey as String] as? String) ??
             (ProcessInfo.processInfo.arguments.first?.split(separator: "/").last.map(String.init)) ??
             "Unknown"
         let bundle = info?[kCFBundleIdentifierKey as String] as? String ?? "Unknown"
         let appVersion = info?["CFBundleShortVersionString"] as? String ?? "Unknown"
         let appBuild = info?[kCFBundleVersionKey as String] as? String ?? "Unknown"
+        #else
+        let executable = "Unknown"
```
jshier, line 380: Which of the constants aren't available? We could just make them literals.

Tests/CacheTests.swift
```diff
@@ -83,8 +83,13 @@ class CacheTestCase: BaseTestCase {
 
         urlCache = {
             let capacity = 50 * 1024 * 1024 // MBs
-            #if targetEnvironment(macCatalyst)
-            return URLCache(memoryCapacity: capacity, diskCapacity: capacity)
+            // swiftformat:disable indent
+            #if swift(>=5.1) && !os(Linux)
```
jshier, line 87: I think the `targetEnvironment` check should work here.

Tests/CachedResponseHandlerTests.swift
```diff
@@ -199,10 +199,20 @@ final class CachedResponseHandlerTestCase: BaseTestCase {
         let configuration = URLSessionConfiguration.af.default
         let capacity = 100_000_000
         let cache: URLCache
+        // swiftformat:disable indent
+        #if swift(>=5.1) && !os(Linux)
```
jshier, line 203: Same here, the `targetEnvironment` check should work.

Tests/CacheTests.swift
jshier: Looks like this also broke the Catalyst build in general.

SlaunchaMan: Fixed!

Source/HTTPHeaders.swift
SlaunchaMan: `kCFBundleExecutableKey`, `kCFBundleIdentifierKey`, and `kCFBundleVersionKey`, but I think it’s fairly unlikely that a Linux app would have those in an `infoDictionary`.

jshier: Sure, but they can be added, so I think it would be good to have minimal Linux-specific behavior here and just use the fallbacks.

SlaunchaMan: These are now just literals. I think that it’s pretty unlikely they’ll change.

Source/AFError.swift
```diff
@@ -220,6 +220,10 @@ public enum AFError: Error {
     case sessionInvalidated(error: Error?)
     /// `URLSessionTask` completed with error.
     case sessionTaskFailed(error: Error)
+    /// An `OutputStream` returned an error during a write operation.
+    case streamError(Error?)
```
jshier, line 224: We can't add error cases here until the next major version, FYI. For minor releases a new error type in the custom case could be used.

jshier: And this may be something we need in deployment, so a separate PR for the change would probably be good.

SlaunchaMan: Yep, I have this pulled out in a separate branch and will put it in a PR shortly. To avoid making this a breaking issue I can avoid adding a new error case.

jshier:
This work has moved to #3446. Thanks for the pioneering effort!

PR #3103, opened 2020-03-08
Fix RetryPolicy's Handling of AFError

jshier:
### Goals :soccer:
This PR fixes `RetryPolicy`'s handling of `AFError` that was broken after we moved to using `AFError` for all errors.

Error code extraction logic in `RetryPolicy` was updated to handle both the `AFError` and direct `URLError` cases.

### Testing Details :mag:
Tests were updated to use `AFError` instances, as we'd normally see.


jshier:
Yes, I'll add some additional tests for both the `URLError` and `AFError` cases and then some usage tests.

PR #3110, opened 2020-03-14
Fix watchOS Memory Leaks

### LINKED ISSUE #3106: Memory retained and never released. urlSession::didFinishCollecting never called on watchOS

## What did you do?

I am making 100s of network requests. My test is ~600 requests in a loop. I see the memory usage continue to go up and not return. I see in Xcode's Memory Graph that there are 1000s of DataRequest objects still allocated, and this number goes up with each run of the test.

I see that there is an object holding all the DataRequest objects: `Dictionary<Request, NSURLSessionTask>` and that object holds all the requests.

Memory to not increase while making network requests. Mainly, the memory usage to return after all the requests are completed.

## What happened instead?

Memory grows as network requests are made. The graph is steady during use. 

I only see the issue on watchOS.

## Alamofire Environment

**Alamofire version:**
github "Alamofire/Alamofire" "5.0.2"
github "Alamofire/AlamofireImage" "4.0.3"
**Xcode version:**
Reproduced on 11.1 and 11.3.1
**Swift version:**
Xcode 11.1: Apple Swift version 5.1 (swiftlang-1100.0.270.13 clang-1100.0.33.7)
**Platform(s) running Alamofire:**
iOS and watchOS
**macOS version running Xcode:**
10.15.3

jshier:
### Issue Link :link:
Fixes #3106.

### Goals :soccer:
This PR fixes memory leaks seen on watchOS by working around the fact that watchOS never collects `URLSessionTaskMetrics` for some reason.

This PR conditionalizes the previously added `disassociateIfNecessary` check on watchOS so that metrics are never expected.

### Testing Details :mag:
Still no testing on watchOS, but a watchOS example as been added so we can more easily use Alamofire on watchOS in the future.

### DIFF
--- a/Alamofire.xcworkspace/contents.xcworkspacedata
+++ b/Alamofire.xcworkspace/contents.xcworkspacedata
@@ -7,4 +7,7 @@
    <FileRef
       location = "group:Example/iOS Example.xcodeproj">
    </FileRef>
+   <FileRef
+      location = "group:watchOS Example/watchOS Example.xcodeproj">
+   </FileRef>
 </Workspace>
--- a/Documentation/AdvancedUsage.md
+++ b/Documentation/AdvancedUsage.md
@@ -363,6 +363,8 @@ AF.request(...)
     }
 ```
 
+> Due to `FB7624529`, collection of `URLSessionTaskMetrics` on watchOS is currently disabled.
+
 ### `DataRequest`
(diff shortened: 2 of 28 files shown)


doublek420:
Sent via the Samsung Galaxy S8+, an AT&T 5G Evolution capable smartphone
Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #3114, opened 2020-03-15
Use version 2 of the checkout action.

SlaunchaMan:
### Goals :soccer:
This PR updates the GitHub Actions to use version 2 of the `checkout` action.


SlaunchaMan:
Looks like the biggest change is only checking out one branch, so it should be faster.

jshier:
Oh, nice.

PR #3115, opened 2020-03-16
Linux Preparation

jshier:
### Goals :soccer:
This PR implements some changes to make Alamofire more cross platform friendly.

Changes in this PR:
- Addition of `MutexLock` and `Lock` protocol abstraction to make `Protected` available.
- Replacement of `CFAbsoluteTimeGetCurrent` with `ProcessInfo.processInfo.systemUptime`.
- Updates to `User-Agent` generation to make it work around missing values more elegantly.
- Replacement of `arc4random()` with `UInt32.random`.
- Add a `typealias` for the events in `RequestTaskMap`.
- Replacement of `CFString` IANA encoding handling with internal implementation.
- Replacement of `CFType` checking with direct Obj-C type checking.

### Testing Details :mag:
`User-Agent` tests were updated for changed results in tests.

Adding @SlaunchaMan as a reviewer since he's been doing the Linux port work.


Source/Protected.swift
```diff
@@ -24,10 +24,64 @@
...
+    fileprivate func lock() {
+        pthread_mutex_lock(mutex)
+    }
+
+    fileprivate func unlock() {
+        pthread_mutex_unlock(mutex)
+    }
+}
+
+#if !os(Linux)
```
SlaunchaMan: This might be better specified as a positive check for Apple platforms to better support arbitrary future platforms.
```suggestion
#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
```

Source/Protected.swift
```diff
@@ -40,38 +94,26 @@ final class UnfairLock {
...
 /// A thread-safe wrapper around a value.
 @propertyWrapper
 @dynamicMemberLookup
 final class Protected<T> {
-    private let lock = UnfairLock()
+    #if os(Linux)
+    private let lock: Lock = MutexLock()
+    #else
+    private let lock: Lock = UnfairLock()
+    #endif
```
SlaunchaMan: As before, doing this without calling out Linux specifically would look like this:
```suggestion
    #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
    private let lock = UnfairLock()
    #else
    private let lock = MutexLock()
    #endif
```

Source/Protected.swift
jshier: If we're going to do that we'd want to be specific about Linux as well.

Source/Protected.swift
```diff
@@ -24,10 +24,64 @@
...
+    }
+
+    fileprivate func lock() {
+        pthread_mutex_lock(mutex)
+    }
+
+    fileprivate func unlock() {
+        pthread_mutex_unlock(mutex)
+    }
+}
```
SlaunchaMan:
```suggestion
#if os(Linux)
final class MutexLock: Lock {
    private var mutex: UnsafeMutablePointer<pthread_mutex_t>

    init() {
        mutex = .allocate(capacity: 1)

        var attr = pthread_mutexattr_t()
        pthread_mutexattr_init(&attr)
        pthread_mutexattr_settype(&attr, .init(PTHREAD_MUTEX_ERRORCHECK))

        let err = pthread_mutex_init(mutex, &attr)

        precondition(err == 0, "Failed to create pthread_mutex")
    }

    deinit {
        pthread_mutex_destroy(mutex)
    }

    fileprivate func lock() {
        pthread_mutex_lock(mutex)
    }

    fileprivate func unlock() {
        pthread_mutex_unlock(mutex)
    }
}
#endif
```

Source/Protected.swift
SlaunchaMan: Updated to do exactly that. Compiling this on any other platform will fail to compile without a `lock` implemented in the `Protected` type.

SlaunchaMan:
💯

Source/MultipartFormData.swift
```diff
@@ -55,7 +55,10 @@ open class MultipartFormData {
         }
 
         static func randomBoundary() -> String {
-            String(format: "alamofire.boundary.%08x%08x", arc4random(), arc4random())
+            let first = UInt32.random(in: UInt32.min...UInt32.max)
+            let second = UInt32.random(in: UInt32.min...UInt32.max)
+
+            return String(format: "alamofire.boundary.%08x%08x", first, second)
```
SlaunchaMan: Random thought: when I reimplemented this using `UInt32.random`, I kept the original logic of two different integers getting formatted, but it sure seems like this could be _one_ integer that gets formatted twice as wide.

Source/Protected.swift
```diff
@@ -24,10 +24,65 @@
...
+    /// - Returns:           The value the closure generated.
+    func around<T>(_ closure: () -> T) -> T {
+        lock(); defer { unlock() }
+        return closure()
+    }
+
+    /// Execute a closure while acquiring the lock.
+    ///
+    /// - Parameter closure: The closure to run.
+    func around(_ closure: () -> Void) {
```
SlaunchaMan: Is there a reason to include this version? The project builds and the tests pass without it; I’m assuming the compiler can infer `T` as `Void` if it needs to.

jshier: My only thought is that the unspecialized version may optimize where the generic fails. Otherwise you're right, they're functionally identical AFAICT. I'll leave it for now pending investigation.

Source/MultipartFormData.swift
SlaunchaMan: Hmm, never mind. There’s not a built-in format specifier for a 64-bit integer in hex format.

Source/HTTPHeaders.swift
```diff
@@ -367,45 +367,42 @@ extension HTTPHeader {
...
-                let alamofireVersion = "Alamofire/\(version)"
-
-                return "\(executable)/\(appVersion) (\(bundle); build:\(appBuild); \(osNameVersion)) \(alamofireVersion)"
-            }
-
-            return "Alamofire"
+        let info = Bundle.main.infoDictionary
+        let executable = (info?[kCFBundleExecutableKey as String] as? String) ??
+            (ProcessInfo.processInfo.arguments.first?.split(separator: "/").last.map(String.init)) ??
+            "Unknown"
```
cnoon: Wow, that's quite a mouthful! 😏

Source/Protected.swift
```diff
@@ -40,38 +95,25 @@ final class UnfairLock {
...
 
 /// A thread-safe wrapper around a value.
 @propertyWrapper
 @dynamicMemberLookup
 final class Protected<T> {
+    #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
     private let lock = UnfairLock()
+    #elseif os(Linux)
+    private let lock = MutexLock()
+    #endif
```
cnoon: Well done.

Source/RequestTaskMap.swift
```diff
@@ -26,9 +26,11 @@ import Foundation
 
 /// A type that maintains a two way, one to one map of `URLSessionTask`s to `Request`s.
 struct RequestTaskMap {
+    private typealias Events = (completed: Bool, metricsGathered: Bool)
+
     private var tasksToRequests: [URLSessionTask: Request]
     private var requestsToTasks: [Request: URLSessionTask]
-    private var taskEvents: [URLSessionTask: (completed: Bool, metricsGathered: Bool)]
+    private var taskEvents: [URLSessionTask: Events]
```
cnoon: Just an improvement? This doesn't affect Linux does it?

Source/ResponseSerialization.swift
```diff
@@ -516,10 +516,8 @@ public final class StringResponseSerializer: ResponseSerializer {
 
         var convertedEncoding = encoding
 
-        if let encodingName = response?.textEncodingName as CFString?, convertedEncoding == nil {
-            let ianaCharSet = CFStringConvertIANACharSetNameToEncoding(encodingName)
-            let nsStringEncoding = CFStringConvertEncodingToNSStringEncoding(ianaCharSet)
-            convertedEncoding = String.Encoding(rawValue: nsStringEncoding)
+        if let encodingName = response?.textEncodingName, convertedEncoding == nil {
+            convertedEncoding = String.Encoding(ianaCharsetName: encodingName)
```
cnoon: Very cool.

Source/RequestTaskMap.swift
SlaunchaMan: In my Linux PR, this typealias is defined to not include metrics; making it an alias makes that a little easier for the rest of the API surface.

PR #3116, opened 2020-03-16
Prevent Multiple finish() Calls

jshier:
### Goals :soccer:
This PR fixes a bug discovered during the development of `DataStreamRequest` where a request that finishes after a state change enqueues a `finish()` call can call `finish()` more than once. That is:
- `cancel()` or other state change called.
- Underlying `task`'s `state` is **not** `.completed` and so a call to `finish()` is enqueued. 
- At the same time, the `task` is actually completed.
- `retryOrFinish` is called and since there's no `error` or no `retriers`, it calls `finish()` directly.
- Serializers are started.
- Enqueued `finish()` arrives and `finish` calls the serializers again, leading to duplicate calls.

This PR adds an additional `Bool` property to `MutableState` and checks it when `finish()` is called. It's then reset when all of the serializers are run. I think this should really be part of the `State` enum so we can properly transition between the states, but since `State` is public we can't make the change just yet.

### Testing Details :mag:
Added an additional test file, `InternalRequestTests`, which we can use to test internal logic like this, as well as a test. If additional tests can be though of that would be great.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -113,6 +113,9 @@ public class Request {
         var retryCount = 0
         /// Final `AFError` for the `Request`, whether from various internal Alamofire calls or as a result of a `task`.
         var error: AFError?
+        /// Whether the instance has had `finish()` called and is running the serializers. Should be replaced with a
+        /// representation in the state machine in the future.
+        var isFinishing = false
     }
 
     /// Protected `MutableState` value that provides thread-safe access to state values.
@@ -456,6 +459,10 @@ public class Request {
     ///
     /// - Parameter error: The possible `Error` with which the instance will finish.
     func finish(error: AFError? = nil) {
+        guard !protectedMutableState.directValue.isFinishing else { return }
+
+        protectedMutableState.directValue.isFinishing = true
+
         if let error = error { self.error = error }
 
         // Start response handlers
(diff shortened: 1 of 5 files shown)


Source/Request.swift
```diff
@@ -556,7 +564,10 @@ public class Request {
         downloadProgress.totalUnitCount = 0
         downloadProgress.completedUnitCount = 0
 
-        protectedMutableState.write { $0.responseSerializerCompletions = [] }
+        protectedMutableState.write { state in
+            state.isFinishing = false
+            state.responseSerializerCompletions = []
+        }
```
cnoon: This is definitely key for retry to work correctly. 👍🏻

Source/Request.swift
```diff
@@ -113,6 +113,9 @@ public class Request {
         var retryCount = 0
         /// Final `AFError` for the `Request`, whether from various internal Alamofire calls or as a result of a `task`.
         var error: AFError?
+        /// Whether the instance has had `finish()` called and is running the serializers. Should be replaced with a
+        /// representation in the state machine in the future.
+        var isFinishing = false
```
cnoon: I agree, this would be a good state change to make in the future. This will certainly work for now.

PR #3137, opened 2020-03-30
Add RequestModifier for Top Level APIs

jshier:
### Goals :soccer:
This PR adds the `RequestModifier` type and adds it to `Session`'s top level API for creating `Request`s. It allows the customization of `URLRequest`s created from the parameters of a request before adaptation. It does *not* allow the modification of `URLRequestConvertible` parameters before adaptation.

`RequestModifier` is defined as `(inout URLRequest) throws -> Void`, which allows users to modify the request value in place while also allowing errors. For simplicity it's integrated into the `URLRequestConvertible` values created from the individual parameters rather than lower down in the stack.

### Testing Details :mag:
A simple test has been added, more are needed for all request types.

TODO:
- [x] Add more tests for each `Request` type, including error handling and retry.
- [x] Decide whether it's the last parameter of `download`, as it would break existing trailing closure usage.
- [x] Investigate whether it should be further down the stack to allow modification of `URLRequestConvertible` values.
- [x] Update relevant documentation.


Source/Session.swift
```diff
@@ -238,6 +243,8 @@ open class Session {
     /// - Parameters:
     ///   - convertible: `URLConvertible` value to be used as the `URLRequest`'s `URL`.
     ///   - method:      `HTTPMethod` for the `URLRequest`. `.get` by default.
+    ///   - timeout:     Time, in seconds, request will spend waiting for an initial response from the server. 60 by
+    ///                  default.
```
cnoon: Docstring.

Source/Session.swift
```diff
@@ -220,15 +220,20 @@ open class Session {
...
         let method: HTTPMethod
         let parameters: Parameters?
         let encoding: ParameterEncoding
         let headers: HTTPHeaders?
+        let requestModifier: RequestModifier?
 
         func asURLRequest() throws -> URLRequest {
-            let request = try URLRequest(url: url, method: method, headers: headers)
+            var request = try URLRequest(url: url, method: method, headers: headers)
+            try requestModifier?(&request)
```
cnoon: This is exactly what I had envisioned @jshier. Looks solid so far with what you have. I don't really follow the reasoning of pointing out that you're not modifying the original `URLRequestConvertible` data. I don't really know why you would. Could you elaborate a bit more on what you're thinking there?

jshier: I just mean that the modifiers only apply to the APIs which take the request components, not to the API which takes `URLRequestConvertible` values directly.

cnoon: Gotcha, that completely makes sense. If anyone is providing `URLRequestConvertible` values directly, they should create `requestModifier` types of their own. Exactly like I'm doing in my internal library.

Source/Session.swift
```diff
@@ -765,6 +825,8 @@ open class Session {
     ///   - interceptor:             `RequestInterceptor` value to be used by the returned `DataRequest`. `nil` by default.
     ///   - fileManager:             `FileManager` to be used if the form data exceeds the memory threshold and is
     ///                              written to disk before being uploaded. `.default` instance by default.
+    ///   - requestModifier:         `RequestModifier` which will be applied to the `URLRequest` created from the
+    ///                              provided parameters. `nil` by default.
     ///
     /// - Returns:                   The created `UploadRequest`.
     open func upload(multipartFormData: @escaping (MultipartFormData) -> Void,
```
cnoon, line 832: Missing `requesModifier` here?

jshier: Will fix in the versioning PR.

PR #3162, opened 2020-04-21
Changed DisabledEvaluator to DisabledTrustEvaluator

MrCloud:
### Goals :soccer:
Trust evaluators have changed in 5.0 and while migrating it is difficult to discover the evaluator that previously was used as `.disableEvaluation` due to semantic inconsistency and typos in the [migration guide](https://github.com/alamofire/Alamofire/blame/master/Documentation/Alamofire%205.0%20Migration%20Guide.md#L36)

Nothing new, I renamed `DisabledEvaluator` to `DisabledTrustEvaluator` and updated documentation and migration guide accordingly

### DIFF
--- a/Source/ServerTrustEvaluation.swift
+++ b/Source/ServerTrustEvaluation.swift
@@ -335,7 +335,10 @@ public final class CompositeTrustEvaluator: ServerTrustEvaluating {
 /// Disables all evaluation which in turn will always consider any server trust as valid.
 ///
 /// **THIS EVALUATOR SHOULD NEVER BE USED IN PRODUCTION!**
-public final class DisabledEvaluator: ServerTrustEvaluating {
+@available(*, deprecated, message: "DisabledEvaluator will be renamed, please use DisabledTrustEvaluator instead")
+public typealias DisabledEvaluator = DisabledTrustEvaluator
+
+public final class DisabledTrustEvaluator: ServerTrustEvaluating {
     /// Creates an instance.
     public init() {}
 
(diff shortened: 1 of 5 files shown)


jshier:
You have a huge number of changes in this PR. Please limit them to just the rename for review.

MrCloud:
@jshier I removed the changes in the `docs/` since they would be generated.

jshier:
I think this change makes sense, but since we shipped `DisabledEvaluator` as part of Alamofire 5, renaming it is a breaking change. Can you implement a deprecated `typealias` for the old name? That way users can automatically rename when they update.

MrCloud:
@jshier Sure !

Source/ServerTrustEvaluation.swift
```diff
@@ -335,7 +335,10 @@ public final class CompositeTrustEvaluator: ServerTrustEvaluating {
 /// Disables all evaluation which in turn will always consider any server trust as valid.
 ///
 /// **THIS EVALUATOR SHOULD NEVER BE USED IN PRODUCTION!**
-public final class DisabledEvaluator: ServerTrustEvaluating {
+@available(*, deprecated, message: "DisabledEvaluator will be renamed, please use DisabledTrustEvaluator instead")
```
jshier, line 338: This should also include the `renamed` parameter, and should be rephrased.

```suggestion
@available(*, deprecated, renamed: "DisabledTrustEvaluator", message: "DisabledEvaluator has been renamed DisabledTrustEvaluator.")
```

Source/ServerTrustEvaluation.swift
```diff
@@ -335,7 +335,10 @@ public final class CompositeTrustEvaluator: ServerTrustEvaluating {
...
+public typealias DisabledEvaluator = DisabledTrustEvaluator
+
+public final class DisabledTrustEvaluator: ServerTrustEvaluating {
```
jshier, line 341: Please copy the previous doc comment to this type, in addition to the `typealias`.

Issue #3166, opened 2020-04-28
The anchor certificates for the self-singed certificates


liuwin7:
## What did you do?

In the file `ServerTrustEvaluation.swift`,
``` swift
/// Sets a custom certificate chain on `self`, allowing full validation of a self-signed certificate and its chain.
///
/// - Parameter certificates: The `SecCertificate`s to add to the chain.
/// - Throws:                 Any error produced when applying the new certificate chain.
func setAnchorCertificates(_ certificates: [SecCertificate]) throws {
    // Add additional anchor certificates.
    let status = SecTrustSetAnchorCertificates(type, certificates as CFArray)
    guard status.af.isSuccess else {
        throw AFError.serverTrustEvaluationFailed(reason: .settingAnchorCertificatesFailed(status: status,
                                                                                           certificates: certificates))
    }

    // Reenable system anchor certificates.
    let systemStatus = SecTrustSetAnchorCertificatesOnly(type, true)
    guard systemStatus.af.isSuccess else {
        throw AFError.serverTrustEvaluationFailed(reason: .settingAnchorCertificatesFailed(status: systemStatus,
                                                                                           certificates: certificates))
    }
}
```

if you want to reenable system anchor certificates, you should call the func with `false` for the second parameter, instead of the `true` as the [apple's document](https://developer.apple.com/documentation/security/certificate_key_and_trust_services/trust/configuring_a_trust) says.

```swift
SecTrustSetAnchorCertificatesOnly(type, false)
```

jshier:
This method is correct, as we only want to trust the certificate chains for the certificates provided. The comment may be incorrect, but this is the behavior we want. This code is carefully designed and tested and any change in behavior must be thoroughly considered and well tested. Is there a specific use case you have in mind here?

liuwin7:
No specific use case.
Just confused when reading the code.
Please correct the comment if possible. :)

jshier:
Sure. We'll update when we do another documentation pass. A PR is also welcome!

kirin521:
i'm confused when reading the code. Is the purpose of this function to trust only provided certificates? How can I also trust the system's default certificate?

jshier:
Yes, it's used to trust the provided certificate when pinning a self-signed cert. There is no system default leaf certificate. And if you want to just use the system trust settings there's no need to pin anything at all, or call any of this code.

PR #3167, opened 2020-04-28
Reenable the system certificates.

liuwin7:
### Associated Issue 
#3166

### Goals :soccer:
Reenable the system certificates, when use the self-signed certificates.

```swift
// Reenable system anchor certificates.
let systemStatus = SecTrustSetAnchorCertificatesOnly(type, true)
```
replaced with
```swift
// Reenable system anchor certificates.
let systemStatus = SecTrustSetAnchorCertificatesOnly(type, false)
```

### Testing Details :mag:
No testing. However, here is the apple's doc.
> If true, disables trusting any anchors other than the ones passed in with the `SecTrustSetAnchorCertificates(_:_:)` function.  If false, the built-in anchor certificates are also trusted. If `SecTrustSetAnchorCertificates(_:_:)` is called and `SecTrustSetAnchorCertificatesOnly(_:_:)` is not called, only the anchors explicitly passed in are trusted.


jshier:
No changes can be accepted to our certificate handling code without extremely strong justification and robust tests. Please better justify this change and add appropriate tests illustrating the motivating issues.

liuwin7:
Ok, it is just an error for an inappropriate comment.

PR #3170, opened 2020-04-30
Expandability

grand-china:


jshier:
Sorry, we're not in a position to support overrides of the internal API at this time. If you create a topic at our [Swift forums](https://forums.swift.org/c/related-projects/alamofire/37) summarizing what you're trying to do we may be able to help you accomplish your goals without needing to subclass.

PR #3180, opened 2020-05-07
Fix memory leak in NetworkReachabilityManager.

dirtmelon:
### Goals :soccer:
- Fix a memory leak in `NetworkReachabilityManager`

As [passRetained(_:)](https://developer.apple.com/documentation/swift/unmanaged/1541288-passretained) described, it creates an unmanaged reference with an unbalanced retain.The instance passed as value will leak if nothing eventually balances the retain.
In this case, we don't need to perform a retain, if the `NetworkReachabilityManager` deinit, we will call `stopListening` to set the context of callback nil.

### DIFF
--- a/Source/NetworkReachabilityManager.swift
+++ b/Source/NetworkReachabilityManager.swift
@@ -174,7 +174,7 @@ open class NetworkReachabilityManager {
         }
 
         var context = SCNetworkReachabilityContext(version: 0,
-                                                   info: Unmanaged.passRetained(self).toOpaque(),
+                                                   info: Unmanaged.passUnretained(self).toOpaque(),
                                                    retain: nil,
                                                    release: nil,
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! Your're right, this is likely a bug. Would it be possible to demonstrate the cycle with a test? Say, create an optional manager, capture it weakly, start and stop it, then nil the reference and check the weak reference to see if it's still around?

dirtmelon:
@jshier Done. Thanks for review :)

jshier:
@dirtmelon Looks like the added tests has an up to 10% failure rate, so there may be something else keeping the reference alive or internal async processes we need to wait for. I'm investigating, but I'd appreciate your attention too.

jshier:
Looks good, thanks!

PR #3181, opened 2020-05-07
adding more flexibility in reusing RetryPolicy

Skoti:
### Goals :soccer:
Hi

I would like to introduce more flexibility in reusing `RetryPolicy` behaviour.

It is very cool that `RetryPolicy` already implements a lot of default retry behaviour based on http methods, status codes and url codes, however, this behaviour is not easily available for reuse.

What I mean is that if you override `retry` function from `RetryPolicy` you are either forced to ignore the inherited behavior or reuse whole of it as is.

For instance you cannot keep the default retry behavior for http statuses and url codes and retry based on a maximum elapsed time (i.e. you keep retrying for 5 minutes) instead of a retry limit but drop the http method check, and if so still retry with exponential delay. You can also imagine any other combination that would be useful or specific in your case.

In other words you cannot reuse useful parts of this retry behaviour (especially private `shouldRetry` function) which is a boilerplate code you would be forced to copy into your codebase for reuse.

So `RetryPolicy` doesn't give you any benefit for partial customization it would be like overriding `retry` function from plain `RequestInterceptor`.
You end up copying what `shouldRetry` function already does because it is not publicly available and because it is an instance method of `RetryPolicy`.

Additional concern:
I am also not sure if the `ConnectionLostRetryPolicy` is a good example of why one would need to inherit after `RetryPolicy`. What it does is just a customized init call. I guess a convenience initializer or a static factory method would be just fine in that case.
In addition, it is also a bit misleading because one might think that the `.networkConnectionLost` url error code is not handled by default `RetryPolicy` and requires a special, different class with probably a different behaviour to be used - which is not true as it is already defined in the `defaultRetryableURLErrorCodes` and used by the base `RetryPolicy`.

This is purely an additive PR, it does not breaks any API or behaviour.

This PR improves reusability by exposing the `shouldRetry` capabilities in several different public static methods.

It also adds `ConditionalRetrier` and `ConditionalRetryPolicy` that easily allow you to separate two parts that make up the whole retry behaviour which are:
- when to retry
- with what delay

By using `ConditionalRetryPolicy` you can easily decide whether you want to customize the default when-to-retry behavior and/or exponential delay of `RetryPolicy`.
By using static `shouldRetry` functions you can easily further break down the customization of when-to-retry behaviour.

It is also very easy now to expand the set of different should-retry behaviour and/or delay-providing in the future.

### Testing Details :mag:
All new methods and classes are tested.


jshier:
Thanks for the PR, it looks very useful! I'll give it a full review over the weekend, but in the mean time could you run SwiftFormat in the Alamofire directory to match our styling? If you have SwiftFormat installed, just run `swiftformat .` in the Alamofire directory.

Skoti:
I run it, although I don't like the results but it is your preference 😃. However, the init in `ConditionalRetryPolicy` is a total mess now.

jshier:
Thanks again for the PR. I agree that `RetryPolicy` should be more useful if the retry logic was more composable. However, I think it would be simpler, and even more flexible, to make the composable methods instance methods rather than static methods, so they can both be composed and overridden themselves. 

Additionally, can you remove the `ConditionalRetryPolicy` from this PR? We'd prefer if refactor and addition PRs are separate, so feel free to move it to a separate PR. However, it does need a strong justification, as it doesn't seem to bring much on its down.

Skoti:
I'm not really sure how making these methods instance methods would increase flexibility? Could you explain or give an example?
Their logic does not depend on any state in `RetryPolicy`, their behaviour can be fully controlled by their parameters and they do not produce any side effects.
What could you possibly gain by making them overridable? Even if I made them `class` methods instead of `static` (to make them overridable but still not instance methods).
In fact not all of these should-retry methods are used in the current `RetryPolicy` default behaviour, so it wouldn't even make sense to override them at all as not all of them are called.

I've added `ConditionalRetryPolicy` to not make breaking changes to `RetryPolicy`. I wanted to make a type that is not tied up to only one way of retrying or providing a delay like the `RetryPolicy` currently is.
Even if you override one of the shouldRetry methods (still not sure what for) the pattern of putting them together in `retry(_:for:dueTo:completion:)` is still tied up in a logical conjunction with the retryCount.

For me it is not the `ConditionalRetryPolicy` that is not justified but the current `RetryPolicy` stiffness. Refactoring `RetryPolicy` would end up just like the `ConditionalRetryPolicy` looks now as it is the most flexible, and you can easily customize the retry behavior (separately) without inheritance by just using different static should-retry or delay-providing methods.
How have you imagined refactoring it?

Customizing the retry behaviour through `ConditionalRetryPolicy` is as easy as calling its constructor with a different logic in one of the closures - how is that less simple and useful than creating a whole type to inherit after `RetryPolicy` and overriding some specific methods?

Mind that one needs to understand how the particular base type works in order to override it with the confidence all will work as intended. Whereas in case of `ConditionalRetryPolicy` you only need to know the particular methods you're gonna use in its constructor closures.

jshier:
We discussed this further and come to a few conclusions:

1. Breaking every bit of logic in `RetryPolicy` into their own methods greatly increases the internal complexity for little gain. Letting `shouldRetry` be overridable would give users most of the same flexibility without the extra complexity. It really shouldn't be necessary to create functions that just check `contains` or compare values.
2. The `open` methods should be instance methods so that they can operate on the initialized state of the `RetryPolicy` instead of those values being passed around. Making them `static` or even `class` methods is both more complex and doesn't get us anything as no one needs to use those methods outside of instances of the type.
3. `ConditionalRetryPolicy` is interesting, but is basically the same as implementing the protocols themselves, as you'd essentially have to implement the same logic in the closures as you would in the protocol methods. We think it needs to be justified separately from the other modifications in the PR.

So, to move forward, we'd like to see this PR simplified down to an overridable `shouldRetry(request:dueTo:)` instance method and we'll evaluate again from there.

Skoti:
Does the above commit is what you've aimed for?

Regarding `ConditionalRetryPolicy` and `ConditionalRetrier`:
If you really don't like the idea of using closures to implement a protocol then why you even have this:
https://github.com/Alamofire/Alamofire/blob/d976690ea9d55a799c8dde418131eb95710e3ca9/Source/RequestInterceptor.swift#L115-L116
and this:
https://github.com/Alamofire/Alamofire/blob/d976690ea9d55a799c8dde418131eb95710e3ca9/Source/RequestInterceptor.swift#L133-L134
in the library?

Probably because it is more convenient to write `let retrier = Retrier { //some logic }` instead of creating a whole type to implement the same - which is more convenient if it is short and self-contained logic, and in such simple and short-line cases it is even more justified to have it.

In fact, `ConditionalRetryPolicy` and `ConditionalRetrier` are not just simple closure invocations inside an implemented protocol method, but they actually provide a convenience flow (see `ConditionalRetrier`) and change the concept of returning `RetryResult` to returning simple values that are only used in particular parts of the flow (`Bool` for _shouldRetry_ and `TimeInterval?` for _delayForRetrying_).
Thus making it easier to **combine and reuse** different parts of the required retry logic.
So it is not like:
> is basically the same as implementing the protocols themselves, as you'd essentially have to implement the same logic in the closures as you would in the protocol methods.

Which is the perfect description of what you alredy have in `Adapter` and `Retrier`.

I've thought that ideally `RetryPolicy` would inherit after `ConditionalRetrier` to provide a convenience default retry policy behaviour, and current `shouldRetry` and `delayForRetrying` functions would provide further customization points.
However, this would introduce breaking changes and this is why, in the first place, I've come up with dividing `RetryPolicy` logic to static functions that could easily be reused in the added `ConditionalRetryPolicy`.

jshier:
Right. We already have `Adapter` and `Retrier` to make simple cases trivial, and they can be arbitrarily composed, along with any number of other adapters and retriers, using `Interceptor`. Adding another type that's only slightly different than what Alamofire already offers isn't sustainable. Also, since `ConditionalRetryPolicy` requires users to supply their own condition and timing logic anyway, it's no different than implementing the protocols directly, which is what I meant before.

In any case, I've asked you to remove `ConditionalRetryPolicy` from this PR. It will not be merged with it. 

Long term I agree there's more that can be done for retriers, but that would include things like backoff strategies that can be applied to all retriers or other reusable functionality.

Source/RetryPolicy.swift
```diff
@@ -309,31 +309,49 @@ open class RetryPolicy: RequestInterceptor {
...
         }
     }
+
+    // MARK: - DelayForRetrying
+
+    /// - Parameters:
+    ///     - request: `Request` that failed due to the provided `Error`.
+    ///     - error:    `Error` encountered while executing the `Request`.
+    /// - Returns: A  delay in seconds or nil to retry without a delay.
+    open func delayForRetrying(request: Request, dueTo error: Error) -> TimeInterval? {
```
jshier, line 351: It doesn't make sense to allow overriding of the delay calculation, as `RetryPolicy` was designed around geometric backoff and doesn't really make sense without it.

Skoti:
> In any case, I've asked you to remove ConditionalRetryPolicy from this PR. It will not be merged with it.

and I did. I thought `ConditionalRetrier` could stay. Anyway, it is removed as well in the commit above.

---
I am sad to see this PR cut down to a simple `shouldRetry` override, and not adding any other means (in any form) to reuse and/or switch between delay strategies; as well as composing different should-retry and delay strategies easily.
I hope this PR will be an inspiration to change this in the near future. 🤞 

> Long term I agree there's more that can be done for retriers, but that would include things like backoff strategies that can be applied to all retriers or other reusable functionality.

I am willing to help if possible. 🧑‍🏭

PR #3195, opened 2020-05-20
Removes customization of URLError localized error messages

### LINKED ISSUE #3194: Alamofire overrides system provided localized error messages
* [X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

When a `URLSessionTask` fails with an error, localization is broken because Alamofire appends a hard coded English string to the localized error.

So for example when a `URLError` reports no internet connectivity available, if we rely on the value of `localizedString` to display an error message to the user we will end up with the following message:

```swift
URLSessionTask failed with error: [[LOCALIZED VERSION OF THE ERROR MESSAGE HERE THAT MAY BE SOMETHING OTHER THAN ENGLISH]]
```

Modifying the localized error message provided by the underlying error serves no purpose and forces library consumers to localize their own messages unnecessarily, or reach out for the underlying error message which can get very complex pretty quickly.

Alamofire should **not** modify the localized error provided by the system.

## What happened instead?

Alamofire appends a hard coded English string to the localized message

## Alamofire Environment

**Alamofire version:** 5.x
**Xcode version:** 11.5
**Swift version:** 5.2
**Platform(s) running Alamofire:** iOS, tvOS
**macOS version running Xcode:** 10.15.x

rpassis:
### Issue Link :link:

Fixes #3194

### Goals :soccer:

Remove localized error message overriding by Alamofire when a URLSessionTask error occurs

Opts to remove the custom value from `localizedString` and relay the underlying localized error message instead.

### DIFF
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<Workspace
-   version = "1.0">
-   <FileRef
-      location = "self:">
-   </FileRef>
-</Workspace>
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -673,7 +673,7 @@ extension AFError: LocalizedError {
         case let .downloadedFileMoveFailed(error, source, destination):
             return "Moving downloaded file from: \(source) to: \(destination) failed with error: \(error.localizedDescription)"
         case let .sessionTaskFailed(error):
-            return "URLSessionTask failed with error: \(error.localizedDescription)"
+            return error.localizedDescription
         }
     }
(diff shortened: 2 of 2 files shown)


jshier:
Thanks for the PR!

I'm a bit torn on this. Generally, I don't think it's a good idea to pass underlying error descriptions directly to the user, especially these sorts of low level networking errors. User's can't really do much about them and they're usually more technical than most users understand anyway. Additionally, this can easily be worked around by extracting the underlying error just like you are in the PR itself. So adding a special case for this error case when Alamofire's errors aren't generally supposed to be shown to the user may not make sense.

@cnoon, any thoughts?

@rpassis Is there any particular reason why, when Alamofire's errors aren't otherwise localized, you'd want this one case to be? If you're special casing it's handling anyway, reaching into the underlying error to get the same value would work fine.

jshier:
> The question I ended up asking is how is this beneficial to any consumers of the API?

I think this is the crux of the issue. My only argument in favor of the current message is that makes it more clear the error came from the underlying `URLSession` and not Alamofire directly.

Ultimately I think it comes down to the use case. Currently, to access the `underlyingError`'s `localizedDescription`, users can use the `.underlyingError` property or switch over the `AFError` instance and capture the instance directly. They'd have to do this if they want to avoid Alamofire's other `localizedDescription`s. With this change, the old approaches would still work, and they'd still have to switch over the `AFError` to avoid the other descriptions, so I'm not sure what it ultimately gets us.

Given that this doesn't actually seem to improve the ergonomics of presenting these errors to the user, as users still have to avoid Alamofire's other non-localized descriptions, I'm going to decline this PR for now. Thanks for the work!

PR #3200, opened 2020-05-22
On Hold: Add withAllRequests API

### LINKED ISSUE #3190: Extend cancelAllRequests method
## What did you do?

Updated to latest version 5.2.0

I want to be able to cancel *some* (not all) previously started requests.

## What happened instead?

I've noted this newly introduced note: `This instance should **NOT** be used to interact with the underlying URLSessionTasks. Doing so will break internal Alamofire logic that tracks those tasks.`
Previously I used the way that is described here to cancel requests: https://github.com/Alamofire/Alamofire/issues/1703
This doesn't seem to be the way to go any longer. I've noticed `cancelAllRequests` method. So my suggestion would be to either make `activeRequests` public or to introduce a filter option that's used in `cancelAllRequests` method, like this:
```
public func cancelAllRequests(completingOnQueue queue: DispatchQueue = .main, isIncluded: ((Request) -> Bool)? = nil, completion: (() -> Void)? = nil) {
    rootQueue.async {
        var requests = self.activeRequests
        if let isIncluded = isIncluded {
            requests = requests.filter(isIncluded)
        }
        requests.forEach { $0.cancel() }
        queue.async { completion?() }
    }
}
```

## Alamofire Environment

**Alamofire version: 5.2.0**
**Xcode version: 11.4.1**
**Platform(s) running Alamofire: iOS**

jshier:
### Issue Link :link:
Fixes #3190

### Goals :soccer:
This PR adds a `withAllRequests` method that allows users to perform actions on every `Request` in `activeRequest`s.

`withAllRequests` is essentially the same thing as `cancelAllRequests` but takes a closure to perform the action. Additionally, `cancelAllRequests` has been reimplemented using `withAllRequests`.

### Testing Details :mag:
An additional test was added for `withAllRequests`.

### DIFF
--- a/Documentation/AdvancedUsage.md
+++ b/Documentation/AdvancedUsage.md
@@ -202,6 +202,27 @@ monitor.requestDidCompleteTaskWithError = { (request, task, error) in
 let session = Session(eventMonitors: [monitor])
 ```
 
+### Operating on All Requests
+Although use should be rare, `Session` provides the `withAllRequests` method to operate on all currently active `Request`s. This work is performed on the `Session`'s `rootQueue`, so it's important to keep it quick. If the work may take some time, creating a separate queue to process the `Set` of `Request`s should be used.
+
+```swift
+let session = ... // Some Session.
+session.withAllRequests { requests in 
+    requests.forEach { $0.suspend() }
+}
+```
+
+Additionally, `Session` offers a convenience method to cancel all `Request`s and call a completion handler when complete.
+
+```swift
+let session = ... // Some Session.
+session.cancelAllRequests(completingOn: .main) { // completingOn uses .main by default.
+    print("Cancelled all requests.")
+}
+```
+
+> Note: These actions are performed asynchronously, so requests may be created or have finished by the time it's actually run, so it should not be assumed the action will be performed on a particular set of `Request`s.
+
 ### Creating Instances From `URLSession`s
 In addition to the `convenience` initializer mentioned previously, `Session`s can be initialized directly from `URLSession`s. However, there are several requirements to keep in mind when using this initializer, so using the convenience initializer is recommended. These include:
 
--- a/Source/Session.swift
+++ b/Source/Session.swift
@@ -204,7 +204,22 @@ open class Session {
         session.invalidateAndCancel()
     }
 
-    // MARK: - Cancellation
+    // MARK: - All Requests API
+
+    /// Perform an action on all active `Request`s.
+    ///
+    /// - Note: The provided `action` closure is performed asynchronously, meaning that some `Request`s may complete and
+    ///         be unavailable by time it runs. Additionally, this action is performed on the instances's `rootQueue`,
+    ///         so care should be taken that actions are fast. Once the work on the `Request`s is complete, any
+    ///         additional work should be performed on another queue.
+    ///
+    /// - Parameters:
+    ///   - action:     Closure to perform with all `Request`s.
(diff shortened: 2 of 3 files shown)


tfonfara:
Looks good to me 👍

Documentation/AdvancedUsage.md
```diff
@@ -202,6 +202,25 @@ monitor.requestDidCompleteTaskWithError = { (request, task, error) in
...
+Additionally, `Session` offers a convenience method to cancel all `Request`s.
+
+```swift
+let session = ... // Some Session.
+session.cancelAllRequests(completingOn: .main) {
+    print("Cancelled all requests.")
+}
+```
+
+> Note: These actions are performed asynchronously, so requests may be created or have finished by the time it's actually run, so it should be assumed the action will be performed on a particular set of `Request`s.
```
jshier, line 222:
```suggestion
> Note: These actions are performed asynchronously, so requests may be created or have finished by the time it's actually run, so it should not be assumed the action will be performed on a particular set of `Request`s.
```

Source/Session.swift
```diff
@@ -204,7 +204,29 @@ open class Session {
...
+    ///         be unavailable by time it runs. Additionally, these actions are performed on the instances's `rootQueue`,
+    ///         so care should be taken that actions are fast.
+    ///
+    /// - Parameters:
+    ///   - queue:      `DispatchQueue` on which the completion handler is called. `.main` by default.
+    ///   - action:     Closure to perform on each `Request`.
+    ///   - completion: Closure to be called when `action` has been called on all `Request`s. `nil` by default.
+    public func withAllRequests(completingOn queue: DispatchQueue = .main,
+                                perform action: @escaping (Request) -> Void,
+                                completion: (() -> Void)? = nil) {
```
cnoon, line 221: This API doesn't really match our other APIs conventions with `completion` closures. Would the following make more sense?

```swift
public func withAllRequests(
    perform action: @escaping (Request) -> Void,
    queue: DispatchQueue = .main,
    completion: (() -> Void)? = nil)
```

This would also put the parameters with default values at the end of the signature. Did you pivot from the convention since there's another closure in the API?

jshier: Yes, there were two reasons.

- Putting the closure parameters together seems more natural since it's likely they'll be the parameters used most often.
- And, in the wake of the acceptance of [SE-0279](https://github.com/apple/swift-evolution/blob/master/proposals/0279-multiple-trailing-closures.md), there is a new multiple trailing closure syntax. This API is designed to be compatible with that spelling.

jshier: It looks like this will all parameters:

```swift
session.withAllRequests(completingOn: .main) { (request) in
    // Do something with all requests.
} completion: {
    // Do something when complete.
}
```

jshier:
I'm putting this PR on hold due to ongoing discussion around multiple closure syntax, which this API was designed around, and bugs in the current autocomplete experience ([SR-12880](https://bugs.swift.org/browse/SR-12880)) which should both be resolved fairly soon.

jshier:
The unfortunate fact of the matter is that there doesn't seem to be a good solution for usability coming any time soon. In order to fix the conflict between wanting the single closure using the action and the multiple closure's version having the action in the first closure position, we need to overload the API to drop the completion closure in one version. 😢 

I'll make the update and we can ship this API.

jshier:
On second thought, if we give the collection of `Request`s users can call their own completion.

PR #3206, opened 2020-05-24
adding `HTTPMethodDependingParameterEncoder`

Skoti:
Following this approach, I think that the most common usage of encoders is to use `JSONParameterEncoder` when one needs to send data in a body (typically in a _post_ method) and a `URLEncodedFormParameterEncoder` when one needs to use a query string (typically in a _get_ method).
As indicated this usually depends on an http method.

This is why I would like to add `HTTPMethodDependingParameterEncoder` which uses different parameter encoders for different http methods.

By default it is configured to use `URLEncodedFormParameterEncoder.default` for _get_, _head_ and _delete_ methods, and `JSONParameterEncoder.default` for all other http methods.

### Testing Details :mag:
Basic test for `HTTPMethodDependingParameterEncoder`. I will add more if requested and the PR is initially approved.

### DIFF
--- a/Source/ParameterEncoder.swift
+++ b/Source/ParameterEncoder.swift
@@ -38,6 +38,45 @@ public protocol ParameterEncoder {
     func encode<Parameters: Encodable>(_ parameters: Parameters?, into request: URLRequest) throws -> URLRequest
 }
 
+/// A `ParameterEncoder` that uses different parameter encoders for different http methods.
+///
+/// `Content-Type` header is set accordingly by the provided encoders.
+open class HTTPMethodDependingParameterEncoder: ParameterEncoder {
+    
+    /// - Returns: an instance using `init(getHeadDeleteMethodsParameterEncoder:defaultParameterEncoder)` with default parameters.
+    public static var `default`: HTTPMethodDependingParameterEncoder { .init() }
+
+    /// A closure that returns the desired `ParameterEncoder` for a given `HTTPMethod`.
+    public let parameterEncoderFor: (HTTPMethod) -> ParameterEncoder
+    
+    /// Creates an instance with the provided escaping closure that returns the desired `ParameterEncoder` for a given `HTTPMethod`.
+    public init(parameterEncoderFor: @escaping (HTTPMethod) -> ParameterEncoder) {
+        self.parameterEncoderFor = parameterEncoderFor
+    }
+    
+    /// Creates an instance with the provided parameter encoders.
+    ///
+    /// - Parameter getHeadDeleteMethodsParameterEncoder: `ParameterEncoder` for `.get`, `.head`, and `.delete` requests. `URLEncodedFormParameterEncoder.default` by default.
+    /// - Parameter defaultParameterEncoder:              `ParameterEncoder`  for all requests other than `.get`, `.head`, and `.delete`.  `JSONParameterEncoder.default` by default.
+    public convenience init(getHeadDeleteMethodsParameterEncoder: ParameterEncoder = URLEncodedFormParameterEncoder.default, defaultParameterEncoder: ParameterEncoder = JSONParameterEncoder.default) {
+        self.init { httpMethod in
+            switch httpMethod {
+            case .get, .head, .delete:
+                return getHeadDeleteMethodsParameterEncoder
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! However, while I like the idea of this PR, the implementation doesn't seem to go far enough to give users real flexibility in composing `ParameterEncoder`s. For example, you've limited it to two encoders: one for `.get`, `.head`, and `.delete` methods, one one for all the rest. In real apps there may be differences in encodings for *every request*! How would this approach help there? 

Additionally, we usually try to make sure the types included with Alamofire have at least one of two things: they make something difficult easy, or they're far more convenient than implementing something manually. Unfortunately this type doesn't do that right now. Demonstrating a more flexible approach and how it could help users would go a long way towards proving this type, or something like it would be worth inclusion in Alamofire.

Finally, the approach we usually recommend for composition like this is an application of the Router pattern or other abstraction using our `URLRequestConvertible` protocol. That way every route for every request can use their own encoder and decoder, if necessary, with any number of other parameters. In addition to justifying the inclusion of this type, it would great to see how composing `ParamterEncoder`s compares to using them directly in a Router.

Skoti:
> In addition to justifying the inclusion of this type, it would great to see how composing ParamterEncoders compares to using them directly in a Router.

As for the Router design pattern indicated here:
https://github.com/Alamofire/Alamofire/blob/master/Documentation/AdvancedUsage.md#routing-requests
Unfortunately, there is no gain as you still need to get the concrete parameters, so you can already use that code to pass the desired encoder as well.

> In real apps there may be differences in encodings for every request! How would this approach help there?

It wouldn't really. In case a particular request needs a very specific encoder it's better to use that specific encoder directly instead of the one that chooses the same encoder based on some logic (similarly as I mentioned for the router pattern above).

---
This PR approach is more suited for a global/default parameter encoder, like here with `URLEncodedFormParameterEncoder.default`:
https://github.com/Alamofire/Alamofire/blob/2777659076fda38bd4e487739ee331c7a65b5924/Source/Session.swift#L307-L313

> Demonstrating a more flexible approach and how it could help users would go a long way towards proving this type, or something like it would be worth inclusion in Alamofire.

I guess it is more convenient to have the most common usage covered by Alamofire out-of-the-box, especially if one doesn't follow the router pattern. 

Of course to achieve that one could also just do this:
```swift
class ExampleSession: Session {
    
    override func request<Parameters: Encodable>(
        _ convertible: URLConvertible,
        method: HTTPMethod = .get,
        encoder: ParameterEncoder? = nil,
        headers: HTTPHeaders? = nil,
        interceptor: RequestInterceptor? = nil,
        requestModifier: RequestModifier? = nil
    ) -> DataRequest {
        let parameterEncoder: ParameterEncoder
        if let encoder = encoder {
            parameterEncoder = encoder
        } else {
            switch method {
            case .get, .head, .delete:
                parameterEncoder = URLEncodedFormParameterEncoder.default
            default:
                parameterEncoder = JSONParameterEncoder.default
            }
        }
        return super.request(convertible, method: method, parameters: parameters, encoder: parameterEncoder, headers: headers, interceptor: interceptor, requestModifier: requestModifier)
    }
}
```
which is basically what `HTTPMethodDependingParameterEncoder` does.

If desired this PR can be altered to implement this in the `Session`, but it would be a breaking change. Although it would be nice to not forget about it in the next Alamofire major version. 🙏 

Anyway, I wanted to make Alamofire more friendly in out-of-the-box usage, especially if you're not following the router pattern as it is not the only justified way of building requests.
Also the convenience `Session` api that is not using `URLRequestConvertible` would be more useful. And as it's probably the most common division to use json for post and url encoding for get request I thought it would be nice to have it in Alamofire.

Although, I understand if this carries too little value for you to include `HTTPMethodDependingParameterEncoder` in Alamofire.

jshier:
Sorry it took so long to get back to this. While we continue to think about more convenient ways to compose parameter encoders, we don't think this approach provides the solution we're looking for. Alamofire is due for reevaluation of how parameter encoding works to support things like composing encoders for queries and bodies simultaneously, or some sort of path encoder. When that work starts we'll keep this sort of composition in mind as well.

PR #3211, opened 2020-05-26
Encode array of objects correctly

MartinPham:
### Goals :soccer:
From an array of objects:
```
"test": [{
   "key1": "val1",
   "key2": "val2"
},{
   "key3": "val3",
   "key4": "val4"
}]
```

It will be transformed to
```
test[0][key1]=val1
&test[0][key2]=val2
&test[1][key3]=val3
&test[1][key4]=val4
```

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -94,6 +94,15 @@ public struct URLEncoding: ParameterEncoding {
                 return key
             }
         }
+        
+        func encode(key: String, index: Int) -> String {
+            switch self {
+            case .brackets:
+                return "\(key)[\(index)]"
+            case .noBrackets:
+                return key
+            }
+        }
     }
 
     /// Configures how `Bool` parameters are encoded.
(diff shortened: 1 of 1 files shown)


MartinPham:
Currently it's transformed into
```
test[][key1]=val1
&test[][key2]=val2
&test[][key3]=val3
&test[][key4]=val4
```

which is understood as
```
"test": [{
   "key1": "val1"
},
{
   "key2": "val2"
},
{
   "key3": "val3"
},
{
   "key4": "val4"
}]
```

jshier:
Thanks for the PR! Given that there is no "correct" for this type of encoding, Alamofire needs to support multiple encodings. That means this encoding should be support alongside Alamofire's currently supported encodings, and that it would need to support `URLEncodedFormParameterEncoder`. This is what #3015 was doing, so if you want to integrate those changes and add specific tests for that case that don't replace existing tests, we might be able to get it merged this time.

jshier:
Given no forward progress on this PR, I'm closing it. If anyone would like the functionality from this PR, please address the comments and open a new PR.

PR #3214, opened 2020-05-29
Use switch to replace if else expression

dirtmelon:
### Goals :soccer:

Makes `queryComponents(fromKey:value:)` more readable.

Use switch to replace if else expression, we can use switch to attempt casting into different types.

### Testing Details :mag:
No new tests.


jshier:
Thanks for the PR! While the improvement is slight, I think it matches some other switches we have.

Please run the latest version of SwiftFormat in the Alamofire directory (`swiftformat .`).

dirtmelon:
@jshier Thanks review.I have updated the format.

PR #3216, opened 2020-05-31
DataStreamRequest and test reliability fixes

jshier:
### Goals :soccer:
This PR fixes reliability issues in `DataStreamRequest` found while performing stress tests. Essentially, due to the extra queue hopping performed during `Data` streaming vs. the completion event, completion events could be received by the handler closure before the stream event. 

Additionally, various changes were made to certain tests to make them more reliable.

`DataStreamRequest` has been made more reliable by the addition of a `numberOfExecutingStreams` property to its mutable state. This value is incremented every time the stream handlers are enqueued and decremented when the handler closures are called. If the completion event is executed during this time, it's saved into the `enqueuedCompletionEvents` array to be performed when all streams are complete.

### Testing Details :mag:
`DataStreamRequest` tests have been updated to test the order of events in the stream handler and an additional case has been added to test behavior when a parallel serialization queue is used, as that case showed the highest failure rate.

Other tests have been updated for reliability.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -1154,6 +1154,8 @@ public final class DataStreamRequest: Request {
         /// Number of currently executing streams. Used to ensure completions are only fired after all streams are
         /// enqueued.
         var numberOfExecutingStreams = 0
+        /// Completion calls enqueued while streams are still executing.
+        var enqueuedCompletionEvents: [() -> Void] = []
     }
 
     @Protected
@@ -1286,7 +1288,17 @@ public final class DataStreamRequest: Request {
         appendResponseSerializer {
             self.underlyingQueue.async {
                 self.responseSerializerDidComplete {
-                    self.enqueueCompletion(on: queue, stream: stream)
+                    self.$streamMutableState.write { state in
+                        guard state.numberOfExecutingStreams == 0 else {
+                            state.enqueuedCompletionEvents.append {
+                                self.enqueueCompletion(on: queue, stream: stream)
+                            }
+
+                            return
+                        }
+
+                        self.enqueueCompletion(on: queue, stream: stream)
+                    }
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


jshier:
Took 3 tries with a *single* timeout failure, but the test suite should be far more reliable with these fixes.

Source/Request.swift
```diff
@@ -1282,17 +1286,29 @@ public final class DataStreamRequest: Request {
...
+                }
+            }
+        }
+    }
+
+    func enqueueCompletion<Success, Failure>(on queue: DispatchQueue,
+                                             stream: @escaping Handler<Success, Failure>) {
+        $streamMutableState.read { state in
+            guard state.numberOfExecutingStreams == 0 else {
+                underlyingQueue.async { self.enqueueCompletion(on: queue, stream: stream) }
```
cnoon, line 1299: Would you want a delay here at all? Seems like you're just re-enqueuing as fast as possible which may not be ideal.

Source/Request.swift
```diff
@@ -1206,15 +1209,16 @@ public final class DataStreamRequest: Request {
...
                 underlyingQueue.async {
                     var bytes = Array(data)
                     stream.write(&bytes, maxLength: bytes.count)
                 }
             }
-
-            underlyingQueue.async { state.streams.forEach { stream in stream.queue.async { stream.stream(data) } } }
+            state.numberOfExecutingStreams += state.streams.count
+            let localState = state
+            underlyingQueue.async { localState.streams.forEach { $0(data) } }
```
cnoon, line 1221: Is the `localState` copy here just to ensure all streams are called in the `async` dispatch if one of the streams were removed in a separate call since it's no longer in the lock?

Source/Request.swift
```diff
@@ -1206,15 +1209,16 @@ public final class DataStreamRequest: Request {
...
-        $streamMutableState.read { state in
+        $streamMutableState.write { state in
             if let stream = state.outputStream {
                 underlyingQueue.async {
                     var bytes = Array(data)
                     stream.write(&bytes, maxLength: bytes.count)
                 }
             }
-
-            underlyingQueue.async { state.streams.forEach { stream in stream.queue.async { stream.stream(data) } } }
```
cnoon, line 1217: Looks like this is the underlying race condition that we're dispatching the streams on the `underlyingQueue` but the response serializer can complete while this is still running. Is the the crux of it?

cnoon: If we didn't dispatch here, would we eliminate the race? Probably not at all the right solution, just trying to better understand the race itself as to why you landed on this solution.

Source/Request.swift
jshier: You're right, this will requeue rather quickly. The check is cheap, but I'm going to investigate a better solution.

Source/Request.swift
jshier: No, it's to work around a language limitation that doesn't allow `inout` values to escape their calling context.

Source/Request.swift
jshier: No, eliminating this dispatch doesn't fix the race. Ultimately, the race lives in the `serializationQueue`, hence the added test. In that test case, the extra work of dispatching onto a separate, parallel queue results in about a 10% failure rate (with some weird behavior), as the completion event avoids that work altogether and can finish much more quickly.

Source/Request.swift
jshier: Okay, I've implemented an alternate solution that stores the delayed completion until all streams have finished. Check it out in https://github.com/Alamofire/Alamofire/pull/3216/commits/9a72a35016ef164f9fcbcb20a8b436a6bb388914.

PR #3219, opened 2020-06-01
Lifetime Event APIs

### LINKED ISSUE #3213: Accessing request.task when task is created.
How is possible get the request.task serialized when available ?

sessionManager.upload(fileNameLocalPathUrl, to: url, method: .put, headers: headers, interceptor: nil, fileManager: .default)
        .request {  task in ....    <----- EXAMPLE
            taskHandler(task)
        }
        .uploadProgress { progress in
            progressHandler(progress)
            size = progress.totalUnitCount
        }
        .response { response in
            switch response.result {

thanks a lot

jshier:
### Issue Link :link:
Resolves #3213

### Goals :soccer:
This PR adds callback APIs for the `URLRequest` and `URLSessionTask`s created for `Request`s. This will allow integration with existing Apple API that may require those values, like an [`NSFileProvider`](https://developer.apple.com/documentation/fileprovider/nsfileprovidermanager/2890932-register). 

Similar to the existing cURL API, this PR adds stored closures which are called at the right moments in a `Request`'s lifetime.

### Testing Details :mag:
Tests have been added, more should be.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -92,8 +92,12 @@ public class Request {
         var redirectHandler: RedirectHandler?
         /// `CachedResponseHandler` provided to handle response caching.
         var cachedResponseHandler: CachedResponseHandler?
-        /// Closure called when the `Request` is able to create a cURL description of itself.
-        var cURLHandler: ((String) -> Void)?
+        /// Queue and closure called when the `Request` is able to create a cURL description of itself.
+        var cURLHandler: (queue: DispatchQueue, handler: (String) -> Void)?
+        /// Queue and closure called when the `Request` creates a `URLRequest`.
+        var urlRequestHandler: (queue: DispatchQueue, handler: (URLRequest) -> Void)?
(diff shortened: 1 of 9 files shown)


Source/Request.swift
```diff
@@ -816,17 +831,65 @@ public class Request {
...
     ///
-    /// - Parameter handler: Closure to be called when the cURL description is available.
+    /// - Parameters:
+    ///   - queue:   `DispatchQueue` on which `handler` will be called. `.main` by default.
+    ///   - handler: Closure to be called when the cURL description is available.
     ///
     /// - Returns:           The instance.
     @discardableResult
-    public func cURLDescription(calling handler: @escaping (String) -> Void) -> Self {
+    public func cURLDescription(on queue: DispatchQueue = .main, calling handler: @escaping (String) -> Void) -> Self {
```
jshier, line 840: As a note, this is not truly an API compatible change, so needs to change.

jshier: Added a separate override of `cURLDescription` to add this parameter while not breaking the previous version.

PR #3220, opened 2020-06-02
Remove @dynamicMemberLookup

dirtmelon:
### Goals :soccer:

Remove @dynamicMemberLookup of `Protected<T>`.

### Testing Details :mag:

Delete tests of @dynamicMemberLookup.

### DIFF
--- a/Source/Protected.swift
+++ b/Source/Protected.swift
@@ -109,7 +109,6 @@ final class UnfairLock: Lock {
 
 /// A thread-safe wrapper around a value.
 @propertyWrapper
-@dynamicMemberLookup
 final class Protected<T> {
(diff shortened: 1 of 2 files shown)


jshier:
Why would we remove this? It may not be used right now, but it's a very useful feature for the wrapper to have.

dirtmelon:
Yes, it not be used right now. The `Protected<T>` provides thread-safe wrapper by `@propertyWrapper`. Is that need to provide "dot" syntax for arbitrary names which are resolved at runtime with `@dynamicMemberLookup`?

jshier:
This is still desirable and so won't be removed. Thanks anyway! 🍻

PR #3242, opened 2020-06-26
fix:(ParameterEncoding.swift)reques URLEncoding bug

nehzx:
AF.request("https://httpbin.org/get?" parameters:["key":"value"]) 
expect "https://httpbin.org/get?key=value" 
result "https://httpbin.org/get?&key=value"

let percentEncodedQuery = (urlComponents.percentEncodedQuery.map { $0.isEmpty ? $0 : $0 + "&" } ?? "") + query(parameters)


jshier:
Thanks for the PR! Looks good, but it needs a test to verify the fix. If you can add one we can look into merging this PR.

nehzx:
> Thanks for the PR! Looks good, but it needs a test to verify the fix. If you can add one we can look into merging this PR.
截屏2020-06-27 上午3 04 29

Thanks for the reply!  The test failed due to an API error. (curl -X GET "https://httpbin.org/redirect-to" -H "accept: text/html")

jshier:
I'm aware. However, you still need to add tests for your specific change. We'll handle the redirect failures on httpbin separately.

PR #3243, opened 2020-06-27
fix: reques URLEncoding bug and add test case

nehzx:
let newQueryString = (components.percentEncodedQuery.map { $0.isEmpty ? $0 : $0 + "&" } ?? "") + query
let percentEncodedQuery = (urlComponents.percentEncodedQuery.map { $0.isEmpty ? $0 : $0 + "&" } ?? "") + query(parameters)

### Testing Details :mag:
add RequestURLTestCase

### DIFF
--- a/Source/ParameterEncoder.swift
+++ b/Source/ParameterEncoder.swift
@@ -162,7 +162,7 @@ open class URLEncodedFormParameterEncoder: ParameterEncoder {
             var components = URLComponents(url: url, resolvingAgainstBaseURL: false) {
             let query: String = try Result<String, Error> { try encoder.encode(parameters) }
                 .mapError { AFError.parameterEncoderFailed(reason: .encoderFailed(error: $0)) }.get()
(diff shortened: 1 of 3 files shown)


jshier:
In my own testing, these changes don't seem to be necessary. Can you provide your bug again?

Also, if you merge master you should pick up the test fixes.

nehzx:
> In my own testing, these changes don't seem to be necessary. Can you provide your bug again?
> 
> Also, if you merge master you should pick up the test fixes.


In get method if the URL path end'? 'or any other symbol, percentEncodedQuery will result in an "" instead of nil which will result in an extra & symbol in the URL concatenation process which is really not canonical

AF.request("https://httpbin.org/get?", parameters: ["key": "value"])
expectation:  https://httpbin.org/get?key=value
result:            https://httpbin.org/get?&key=value

Of course, If you don't think it is necessary, I will cancel the PR this time

PR #3264, opened 2020-07-23
Prevent manipulating request header fields directly

tomisacat:
### Goals :soccer:
Prevent manipulating request header fields directly.

Use HTTPHeaders methods to modify request `Content-Type` header field.

### Testing Details :mag:
No new tests added.


tomisacat:
It seems like a timeout error, any idea? @jshier

jshier:
Thanks, looks good!

PR #3267, opened 2020-08-04
update iOS Example

rain2540:


rain2540:
I've insteaded it with PR #3268, please ignore this PR

Issue #3271, opened 2020-08-06
Alamofire 5 - AFError should provide the original error


danielsaidi:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

Migrated from Alamofire 4.x to 5.2, then made failing network calls.

## What did you expect to happen?

I expected the `DataResponse` error to be the original network error.

## What happened instead?

The `DataResponse`error is an `AFError`

## More information

Our domain layer and app doesn't know anything about Alamofire - it's an implementation detail in our API-layer. We have `Error` extensions that tells us what kind of error an error is, e.g. `isOfflineError`. This worked before, when AF returned raw errors, but no longer works, since `AFError` abstracts this.

It would be great if `AFError` could provide an optional `original error`, so that we don't have to switch over the entire enum and do this manually. Or is there such a prop somewhere that we just have missed?

## Alamofire Environment

**Alamofire version:** 5.2
**Xcode version:** 11.5
**Swift version:** 5.2
**Platform(s) running Alamofire:** iOS

danielsaidi:
I completely missed the `underlyingError` property.

PR #3272, opened 2020-08-07
Add originalError to AFError and all its nested enums

danielsaidi:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/3271

### Goals :soccer:
To simplify accessing the original error of an `AFError` instance.

I have placed the implementation and unit tests in separate files to simplify review and merge. Feel free to move this if it's something that you want to include in the library.

### Testing Details :mag:
I have added tests for all enums. You can find them in `AFError+OriginalErrorTests`.

The only tests that I had to disable were SecTrust-related ones, since I couldn't construct a trust instance.

### DIFF
--- a/Source/AFError+OriginalError.swift
+++ b/Source/AFError+OriginalError.swift
@@ -1,0 +1,152 @@
+//
+//  AFError+OriginalError.swift
+//
+//  Copyright (c) 2014-2020 Alamofire Software Foundation (http://alamofire.org/)
+//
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! However, this role is already served by the `underlyingError` property, so why would we need this as well?

danielsaidi:
Hahaha, I looked and looked for such a property, but completely missed it. I will use that then 😅 

Thank you for the quick review.

PR #3306, opened 2020-09-18
Fix to throw an error in MultipartUpload when copying a file to a temporary path fails.

sudopark:
### Goals :soccer:
When uploading a file with MultipartFormData, if the file copy to the temporary path fails, the entire upload process also fails.

I'm using Alamofire very well in my project. Thank you👍👍. Recently, I found the following errors when uploading files with MultipartFormData.

_Cannot read file at file:///private/var/mobile/Containers/Data/Application/041F8DB7-8C7D-4DC8-A976-4F3B9EACE686/tmp/org.alamofire.manager/multipart.form.data/D80342DA-28F4-40A9-913E-CC42161AE546_

The error above occurred when the uploadTask (with: fromFile:) of URLSsession was called to the nonexistent file. To correct the above problem, I modified the build() of MultipartUpload as follows.

```swift
do {
    try multipartFormData.writeEncodedData(to: fileURL)
} catch {
    // Cleanup after attempted write if it fails.
    try? fileManager.removeItem(at: fileURL)

    // WriteEncodedData(to:) can fail for several reasons, such as when a non-existent file url is entered.
    // In this case, I propose a way to throw through errors to ensure that the build step and all upload processes fail.
    throw error
}

```

### Testing Details :mag:
I have added a test to determine if the upload failed when the actual nonexistent file URL was entered.

### DIFF
--- a/Tests/UploadTests.swift
+++ b/Tests/UploadTests.swift
@@ -280,9 +280,9 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
 
         // When
         AF.upload(multipartFormData: { multipartFormData in
-            multipartFormData.append(uploadData, withName: "upload_data")
-            formData = multipartFormData
-        },
+                      multipartFormData.append(uploadData, withName: "upload_data")
+                      formData = multipartFormData
+                  },
                   to: urlString)
             .response { resp in
                 response = resp
@@ -351,9 +351,9 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
 
         // When
         AF.upload(multipartFormData: { multipartFormData in
-            multipartFormData.append(frenchData, withName: "french")
-            multipartFormData.append(japaneseData, withName: "japanese")
-        },
+                      multipartFormData.append(frenchData, withName: "french")
+                      multipartFormData.append(japaneseData, withName: "japanese")
+                  },
                   to: urlString)
             .response { resp in
                 response = resp
@@ -388,9 +388,9 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
 
         // When
         let request = AF.upload(multipartFormData: { multipartFormData in
-            multipartFormData.append(frenchData, withName: "french")
-            multipartFormData.append(japaneseData, withName: "japanese")
-        },
+                                    multipartFormData.append(frenchData, withName: "french")
+                                    multipartFormData.append(japaneseData, withName: "japanese")
+                                },
                                 to: urlString)
             .response { resp in
                 response = resp
@@ -420,9 +420,9 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
 
         // When
         let request = AF.upload(multipartFormData: { multipartFormData in
-            multipartFormData.append(uploadData, withName: "upload_data")
-            formData = multipartFormData
-        },
+                                    multipartFormData.append(uploadData, withName: "upload_data")
+                                    formData = multipartFormData
+                                },
                                 to: urlString)
             .response { resp in
                 response = resp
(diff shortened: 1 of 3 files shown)


Tests/UploadTests.swift
```diff
@@ -518,6 +518,32 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
             XCTFail("Content-Type header value should not be nil")
         }
     }
+    
+    func testThatUploadingMultipartFormDataWithNotExistFileShouldFail() {
```
jshier, line 522:
```suggestion
    func testThatUploadingMultipartFormDataWithNonexistentFileThrowsAnError() {
```

This can be made a bit easier to read.

Tests/UploadTests.swift
```diff
@@ -518,6 +518,32 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
...
+        // Given
+        let urlString = "https://httpbin.org/post"
```
jshier, line 524:
```suggestion
        let urlString = URL.makeHTTPBinURL(path: "post")
```

We have APIs for generating our URLs which should be used instead of raw `String`s.

Tests/UploadTests.swift
```diff
@@ -518,6 +518,32 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
...
+        let imageURL = URL(string: "file://not_exists_file.jpg")!
+
+        let expectation = self.expectation(description: "multipart form data upload with wrong file should fail")
```
jshier, line 527:
```suggestion
        let expectation = self.expectation(description: "multipart form data upload from nonexistent file should fail")
```

Again, this can be made a bit easier to read.

Tests/UploadTests.swift
```diff
@@ -518,6 +518,32 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
             XCTFail("Content-Type header value should not be nil")
         }
     }
+    
+    func testThatUploadingMultipartFormDataWithNotExistFileShouldFail() {
+        // Given
+        let urlString = "https://httpbin.org/post"
+        let imageURL = URL(string: "file://not_exists_file.jpg")!
```
jshier, line 525:
```suggestion
        let imageURL = URL(fileURLWithPath: "does_not_exist.jpg")
```

This makes it a bit safer.

Tests/UploadTests.swift
```diff
@@ -518,6 +518,32 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
...
+        let request = AF.upload(multipartFormData: { multipartFormData in
+            multipartFormData.append(imageURL, withName: "upload_file")
+        },
+                                to: urlString,
+                                usingThreshold: 0).response { resp in
+            response = resp
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: timeout, handler: nil)
```
jshier, line 540:
```suggestion
        waitForExpectations(timeout: timeout)
```

`handler` is unnecessary when `nil`.

Tests/UploadTests.swift
```diff
@@ -518,6 +518,32 @@ class UploadMultipartFormDataTestCase: BaseTestCase {
...
+
+        // Then
+        let uploadable = request.uploadable
+        XCTAssertNil(uploadable)
```
jshier, line 544:
```suggestion
        XCTAssertNil(request.uploadable)
```

Since the `uploadable` value isn't being used it can just be checked inline.

jshier (changes requested):
Nice catch! This change is fundamentally correct but I think the test could be a bit clearer. Also, could you install the latest SwiftFormat and run it in the Alamofire directory? (`swiftformat .`) Once this is updated it can be merged and go out in the next release.

sudopark:
@jshier  Done! thank you for the review 👍

jshier (approved):
Looks good, I'll merge once the tests pass.

jshier (approved):
👍

MarcSteven (commented):
Looks so cool

Issue #3293, opened 2020-09-02
SessionDelegate crash when downloading a file


mbarashkov:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

Our users reported several crashes which originated in SessionDelegate.swift: 302
`fatalError("URLSessionDownloadTask finished downloading with no response.")`
Apparently that happened when request.response was nil

## What did you expect to happen?
I expected Alamofire not to crash the entire app, instead, return error to the file download caller.
Also, I don't understand why in this area of code Alamofire needs request.response at all. It seems like it could be safely deleted.

## What happened
The app crashed

## Alamofire Environment

**Alamofire version:*5.2.2*
**Xcode version:*11.7*
**Swift version:*5.2*
**Platform(s) running Alamofire:*iOS*
**macOS version running Xcode:*10.15.6*

## Demo Project

Can't provide a Demo Project.

jshier:
A `response` is required there because that's how Alamofire's download location APIs work. It's not possible for `URLSessionDownloadTask`s to finish without one, so the assertion ensures the world isn't broken. 

How are you downloading files? Are you using resume data? Are you using any sort of network analytics framework that might be interfering with the proper delegate callbacks? Are you trying to use a background `URLSessionConfiguration`?

mbarashkov:
@jshier 
We are not using resume data or network analytics, or a background configuration. Downloads are very simple:
`
ServerAPIBase.APIManager.sharedManager.download(url, to: destination).response { response in
                if response.error == nil, let downloadPath = response.fileURL?.path {
                    completion(downloadPath)
                } else {
                    let error = response.error?.underlyingError as? URLError
                    let offline = error?.code == URLError.Code.notConnectedToInternet || error?.code == URLError.Code.timedOut
                    failure(offline)
                }
            }
`
(ServerAPIBase.APIManager.sharedManager is:
`
static let sharedManager: Alamofire.Session = {
            if APIManager._sharedManager != nil {
                return APIManager._sharedManager!
            }
            let configuration = URLSessionConfiguration.default
            configuration.timeoutIntervalForRequest = 15
            APIManager._sharedManager = Alamofire.Session(configuration: configuration)
            return APIManager._sharedManager!
         }()
`

Also, I can understand that download failed, ok. But why should it crash the entire app with a fatalError?

[...]

PR #3318, opened 2020-10-03
Fix crash when request download with fileURL

devtak:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/3293

### Goals :soccer:
Crash occurs when request download with fileURL (AF.download)
- modified to throw an error without causing a crash.

If URL is fileURL, throw AFError.invalidURL
```swift
switch request {
    case is DownloadRequest:
        if let url = initialRequest.url, url.isFileURL {
            throw AFError.invalidURL(url: url)
        }
    default:
        break
}
```
### Testing Details :mag:
add downloadTestCase (testDownloadRequestWithLocalFileURLStringThrowsAnError)

### DIFF
--- a/Source/URLRequest+Alamofire.swift
+++ b/Source/URLRequest+Alamofire.swift
@@ -32,6 +32,11 @@ public extension URLRequest {
     }
 
     func validate() throws {
+        if let url = url, url.isFileURL {
+            // This should become another urlRequestValidationFailed error in Alamofire 6.
+            throw AFError.invalidURL(url: url)
+        }
+
         if method == .get, let bodyData = httpBody {
             throw AFError.urlRequestValidationFailed(reason: .bodyDataInGETRequest(bodyData))
         }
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR. However, can you be more specific about the crash you're seeing? Additionally, if a `file` URL doesn't work for `DownloadRequest`, is it also broken for `DataRequest` and the others as well? If that's the case we probably want this logic in a more general location, like our `URLRequest` validation.

jshier:
Investigating this error more, this definitely seems like something that should just be general check in our `validate()` method on `URLRequest`. This lives in our `Alamofire+URLRequest.swift` file. You could add this to the top of the validate function:

```swift
if let url = url, url.isFileURL {
    // This should become another urlRequestValidationFailed error in Alamofire 6.
    throw AFError.invalidURL(url: url)
}
```

devtak:
@jshier i updated it. thanks for the review

jshier (approved):
👍 Nice catch, thanks!

Issue #3294, opened 2020-09-02
App crash, on archived version


aabanaag:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

I experienced a crash but I can't replicate it locally only when I build the app and install it.

## What did you expect to happen?

The app should not crash in the first place

## What happened instead?

ℹ Please replace this with of what happened instead.  

## Alamofire Environment

**Alamofire version: 5.2.2**
**Xcode version: 11.6**
**Swift version: 5**
**Platform(s) running Alamofire:**
**macOS version running Xcode: 10.15.5**

## Demo Project

ℹ Please link to or upload a project we can download that reproduces the issue.

This is the stack trace from firebase crashlytics
```
// This is how I created or initialized sessionManager
internal lazy var sessionManager: Session = {
        let configuration = URLSessionConfiguration.default
        let delegate = Session.default.delegate
        let manager = Session(configuration: configuration, delegate: delegate, startRequestsImmediately: true, cachedResponseHandler: nil, eventMonitors: [AlamofireLogger()])

        return manager
}()
```
```
Crashed: org.alamofire.session.rootQueue
0  libsystem_platform.dylib       0x18dfcc3d0 os_unfair_lock_lock$VARIANT$armv81 + 16
1  Alamofire                      0x104c39fbc Request.lastRequest.getter + 127 (__hidden#2399_:127)
2  Alamofire                      0x104c77f58 _hidden#6970_ (__hidden#7540_)
3  Alamofire                      0x104c2d9a0 _hidden#1656_ (__hidden#855_)
4  libdispatch.dylib              0x18df8a9a8 _dispatch_call_block_and_release
5  libdispatch.dylib              0x18df8b524 _dispatch_client_callout
6  libdispatch.dylib              0x18df68b3c _dispatch_lane_serial_drain$VARIANT$armv81
7  libdispatch.dylib              0x18df6954c _dispatch_lane_invoke$VARIANT$armv81
8  libdispatch.dylib              0x18df7284c _dispatch_workloop_worker_thread
9  libsystem_pthread.dylib        0x18dfdcb74 _pthread_wqthread + 272
10 libsystem_pthread.dylib        0x18dfdf740 start_wqthread + 8
```

jshier:
If you're able to reproduce it, can you create a test project we can use to test?

[...]

PR #3322, opened 2020-10-05
Fix Double Lock in AuthenticationInterceptor

jshier:
### Goals :soccer:
This PR fixes the reported issue reported in [this comment](https://github.com/Alamofire/Alamofire/issues/3294#issuecomment-703241558) where a `Credential` that immediately needs refresh where the refresh is synchronous produces a double lock crash.

This PR fixes the double lock by hopping out of the lock using the interceptor's `queue` before calling the `authenticator`'s `refresh` method.  

### Testing Details :mag:
An additional test case has been added that exercises the bug, along with some test renaming to make it a less attractive copy/paste target.

### DIFF
--- a/Source/AuthenticationInterceptor.swift
+++ b/Source/AuthenticationInterceptor.swift
@@ -210,7 +210,7 @@ public class AuthenticationInterceptor<AuthenticatorType>: RequestInterceptor wh
         var refreshWindow: RefreshWindow?
 
         var adaptOperations: [AdaptOperation] = []
-        var requestsToRetry: [(Alamofire.RetryResult) -> Void] = []
+        var requestsToRetry: [(RetryResult) -> Void] = []
     }
 
     // MARK: Properties
@@ -338,14 +338,16 @@ public class AuthenticationInterceptor<AuthenticatorType>: RequestInterceptor wh
         mutableState.refreshTimestamps.append(ProcessInfo.processInfo.systemUptime)
         mutableState.isRefreshing = true
 
-        authenticator.refresh(credential, for: session) { result in
-            self.$mutableState.write { mutableState in
-                switch result {
-                case let .success(credential):
-                    self.handleRefreshSuccess(credential, insideLock: &mutableState)
-
-                case let .failure(error):
-                    self.handleRefreshFailure(error, insideLock: &mutableState)
+        // Dispatch to queue to hop out of the lock in case authenticator.refresh is implemented synchronously.
+        queue.async {
+            self.authenticator.refresh(credential, for: session) { result in
+                self.$mutableState.write { mutableState in
+                    switch result {
+                    case let .success(credential):
+                        self.handleRefreshSuccess(credential, insideLock: &mutableState)
+                    case let .failure(error):
+                        self.handleRefreshFailure(error, insideLock: &mutableState)
+                    }
                 }
             }
         }
--- a/Tests/AuthenticationInterceptorTests.swift
+++ b/Tests/AuthenticationInterceptorTests.swift
@@ -26,10 +26,10 @@
 import Foundation
 import XCTest
 
-class AuthenticationInterceptorTestCase: BaseTestCase {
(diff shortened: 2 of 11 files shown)


Tests/AuthenticationInterceptorTests.swift
```diff
@@ -447,10 +455,57 @@ class AuthenticationInterceptorTestCase: BaseTestCase {
         XCTAssertEqual(request.retryCount, 1)
     }
 
+    // Produces double lock reported in https://github.com/Alamofire/Alamofire/issues/3294#issuecomment-703241558
+    func testThatInterceptorNeedingImmediateRefreshWithSynchronouslyRefreshRefreshesRequestThatFailedWithOutdatedCredential() {
```
jshier, line 459: This test case could likely use some simplification.

hernangonzalez:
Hi @jshier, changes look good on my project 👍. Thanks!

Source/AuthenticationInterceptor.swift
```diff
@@ -338,14 +338,16 @@ public class AuthenticationInterceptor<AuthenticatorType>: RequestInterceptor wh
...
+                self.$mutableState.write { mutableState in
+                    switch result {
+                    case let .success(credential):
+                        self.handleRefreshSuccess(credential, insideLock: &mutableState)
+                    case let .failure(error):
+                        self.handleRefreshFailure(error, insideLock: &mutableState)
+                    }
                 }
             }
         }
```
cnoon, line 353: I think this approach is fine. It's unfortunate that we need to `hop` for the rare case that someone didn't already hop. The only other thing I'd say we could consider is instead hopping in the completion rather than before the call. In the end though I don't think it matters either way. This will fix the issue and I don't have any better solution.

Tests/AuthenticationInterceptorTests.swift
```diff
@@ -50,47 +50,55 @@ class AuthenticationInterceptorTestCase: BaseTestCase {
...
-            // The 100 ms delay here is important to allow multiple requests to queue up while refreshing
-            DispatchQueue.global(qos: .utility).asyncAfter(deadline: .now() + 0.1) { completion(refreshResult) }
+            if shouldRefreshAsynchronously {
+                // The 10 ms delay here is important to allow multiple requests to queue up while refreshing.
+                DispatchQueue.global(qos: .utility).asyncAfter(deadline: .now() + 0.01) { completion(result) }
+                lock.unlock()
+            } else {
+                lock.unlock()
+                completion(result)
+            }
```
cnoon, line 101: Seems like a solid addition. And I'm all for dropping down to 10 ms as long as CI can handle it.

Tests/AuthenticationInterceptorTests.swift
cnoon: Test seems fine to me overall. Here's my best stab at test name simplification.

```swift
func testThatInterceptorDoesNotDeadlockWhenAuthenticatorCallsRefreshCompletionSynchronouslyOnCallingQueue() {
```

Something like that?

If you have more specific concerns with the test, could you call them out? I'd be happy to dig deeper, but what you doing here makes sense to me and reproduces the reported issue. Also a great idea to add the issue that this test is designed for for reference 👌🏻

cnoon (approved):
Looks good overall @jshier. I had a couple comments back if they're useful. Just let me know if you'd like me to dig in a bit deeper on some of the specifics of that test.

Source/AuthenticationInterceptor.swift
jshier: Tests work either way. However, I am a bit concerned about executing user code while still locked, so I'll leave it as it is. Unfortunately I think the real fix to refactor with direct locking rather than the closure API, but we can do that later.

jdelaune:
I feel like, or something very similar, might have come back. Seeing crashes like this being reported by users.

Within `func refresh(_ credential: Credential, for session: Session, completion: @escaping (Result<Credential, Error>) -> Void)`.

Screenshot 2021-09-13 at 08 05 35

Issue #3374, opened 2020-12-15
Memory leak on request with parameters


Dalein:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

Make a GET request with parameters, and see Debug Memory Graph.

## What did you expect to happen?

No memory leaks at all.

## What happened instead?

Memory leak: on runtime: Memory Issues - (2 leaked types): 1 instance of __NSCFCharacterSet leaked

<img width="1506" alt="Screenshot 2020-12-15 at 13 42 28" src="https://user-images.githubusercontent.com/4835671/102191618-7c5a8880-3edb-11eb-9412-6d9bd4ac52dd.png">

<img width="1198" alt="Screenshot 2020-12-15 at 13 42 59" src="https://user-images.githubusercontent.com/4835671/102191683-909e8580-3edb-11eb-980c-8aa393385b8d.png">

## Alamofire Environment

**Alamofire version: Alamofire (5.4.0)**
**Xcode version: Version 12.2 (12B45b)**
**Swift version: Swift 5**
**Platform(s) running Alamofire: iOS 14.2**
**macOS version running Xcode: 10.15.7 (19H15)**

## Demo Project

You can easily reproduce this issue from here https://github.com/Dalein/AlamofireLeaksExample
Leaks only happens when you make a request with parameters, without it - no leaks happens.

jshier:
This looks like a false positive. As you can see, that `CharacterSet` instance is a static singleton, so it will never be deinited over the lifetime of the application.

bangaruraju:
I'm getting a Memory leak also at this posotion.

PR #3383, opened 2020-12-29
Fixed a memory leak issue in iOS 14.

pcugogo:
### Issue Link :link:
#3374

### Goals :soccer:
Allocating a closure to afURLQueryAllowed results in a memory leak in the NSCFCharacterSet.
This issue occurs since iOS 14.

If don't necessarily need to create afURLQueryAllowed as a closure,
I think it's a better to get rid of the memory issue using the get-only property.

```
public static var afURLQueryAllowed: CharacterSet {
        let generalDelimitersToEncode = ":#[]@" // does not include "?" or "/" due to RFC 3986 - Section 3.4
        let subDelimitersToEncode = "!$&'()*+,;="
        let encodableDelimiters = CharacterSet(charactersIn: "\(generalDelimitersToEncode)\(subDelimitersToEncode)")

        return CharacterSet.urlQueryAllowed.subtracting(encodableDelimiters)
    }
```

### Testing Details :mag:
testThatEscapedCharactersCanBeCustomized case passed, and it runs well in the Example project.

Thanks for Alamofire.

### DIFF
--- a/Source/URLEncodedFormEncoder.swift
+++ b/Source/URLEncodedFormEncoder.swift
@@ -966,11 +966,11 @@ extension CharacterSet {
     /// In RFC 3986 - Section 3.4, it states that the "?" and "/" characters should not be escaped to allow
     /// query strings to include a URL. Therefore, all "reserved" characters with the exception of "?" and "/"
     /// should be percent-escaped in the query string.
-    public static let afURLQueryAllowed: CharacterSet = {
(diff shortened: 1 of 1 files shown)


pcugogo:
Sorry I now understand why I need to create it as a closure
I will close the pull request.

pcugogo:
Sorry I now understand why I need to create it as a closure

PR #3395, opened 2021-01-16
Perhaps an unnecessary check?

slidoooor:
Goals ⚽️:
Because in the later init will go according to the session to determine, whether the early detection here is unnecessary? I don't know much about it, so if I'm wrong, I'm sorry : ) .

Implementation :
Delete a precondition


jshier:
No, this is specifically intended, as Alamofire doesn't support background configurations.

PR #3407, opened 2021-02-09
Use serveral URLCredentials for different protection spaces

Looking4Cache:
### Goals :soccer:
I have some calls that are protected by more than one authentication challenge. At my example I have a client certificate and a http basic auth. The goal is to add the possibility to add several URLCredential objects depending on the authenticationMethod of the protectionSpace.

I added the dictionary credentialsByAuthenticationMethod to the mutableState of a Request. This can hold a URLCredential for each authenticationMethod. In Session.credential I look first if a specific URLCredential is available in credentialsByAuthenticationMethod. If not, I fallback to the existing behaviour. 

### Testing Details :mag:
None.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -106,6 +106,8 @@ public class Request {
         var responseSerializerProcessingFinished = false
         /// `URLCredential` used for authentication challenges.
         var credential: URLCredential?
+        /// All `URLCredential` added by a authenticationMethod of the protection space used for authentication challenges.
+        var credentialsByAuthenticationMethod: [String: URLCredential] = [:]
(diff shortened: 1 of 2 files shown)


EmadBeyrami:
👍🏼

jshier:
Thanks for the PR! Sorry for the delay but this will need testing before it can be merged, which may require some updates to our test server, Firewalk. Can you outline the use cases this supports? What combinations of credentials have you used in the past?

Looking4Cache:
For me the need was that my server uses a client certificate to authorise the user. The server tunnels https requests and if there is a basic auth on the tunnelled page, it was not possible to add these two (client certificate and basic auth) URLCredential to the request.

PR #3417, opened 2021-03-02
Fix #3416

### LINKED ISSUE #3416: The resumeData always returns nil if the download request is interrupted.
This bug may be introduced from Alamofire 5 because resume data was not updated when the download request was interrupted.

Need to get `NSURLSessionDownloadTaskResumeData` from the `error.userInfo`.

tsuiyuenhong:
### Issue Link :link:
#3416 

update the resume data from the `error.user[NSURLSessionDownloadTaskResumeData]` when the download request was interrupted.

### DIFF
--- a/Source/AFError.swift
+++ b/Source/AFError.swift
@@ -433,6 +433,13 @@ extension AFError {
         guard case let .downloadedFileMoveFailed(_, _, destination) = self else { return nil }
         return destination
     }
+   
+    /// The resumeData of a `.sessionTaskFailed` error.
+    /// Only works for `DownloadRequest`
+    public var resumeData: Data? {
+        guard case let .sessionTaskFailed(error) = self else { return nil }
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR but we won't be accepting it. Accessing the `resumeData` through the response error and not derived from the local error should work fine and keeps the API surface that much smaller. If you need the computed property you can add it yourself.

tsuiyuenhong:
is it as expected that now the resume data return nil when the network is interrupted?

jshier:
You're right, it does make sense to surface this from the error when it's available. Hard part's going to be testing it. Reopening while I investigate.

tsuiyuenhong:
Thanks!

I found the issue when migrating Alamofire4 to Alamofire5. It may be more reasonable to allow users to access resume data directly instead of accessing from the error.

jshier:
Fixed in #3419.

PR #3419, opened 2021-03-07
Add Support for Resume Data on Error

jshier:
### Issue Link :link:
#3416

### Goals :soccer:
This PR adds support for accessing any resume data produced when a `DownloadRequest` fails.

This PR adds a convenience property to `AFError` to access the data and uses it as a fallback in the `DownloadRequest` property.

### Testing Details :mag:
With Firewalk's help the `DownloadRequest` resume data tests have been updated and an additional case added for availability on error.

### DIFF
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -1528,11 +1528,11 @@ public class DownloadRequest: Request {
     @Protected
     private var mutableDownloadState = DownloadRequestMutableState()
 
-    /// If the download is resumable and eventually cancelled, this value may be used to resume the download using the
-    /// `download(resumingWith data:)` API.
+    /// If the download is resumable and is eventually cancelled or fails, this value may be used to resume the download
+    /// using the `download(resumingWith data:)` API.
     ///
     /// - Note: For more information about `resumeData`, see [Apple's documentation](https://developer.apple.com/documentation/foundation/urlsessiondownloadtask/1411634-cancel).
-    public var resumeData: Data? { mutableDownloadState.resumeData }
+    public var resumeData: Data? { mutableDownloadState.resumeData ?? error?.downloadResumeData }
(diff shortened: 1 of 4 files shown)


Source/Request.swift
```diff
@@ -1528,11 +1528,11 @@ public class DownloadRequest: Request {
...
     private var mutableDownloadState = DownloadRequestMutableState()
 
-    /// If the download is resumable and eventually cancelled, this value may be used to resume the download using the
-    /// `download(resumingWith data:)` API.
+    /// If the download is resumable and is eventually cancelled or fails, this value may be used to resume the download
+    /// using the `download(resumingWith data:)` API.
     ///
     /// - Note: For more information about `resumeData`, see [Apple's documentation](https://developer.apple.com/documentation/foundation/urlsessiondownloadtask/1411634-cancel).
-    public var resumeData: Data? { mutableDownloadState.resumeData }
+    public var resumeData: Data? { mutableDownloadState.resumeData ?? error?.downloadResumeData }
```
tsuiyuenhong, line 1535: Return the `error?.downloadResumeData` in computed property may not be a good solution. It will still get wrong data when other developers use `mutableDownloadState` internally. In my view, a better solution is to assign `error?.downloadResumeData`  directly to `mutableDownloadState`.

jshier: Nobody should be doing that, as it's internal to the library.

tsuiyuenhong: Sorry, I didn't make it clear. The other developers mean the contributor of Alamofire, if they wanna use `mutableDownloadState` in the `Request.swifit`, they may get the wrong return.

jshier: That value is documented to be the value from cancellation, and the documentation of the property should make it clear it can come from two sources, so I don't think there's an issue.

PR #3427, opened 2021-03-25
Convert header string to constant

JYlab:
### Goals :soccer:
Convert header string to constant.

Although there are many re-used strings, continuing to use them as strings can lead to messy code and bugs.

### DIFF
--- a/Source/ConstantPool.swift
+++ b/Source/ConstantPool.swift
@@ -1,0 +1,51 @@
+//
+//  ConstantPool.swift
+//  Alamofire
+//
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! However, strings which are only used once are not subject to the typical rules around constants. And as you can see in the code, your changes make it harder to read. So we will not be accepting these changes. Thanks anyway!

JYlab:
Thank you for your quick feedback. I'm analyzing this open source because it's well developed. I'll tell you again if I see anything to contribute  :)

PR #3431, opened 2021-04-06
Update Alamofire 5.0 Migration Guide

DevYeom:
### Goals :soccer:
- Fixed a typo so that there is no confusion when referring to the guide.


jshier (approved):
Thanks!

PR #3437, opened 2021-04-15
Add We can also work with HTTPS Protocol.

sarvesh567:
remove confusion b/w user we can also use alamofire networking with https protocol.
Although this is very small change but  create very effective impact on the users.


jshier:
Thanks for the PR! However, there's no need to update the Alamofire 2 Migration Guide, so we won't be merging it.

sarvesh567:
Thanks Joe
Best Regards
Sarvesh Patel
iOS Developer

PR #3438, opened 2021-04-16
Multipart Form Data Encoding Fix

cnoon:
This PR fixes up a regression introduced by PR #3421 where clients relying on the `URLRequest` creation process to append body parts to `MultipartFormData` objects were no longer doing so. The body parts would no longer be appended since the `Uploadable` creation was happening before the `URLRequest` creation took place.

As @jshier and I looked into the issue further, it became apparent that we needed to redesign the logic to instead operate as follows:

1. Create the initial `URLRequest` (this is where I'm appending the body parts)
2. Adapt the `URLRequest` with any appended interceptors
3. Create the `Uploadable` which will encode the `MultipartFormData`
4. Create the`URLSessionUploadTask` which will actually upload the data

This approach will ensure we not only allow clients to append body parts as part of the `URLRequestConvertible` conformance prior to encoding, but it also allows us to avoid encoding the `MultipartFormData` if the `URLRequest` creation or adaptation fails. Win win!

### Issue Link :link:
No original issue filed.

### Goals :soccer:
- To allow clients to continue to append body parts to `MultipartFormData` as part of the `URLRequest` creation process
- To streamline the `MultipartFormData` upload process to only encode the data if the `URLRequest` creation and adaptation process succeeds

Minor change internally where we're just re-ordering the operations for uploads.

### Testing Details :mag:
Added a test which ensures the body parts are appended to the `MultipartFormData` as part of the `URLRequest` creation process and also verifies the size of the data being uploaded.

### DIFF
--- a/Source/Session.swift
+++ b/Source/Session.swift
@@ -420,7 +420,7 @@ open class Session {
                             requestModifier: RequestModifier? = nil) -> DataStreamRequest {
         let convertible = RequestEncodableConvertible(url: convertible,
                                                       method: method,
-                                                      parameters: Optional<Empty>.none,
+                                                      parameters: Empty?.none,
                                                       encoder: URLEncodedFormParameterEncoder.default,
                                                       headers: headers,
                                                       requestModifier: requestModifier)
--- a/Source/Session.swift
+++ b/Source/Session.swift
@@ -1019,13 +1019,15 @@ open class Session {
     func performUploadRequest(_ request: UploadRequest) {
         dispatchPrecondition(condition: .onQueue(requestQueue))
 
-        do {
-            let uploadable = try request.upload.createUploadable()
-            rootQueue.async { request.didCreateUploadable(uploadable) }
-
-            performSetupOperations(for: request, convertible: request.convertible)
-        } catch {
-            rootQueue.async { request.didFailToCreateUploadable(with: error.asAFError(or: .createUploadableFailed(error: error))) }
+        performSetupOperations(for: request, convertible: request.convertible) {
+            do {
+                let uploadable = try request.upload.createUploadable()
+                self.rootQueue.async { request.didCreateUploadable(uploadable) }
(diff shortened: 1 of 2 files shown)


Source/Session.swift
```diff
@@ -1058,6 +1063,7 @@ open class Session {
         guard !request.isCancelled else { return }
 
         guard let adapter = adapter(for: request) else {
+            guard shouldCreateTask() else { return }
```
jshier, line 1066: I don't think we should do this work twice. The check below should be enough, so do we want to remove this one?

Tests/UploadTests.swift
```diff
@@ -580,6 +580,54 @@ final class UploadMultipartFormDataTestCase: BaseTestCase {
...
+        waitForExpectations(timeout: timeout)
+
+        // Then
+        XCTAssertNotNil(response?.request)
+        XCTAssertNotNil(response?.response)
+        XCTAssertNotNil(response?.data)
+        XCTAssertNil(response?.error)
+
+        switch uploadRequest.uploadable {
+        case .data(let data):
```
jshier, line 623: Our usual style is `case let`, so make sure to run the formatter and it'll fix it.

jshier (changes requested):
Looks great! I don't think we need to call `shouldCreateTask` twice, so I'm pretty sure one use is left over. That and running the formatter and this should be good to go.

Source/Session.swift
jshier: Ah I see why, sorry.

cnoon: It doesn't. This is inside the `guard` for when we don't have an `adapter`.

Tests/UploadTests.swift
cnoon: On it!

jshier (approved):
👍

PR #3443, opened 2021-04-22
ServerTrustEvaluating add task

lijian096870:
Problem:

in the same session, there may be many requests at the same time. For the same host service host, there may be different types of services that need to be verified differently. At this time, it is impossible to distinguish which request is issued by the host only,

Solution:

after task is supplemented, you can obtain URLsessionTask by override func servertrustevaluator (for task: urlsessiontask, for host: String) task.taskIdentifier Attribute matching request.onURLSessionTaskCreation (performance handler: @escaping (urlsessiontask) - > void) identifies which request is specific.

### DIFF
--- a/Source/ServerTrustEvaluation.swift
+++ b/Source/ServerTrustEvaluation.swift
@@ -69,6 +69,22 @@ open class ServerTrustManager {
 
         return evaluator
     }
+    
+    /// Returns the `ServerTrustEvaluating` value for the given host, if one is set.
+    ///
+    /// By default, this method will return the policy that perfectly matches the given host. Subclasses could override
+    /// this method and implement more complex mapping implementations such as wildcards.
+    ///
+    /// - Parameter host: The host to use when searching for a matching policy.
+    ///
+    /// - Parameter task: The task to use when searching for a matching policy.
+    ///
+    /// - Returns:        The `ServerTrustEvaluating` value for the given host if found, `nil` otherwise.
+    /// - Throws:         `AFError.serverTrustEvaluationFailed` if `allHostsMustBeEvaluated` is `true` and no matching
+    ///                   evaluators are found.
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR! Before evaluating the change, can you expand on why you'd ever need more than one trust for a single host? I mean, it's technically possible to have any number of certificates for host, but why would you ever need to do so?

Additionally, any changes to the server trust logic must be accompanied by extensive tests, as it's a hugely sensitive part of Alamofire. Currently our Firewalk test server doesn't support TLS at all, much less returning multiple certs for a single host, so we'd need an enhancement there to enable this feature.

Finally, it seems unlikely we'd want to implement this on a per-task basis, especially when the functionality you've implemented requires subclassing to actually use. Instead, we'd want to allow `Request`s provide their own trust checking, which the `SessionDelegate` would then call, if available. This way you can easily customize this behavior when making a request. For example, it could look something like this:

```swift
AF.request(...)
  .serverTrustEvaluator(...)
  .response { ... }
```

lijian096870:
Thank you for your reply. I also think it's rare to have different certificates for the same host. It's not necessary. If we really need to identify the specific request, we can implement the subclass of SessionDelegate, but it cannot be used Alamofire.ServerTrustManager Related functions. So, this is how I did it

AF.request (...)
.serverTrustEvaluator{(task,challenge,completionHandler) -> bool }
.response { ... }

I added a return value to represent whether it has been processed. If it returns false, it will be processed according to the Alamofire.ServerTrustManage handle. If it returns true, it is not executed Alamofire.ServerTrustManage method.

Not every request requires call request.serverTrustEvaluator{(task,challenge,completionHandler) -> bool }. It's just a general idea.   The specific request can be identified according to the task.

In addition, this PR is not very necessary. I've solved the problem. At present, our test team has not found any problems. Your can handle it according to your evaluation.

Thanks

sweetmans (approved):

PR #3449, opened 2021-05-02
Add watchOS Testing

jshier:
### Goals :soccer:
This PR adds watchOS testing now that it's possible in Xcode 12.5.

This PR adds a test target and fixes a few build issues with the tests on watchOS.

A test runner was added.

### Testing Details :mag:
It was necessary to disable a few tests:
- `testThatRevokedCertificateRequestFailsWithRevokedServerTrustPolicy` was disabled, as watchOS doesn't seem to do any revocation checking at all.
- A few retrier tests in `SessionTests` were failing, as watchOS seems to have some sort of implicit credential storage that we can't clear, making these tests fail since earlier tests provide a credential with the same user / password. Tests using HTTP Basic auth have been updated to all use different user names, allowing the tests to pass on watchOS.


Tests/SessionTests.swift
```diff
@@ -815,17 +815,20 @@ final class SessionTestCase: BaseTestCase {
 
     // MARK: Tests - Request Retrier
 
-    func testThatSessionCallsRequestRetrierWhenRequestEncountersError() {
+    func testThatSessionCallsRequestRetrierWhenRequestInitiallyEncountersAdaptError() {
```
jshier, line 818: No tests were changed, I just moved some around so all of the retrier tests could be together so they can be disabled.

PR #3455, opened 2021-05-23
Add WebSocketRequest Under @_spi(WebSocket)

jshier:
### Goals :soccer:
This PR adds `WebSocketRequest`, an Alamofire wrapper around `URLSessionWebSocketTask`, as well as assorted functionality for dealing with WebSockets. This includes:

- Automatic listening: no more manual `listen` calls on the task.
- Automatic ping support: ping the server automatically given an interval.
- Response serialization support: parse messages into objects, including using `Decodable`.
- Concurrency support: full streams of events and messages.

`WebSocketRequest` follows most of Alamofire's existing patterns. There are a few differences:

- Closure vs. cancellation: web sockets can close expectedly, yet this closure produces an error from `URLSession`. This had to be customized in order to ignore the error and allow the user to close a connection with no additional issue.
- Like `DataStreamRequest`, `WebSocketRequest` deals with incoming streams of data. However, that data is already framed and then partially decoded into `Message`s by `URLSession`, so serialization had to be built at that level.
- Unlike `DataStreamRequest`, `WebSocketRequest` can send `Message` back up the connection. This requires some queuing on the request side to ensure messages sent before the connection is established are properly stored for sending.

### Testing Details :mag:
Full suite of `WebSocketTests` added, as well as the concurrency versions. Firewalk was updated with several new `websocket` cases to support scenarios like an echo server or expected ping count.

P.S. This feature was delayed so long due to the many hours of frustration I had to deal with when Apple shipped a message dropping bug in iOS 14.5 - 14.7. In trying to test the socket connections I would connect to the test server, have it reply with a message, and then close the connection. Shockingly that meant that *sometimes* the messages wouldn't come through. But sometimes they would. Investigating other web socket clients and inspecting network traffic with Wireshark revealed that the messages were being sent by the server but the client just wasn't receiving them. Further investigation showed that delaying the close of the connection by 60ms or so let the client receive the message reliably. After many hours I finally had reliable tests! Only to then test on the iOS 15 betas and see that I didn't need the delay at all, the tests were reliable without it! Turns out it was just a bug in iOS 14.5 - 14.7. I was so disgusted by this fact that I gave up on the feature for years. I'll call out this bug in the documentation.

### DIFF
--- a/Tests/WebSocketTests.swift
+++ b/Tests/WebSocketTests.swift
@@ -64,6 +64,7 @@ final class WebSocketTests: BaseTestCase {
     func testThatWebSocketsCanReceiveAMessage() {
         // Given
         let didReceiveMessage = expectation(description: "didReceiveMessage")
+        let didComplete = expectation(description: "didComplete")
 
         let session = stored(Session())
 
@@ -73,6 +74,9 @@ final class WebSocketTests: BaseTestCase {
         session.websocketRequest(.websocket()).streamMessages { message in
             receivedMessage = message
             didReceiveMessage.fulfill()
+        }
+        .onCompletion {
+            didComplete.fulfill()
         }
 
         waitForExpectations(timeout: timeout)
@@ -132,6 +136,7 @@ final class WebSocketTests: BaseTestCase {
     func testThatWebSocketsCanReceiveADecodableValue() {
         // Given
         let didReceiveValue = expectation(description: "didReceiveMessage")
+        let didComplete = expectation(description: "didComplete")
 
         let session = stored(Session())
 
@@ -141,6 +146,9 @@ final class WebSocketTests: BaseTestCase {
         session.websocketRequest(.websocket()).streamDecodable(TestResponse.self) { value in
             receivedValue = value
             didReceiveValue.fulfill()
+        }
+        .onCompletion {
+            didComplete.fulfill()
         }
 
         waitForExpectations(timeout: timeout)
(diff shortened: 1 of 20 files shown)


lmcd:
Awesome to see this 👍 
Is this landing anytime soon?

Tests/WebSocketTests.swift
```diff
@@ -0,0 +1,300 @@
...
+            return left == right
+        case let (.data(left), .data(right)):
+            return left == right
+        default:
+            return false
+        }
+    }
+}
+
+#endif
```
ux-8, line 300: TesTests/WebSocketTests.swiftts/WebSocketTests.swift

Mkultra555:
Thank you 😊

nikolay-kapustin:
OMG yeah! Push it!

lmcd:
@jshier - is there any intent on getting this in at all, or are these commits for your own fork/personal use?

jshier:
@lmcd Yes, it's intended to ship publicly. It's just a very large feature I need to get right. I'm about to merge it with the feature under `@_spi(WebSocket)` as I need to do a bunch of project cleanup and don't have time to write all the documentation I need. Once that's done I can get back to finalizing it with docs.

jshier:
I'm merging this under `@_spi(WebSocket)` while I work on some project cleanup PRs. That SPI can be used but will be unstable until it's removed. If I need to ship an intermediate version I can, but I intend to get back to finishing up tests and documentation before shipping this feature in general.

galipMNL:
Any updates on moving this out of SPI into stable?

jshier:
@galipMNL I'm completing WebSocketRequest in https://github.com/Alamofire/Alamofire/pull/4039. I believe it's functionally complete but I need to finish the documentation and test refactor before it can be released. I just got back from vacation, so I hope to get it finished soon. Feel free to give it a try!

Issue #3461, opened 2021-05-28
Windows Support


alex-taffe:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## Problem

Alamofire does not support Windows. This is related to #1935 as much of the work is the same

## Feature Request

Alamofire support on Windows

## Value to Alamofire

Alamofire should support all Swift supported platforms

## Work to be done

- [x] #3460 excludes NetworkReachabilityManager from Windows, as it is not available (issue #3459)
- [ ] The work from #3446 needs to be merged, as much of it applies to Windows as well
- [x] #3462 needs to be merged into Linux

jshier:
This has been merged and will go out with the next release. Note that there will be no official support, as Foundation on Linux and Windows is buggy and incomplete. Thanks for your help!

PR #3462, opened 2021-05-28
Initial Windows Support

alex-taffe:
### Issue Link :link:
#3461

### Goals :soccer:
To add support to Alamofire for Windows

A Windows specific mutex wrapper had to be created. Otherwise, the same changes that apply to Linux apply here as well.

### Testing Details :mag:
I’ve added Windows CI, but haven’t been able to test it yet because I need maintainer approval

### DIFF
--- a/Source/Protected.swift
+++ b/Source/Protected.swift
@@ -51,18 +51,8 @@ extension Lock {
 
 #if os(Linux) || os(Windows)
 
-/// An `NSLock` wrapper.
-final class MutexLock: Lock {
-    private var mutex = NSLock()
+extension NSLock: Lock { }
 
-    fileprivate func lock() {
-        self.mutex.lock()
-    }
-
-    fileprivate func unlock() {
-        self.mutex.unlock()
-    }
-}
 #endif
 
 #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
@@ -97,7 +87,7 @@ final class Protected<T> {
     #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
     private let lock = UnfairLock()
     #elseif os(Linux) || os(Windows)
-    private let lock = MutexLock()
+    private let lock = NSLock()
     #endif
     private var value: T
 


alex-taffe:
@jshier this should be ready for review

Source/Protected.swift
```diff
@@ -82,6 +82,45 @@ final class MutexLock: Lock {
...
 
+#if os(Windows)
+import WinSDK
+private func randomString(length: Int) -> String {
+    let letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+    return String((0..<length).map{ _ in letters.randomElement()! })
+}
+
+/// A `CreateMutexA` wrapper.
+final class MutexLock: Lock {
```
jshier, line 93: I may be better, at least for this initial version, to wrap `NSLock` on Windows.  I don't understand the Windows locking primitives at all, so it would probably be better to use an implementation that has already been proven.

.github/workflows/ci.yml
```diff
@@ -137,3 +137,34 @@ jobs:
...
+        echo "C:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+    - name: Install Supporting Files
+      run: |
+        Copy-Item "$env:SDKROOT\usr\share\ucrt.modulemap" -destination "$env:UniversalCRTSdkDir\Include\$env:UCRTVersion\ucrt\module.modulemap"
+        Copy-Item "$env:SDKROOT\usr\share\visualc.modulemap" -destination "$env:VCToolsInstallDir\include\module.modulemap"
+        Copy-Item "$env:SDKROOT\usr\share\visualc.apinotes" -destination "$env:VCToolsInstallDir\include\visualc.apinotes"
+        Copy-Item "$env:SDKROOT\usr\share\winsdk.modulemap" -destination "$env:UniversalCRTSdkDir\Include\$env:UCRTVersion\um\module.modulemap"
+    - name: SPM Windows build
+      shell: cmd
+      run: |
```
jshier, line 166: Does this just build or does it run tests? I see `--enable-test-discovery` which is why I ask. I also though test discovery was enabled by default now?

jshier (changes requested):
Thanks for the work! I have one question as well as one change: we should probably use `NSLock` for Windows since the locking primitives are so different from the other platforms.

.github/workflows/ci.yml
alex-taffe: Currently this just builds. The test discovery can go, that was just leftover from a copy paste of an old swift build of mine, will fix

Source/Protected.swift
alex-taffe: Done

Source/Protected.swift
```diff
@@ -49,35 +49,27 @@ extension Lock {
...
-        pthread_mutexattr_settype(&attr, .init(PTHREAD_MUTEX_ERRORCHECK))
+#if os(Linux) || os(Windows)
 
-        let error = pthread_mutex_init(mutex, &attr)
-        precondition(error == 0, "Failed to create pthread_mutex")
-    }
+/// An `NSLock` wrapper.
+final class MutexLock: Lock {
+    private var mutex = NSLock()
+    private var locked = false
```
jshier, line 57: Do we need this at all? Shouldn't `NSLock` check this itself?

alex-taffe: I just left it so if our wrapper goes out of scope and gets deallocated, the mutex still unlocks. To be honest, now that I’m thinking about it. We should use `DispatchQueue` instead of NSLock. NSLock has the following disadvantage:

```
The NSLock class uses POSIX threads to implement its locking behavior. When sending an unlock message to an NSLock object, you must be sure that message is sent from the same thread that sent the initial lock message. Unlocking a lock from a different thread can result in undefined behavior.
```

jshier: We already make that guarantee on Alamofire's internal queues with our closure wrapper. `DispatchQueue`s are basically the slowest locking implementation and so we avoid them for simply guaranteeing state safety.

In any case, I don't think we need to track this locking at all, just let the lock do it if it's important.

alex-taffe: Got it

alex-taffe: Pushed

Source/Protected.swift
jshier: Sorry, now that I think about it, what about just making `NSLock` conform to our `Lock` protocol and using it directly?

alex-taffe: That’s probably better, do you think we even need the apple OS implementation of `UnfairLock`? Just add an extension to NSLock. This on its own compiles:
```
extension NSLock: Lock { }
```
because NSLock already has `lock` and `unlock` methods?

jshier: Yeah, then we can use it on the other platforms. We should keep `UnfairLock` since it's much faster on Apple platforms.

alex-taffe: Got it, pushed

jshier (approved):

PR #3463, opened 2021-06-08
Swift Concurrency Support

jshier:
### Goals :soccer:
This PR tracks the implementation of Swift Concurrency support as an extension to Alamofire. This ensures it remains buildable on all platforms.

Additions:

- `DataTask`, which encapsulates the originating `DataRequest` and a `Task` which tracks an underlying continuation for async work. It also provides async access to the `response`, `result`, and `value` properties, as well as read-only access to the properties of the underlying `DataRequest`.
- `DownloadTask`, which does the same for `DownloadRequest`.
- `DataStreamTask`, which encapsulates a `DataStreamRequest` and vends custom `StreamOf` values for the stream output. Unlike the data and download cases, this task doesn't trigger the request immediately, only once a stream has been requested.

### Testing Details :mag:
Tests added for all task types.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        destination: ["OS=15.0,name=iPhone 13 Pro"]
+        destination: ["OS=15.2,name=iPhone 13 Pro"]
     steps:
       - uses: actions/checkout@v2
       - name: Install Firewalk
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,9 +96,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install Firewalk
-        run: arch -arch arm64e brew install alamofire/alamofire/firewalk || arch -arch arm64e brew upgrade alamofire/alamofire/firewalk && arch -arch x86_64 firewalk &
+        run: brew install alamofire/alamofire/firewalk || brew upgrade alamofire/alamofire/firewalk && firewalk &
       - name: iOS - ${{ matrix.destination }}
-        run: set -o pipefail && arch -arch arm64e env NSUnbufferedIO=YES xcodebuild -project "Alamofire.xcodeproj" -scheme "Alamofire iOS" -destination "${{ matrix.destination }}" -enableThreadSanitizer NO clean test | xcpretty
+        run: set -o pipefail && env NSUnbufferedIO=YES xcodebuild -project "Alamofire.xcodeproj" -scheme "Alamofire iOS" -destination "${{ matrix.destination }}" -enableThreadSanitizer NO clean test | xcpretty
   tvOS:
     name: Test tvOS
     runs-on: firebreak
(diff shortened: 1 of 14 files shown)


Tlz-1994:
所以这个 async/await 的特性只能用在 iOS15+ 吗？

jshier:
Right now, yes, the Swift Concurrency features are only available on the 2021 OS versions.

Tlz-1994 (commented):
> ‎现在，是的，Swift 并发功能仅在 2021 OS 版本上提供。‎

It's hard to understand why this feature wasn't embedded in Swift5.5,
why do you think it was bundled with the OS.

Source/Async.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    public let handle: Task.Handle<AFDataResponse<Value>, Never>
+
+    fileprivate init(request: DataRequest, handle: Task.Handle<AFDataResponse<Value>, Never>) {
+        self.request = request
+        self.handle = handle
+    }
+}
+
+extension DispatchQueue {
```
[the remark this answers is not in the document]
jshier: Alamofire style puts visibility modifiers on individual items rather than the extension declaration, to preserve local reasoning.

jshier:
> It's hard to understand why this feature wasn't embedded in Swift5.5,
> why do you think it was bundled with the OS.

You can read more about the issue on the [Swift Forums](https://forums.swift.org/t/will-swift-concurrency-deploy-back-to-older-oss/49370).

Tlz-1994:
> > It's hard to understand why this feature wasn't embedded in Swift5.5,
> > why do you think it was bundled with the OS.
> 
> You can read more about the issue on the [Swift Forums](https://forums.swift.org/t/will-swift-concurrency-deploy-back-to-older-oss/49370).

Wow, thanks for your reply.

Source/Async.swift
```diff
@@ -0,0 +1,177 @@
...
+                          completionHandler: $0)
+        }
+    }
+
+    private func dataTask<Value>(forResponse onResponse: @escaping (@escaping (DataResponse<Value, AFError>) -> Void) -> Void) -> DataTask<Value> {
+        let handle = async {
+            await withTaskCancellationHandler {
+                self.cancel()
+            } operation: {
+                await withCheckedContinuation { continuation in
```
horatiu03, line 165: I think that using `withUnsafeContinuation` will be beter for performance.

jshier: Not in any noticeable fashion since we're waiting for a network response and all of Alamofire's other internal machinery to complete anyway. Generally Swift prefers the safe operations (checked arithmetic, bounds-checked collections, etc.) to unsafe unless there's critical performance to be reclaimed, which isn't the case here.

afern247:
Is there any way to wrap completion handlers in the meantime? https://stackoverflow.com/questions/68694917/convert-alamofire-completion-handler-to-async-await-swift-5-5/68694995#68694995

Flatout73:
Any ETA when it will be merged? btw, now concurrency will be supported from iOS 13 since xcode 13.2.

jshier:
The recent update to this branch requires Xcode 13.2, as there's no way to deploy versions of async code that can deploy back to the 2019 OSes and the original deployment targets at the same time. Naming has also been updated, as well as the addition of configurable cancellation handling. I hope to round out the various types of support soon.

horatiu03:
Will there be some example code on how async should/inteaded to be used with Alamofire?

jshier:
Yes, updated docs will be part of this PR.

kyzmitch:
probably it doesn't fully relates to this PR, but did apple change something in Xcode 13 (release version) which affected Alamofire compilation in this PR? I had my own compilable async code in Xcode 13 Beta but now in released version it doesn't want to compile URLSession related async code:
```
let (data, response) = try await urlSession.data(for: someHttpRequest, delegate: nil)
```
with an error which is not easy to understand: `Type of expression is ambiguous without more context`

flowbe:
Hello,
Now that Xcode 13.2 is in RC, will this be merged soon?

jshier:
I've added some summary documentation to our AdvancedUsage document, as well as extensive inline documentation. Please let be know if there are any additional questions that can be answered in the docs.

This work will be merged and released as soon after the release of Xcode 13.2 as I can manage.

cnoon (approved):
This looks great @jshier! Great work across the board!

PR #3476, opened 2021-07-09
doc: fix CachedResponseHandler ResponseCacher documentation

r-mckay:
Hello!

### Goals :soccer:
This PR updates the documentation of the ResponseCacher in CachedResponseHandler.
It changes the notion of Follow by the notion of Cache. It tries to follow the convention used in Redirector from RedirectHandler.swift.

The documentation is also generated in a second commit, which is dedicated to the jazzy run.

Thanks!


jshier:
Thanks for the PR! We only update the Jazzy docs as part of a release, so please remove those changes from the PR.

jshier:
Thanks!

PR #3478, opened 2021-07-13
changed count >= 1 to not isEmpty

shafeerp:
changed `path.count >= 1 ` to `!path.isEmpty`.
`count` is costly operation here.


luvacu (approved):

jshier:
There's no performance difference here, as `path` is an array, where both `count` and `isEmpty` have the same performance. But fair enough. Thanks for the PR!

Goga12345678:
Goga

avanishmaurya:
how can i contribute.

Issue #3481, opened 2021-07-16
Xcode 13 betas: Build fails in Combine


phiren:
Facing below error:
SourcePackages/checkouts/Alamofire/Source/Combine.swift:25:5: Failed to build module 'Combine'; this SDK is not supported by the compiler (the SDK is built with 'Apple Swift version 5.5 (swiftlang-1300.0.24.14 clang-1300.0.25.10)', while this compiler is 'Apple Swift version 5.5 (swiftlang-1300.0.24.13 clang-1300.0.25.10)'). Please select a toolchain which matches the SDK.

## What did you expect to happen?

it should build project successfully.

## What happened instead?

Its throwing error. Same problem was in Kingfisher GitHub repository, They created new branch with fix, changed deployment target from v10 to 12

## Alamofire Environment

**Alamofire Version:** 5.2.1
**Dependency Manager:** Swift Package Manager
**Xcode Version:** Version 13.0 beta 3 (13A5192i)
**Swift Version:** Swift 5
**Platform(s) Running Alamofire:** iOS
**macOS Version Running Xcode:** 12.0 Beta (21A5284e)

jshier:
Yeah, looks like they broken Combine compilation in beta 3. You'll need to go back to beta 2 or wait for the next beta.

jeremyfalldev:
Don't we have a temporary solution?

jshier:
@jeremyfalldev Only solution is to remove all use of Combine until they fix the issue. I don't think that's worth a branch that will (hopefully) only live a week, but anyone who really needs it can remove the `Combine.swift` file which should prevent the linking of Combine. If the issue continues we could create a longer running fix branch.

jeremyfalldev:
@jshier Thank you for the quick response. I thought there is an easy & quick solution that is similar to what Kingfisher did.
Comments from @phiren:
> Same problem was in Kingfisher GitHub repository, They created new branch with fix, changed deployment target from v10 to 12

jeremyfalldev:
@jshier Any idea?

castor55:
Same issue here. I use SPM and get an error while archiving a new build.
Any solution?

zhangqifan:
Beta: Xcode 13 Beta 4

I am using CocoaPods and switch the version back to 5.1.0(without Combine). It works for me. Still waiting for the solution, maybe on beta 5?

jshier:
I can't believe this is still broken. I'll see if I can get a branch up over the next few days.

[...]

PR #3486, opened 2021-08-11
Xcode 13 Support

jshier:
### Issue Link :link:
#3481

### Goals :soccer:
This PR tracks changes necessary for Xcode 13. It also adds a workaround for Combine-related build issues by limiting Combine features to 64-bit platforms as well as the ability to import Combine.

No functional changes, just deployment targets.

### Testing Details :mag:
No new tests.

### DIFF
--- a/Source/Combine.swift
+++ b/Source/Combine.swift
@@ -22,7 +22,7 @@
 //  THE SOFTWARE.
 //
 
-#if canImport(Combine)
+#if !(os(iOS) && (arch(i386) || arch(arm)))
 
 import Combine
 import Dispatch
--- a/Source/Combine.swift
+++ b/Source/Combine.swift
@@ -22,7 +22,7 @@
 //  THE SOFTWARE.
 //
 
-#if !(os(iOS) && (arch(i386) || arch(arm)))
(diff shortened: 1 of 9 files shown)


Source/Combine.swift
```diff
@@ -22,7 +22,7 @@
 //  THE SOFTWARE.
 //
 
-#if canImport(Combine)
+#if !(os(iOS) && (arch(i386) || arch(arm)))
```
jlcvp, line 25: this check must be broader to check for windows and linux that doesn't support Combine framework

Tests/CombineTests.swift
```diff
@@ -22,7 +22,7 @@
 //  THE SOFTWARE.
 //
 
-#if canImport(Combine)
+#if !(os(iOS) && (arch(i386) || arch(arm)))
```
jlcvp, line 25: Same here

jlcvp (commented):
As the lib also supports running on Linux and Windows, we need to be broader in the `#if` directive to embrace those systems in the verification

jlcvp (approved):
Seems great now, thanks

PR #3494, opened 2021-09-11
Fix the screen title for each request in the sample app

TiagoMaiaL:
### Goals :soccer:
Fix the title of the `DetailViewController`, in the iOS sample app. It permanently displays the following title:  **No request created yet.**, even after the test request is performed (more context in the gif files).

This happens because the `title` is changed too early, before the internal `URLRequest` gets configured (which is used in the `description` property).

To fix this issue, the controller now assigns a closure to the `onURLRequestCreation` method on the `Request` type. Inside this closure, it updates its title.

### DIFF
--- a/Example/Source/DetailViewController.swift
+++ b/Example/Source/DetailViewController.swift
@@ -35,6 +35,10 @@ class DetailViewController: UITableViewController {
             oldValue?.cancel()
 
             title = request?.description
+            request?.onURLRequestCreation { [weak self] _ in
+                self?.title = self?.request?.description
+            }
+            
             refreshControl?.endRefreshing()
             headers.removeAll()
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR, this something we've wanted to fix for a while. However, your approach, while it works now, is still racy with the actual creation of the `URLRequest` used create the `description`. A more correct solution would be adding an `onURLRequestCreation` closure to the passed `Request` and using that value to set the title. Once that's changed, I think we can merge this.

TiagoMaiaL:
@jshier Thanks for your reply. I changed the branch to use the `onURLRequestCreation` event handler.

jshier:
Thanks!

PR #3502, opened 2021-10-03
Deprecate responseJSON

jshier:
### Goals :soccer:
Despite the extensive documentation, `responseJSON` seems to be greatly misunderstood. As I've observed in many code samples on StackOverflow and elsewhere, people seem to think you need to call it for every response that returns JSON, even if you then want to use `JSONDecoder` on the `Data` returned. Since this convenience is largely unneeded nowadays, it seems like a good idea to deprecate and point everyone to the much superior `responseDecodable`.

This PR simply deprecates the `responseJSON` methods and `JSONResponseSerializer`.

### Testing Details :mag:
This PR updates our tests to replace `responseJSON` with `responseDecodable`, except for tests which exercised `responseJSON` specifically, which have themselves been deprecated as well.

### DIFF
--- a/Documentation/AdvancedUsage.md
+++ b/Documentation/AdvancedUsage.md
@@ -301,7 +301,7 @@ AF.request(...)
     .downloadProgress { progress in
         print(progress)
     }
-    .responseDecodable(of: SomeType.self) { response in
+    .responseDecodable(of: DecodableType.self) { response in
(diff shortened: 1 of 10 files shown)


cnoon (approved):
Looks good @jshier, and completely makes sense to deprecate. 👌🏻

PR #3503, opened 2021-10-03
Fix the tests target build 

TiagoMaiaL:
### Goals :soccer:
Fix the build of the tests target. The code in the `ServerTrustEvaluatorTests.swift` isn't compiling due to the usage of  the `.revocation()` static accessor, which was renamed to `.revocationChecking()`.

### Testing Details :mag:
No tests were added.


jshier:
This was fixed in #3502, but thanks for looking out!

PR #3510, opened 2021-10-30
correct docs for DownloadResponse initializer

unknown-undefined:
improve docs


jshier:
Thanks for the PR, but this change is included in #3511, along with several others, so I'll be going with that one. Thanks anyway!

PR #3511, opened 2021-10-30
Fixes for parameters in doc comments throughout the project

evnik:
Noticed several warnings for "Mismatched doc comments and function signature"  inspection displayed by AppCode, so I run the inspection with "Whole project" scope and fixed all of the issues.

### DIFF
--- a/Source/Combine.swift
+++ b/Source/Combine.swift
@@ -201,17 +201,17 @@ extension DataRequest {
     /// response.
     ///
     /// - Parameters:
-    ///   - type:                `Decodable` type to which to decode response `Data`. Inferred from the context by default.
-    ///   - queue:               `DispatchQueue` on which the `DataResponse` will be published. `.main` by default.
-    ///   - preprocessor:        `DataPreprocessor` which filters the `Data` before serialization. `PassthroughPreprocessor()`
-    ///                          by default.
-    ///   - decoder:             `DataDecoder` instance used to decode response `Data`. `JSONDecoder()` by default.
-    ///   - emptyResponseCodes:  `Set<Int>` of HTTP status codes for which empty responses are allowed. `[204, 205]` by
-    ///                          default.
-    ///   - emptyRequestMethods: `Set<HTTPMethod>` of `HTTPMethod`s for which empty responses are allowed, regardless of
-    ///                          status code. `[.head]` by default.
+    ///   - type:                 `Decodable` type to which to decode response `Data`. Inferred from the context by default.
+    ///   - queue:                `DispatchQueue` on which the `DataResponse` will be published. `.main` by default.
+    ///   - preprocessor:         `DataPreprocessor` which filters the `Data` before serialization.
+    ///                           `PassthroughPreprocessor()` by default.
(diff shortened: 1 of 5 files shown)


Source/Validation.swift
```diff
@@ -253,9 +251,8 @@ extension DownloadRequest {
     ///
     /// If validation fails, subsequent calls to response handlers will have an associated error.
     ///
-    /// - Parameter statusCode: `Sequence` of acceptable response status codes.
-    ///
-    /// - Returns:              The instance.
+    /// - Parameter acceptableStatusCodes: `Sequence` of acceptable response status codes.
+    /// - Returns:                         The instance.
```
jshier, line 255: Please keep the extra line between the parameter and return.

Source/Validation.swift
```diff
@@ -200,9 +199,8 @@ extension DataStreamRequest {
     ///
     /// If validation fails, subsequent calls to response handlers will have an associated error.
     ///
-    /// - Parameter statusCode: `Sequence` of acceptable response status codes.
-    ///
-    /// - Returns:              The instance.
+    /// - Parameter acceptableStatusCodes: `Sequence` of acceptable response status codes.
```
jshier, line 202: Please keep the extra line between the parameter and return.

Source/Validation.swift
```diff
@@ -152,9 +152,8 @@ extension DataRequest {
...
```
jshier, line 155: Please keep the extra line between the parameter and return.

jshier (changes requested):
Thanks, this is very useful! Just a few style notes about the extra line between parameters and return, otherwise this looks good to merge.

Source/Combine.swift
```diff
@@ -571,17 +571,17 @@ extension DownloadRequest {
...
-    ///
-    /// - Returns:               The `DownloadResponsePublisher`.
+    ///   - type:                 `Decodable` type to which to decode response `Data`. Inferred from the context by default.
+    ///   - queue:                `DispatchQueue` on which the `DataResponse` will be published. `.main` by default.
+    ///   - preprocessor:         `DataPreprocessor` which filters the `Data` before serialization.
+    ///                           `PassthroughPreprocessor()` by default.
+    ///   - decoder:              `DataDecoder` instance used to decode response `Data`. `JSONDecoder()` by default.
+    ///   - emptyResponseCodes:   `Set<Int>` of HTTP status codes for which empty responses are allowed. `[204, 205]` by
+    ///                           default.
+    ///   - emptyResponseMethods: `Set<HTTPMethod>` of `HTTPMethod`s for which empty responses are allowed, regardless
```
jshier, line 581: This should actually be `emptyRequestMethods` to match the underlying function call. We can leave it for now or you can create a properly named method and deprecate this one, copying the docs from this method and then removing them. The deprecated method could then call through to the proper method.

jshier (approved):
👍 Looks good. Let me know if you want to fix the naming issue on the `publish` methods, otherwise this is good to merge.

jshier:
I'll take care of the publish rename.

PR #3512, opened 2021-10-31
Allow throwing from `multipartFormData` builders

### LINKED ISSUE #3493: Add throwing support to MultipartFormData builders

## Problem

When creating a multipart form data request, there is no place for custom encoding errors using Alamofire's APIs. Sometimes we need to convert images into data which can potentially return nil. In these scenarios it would be convenient to throw an error at this layer rather than wrapping `AFError` to use our own custom error type.

## Feature Request

Add `throws` to MultipartFormData building closures on the session's `upload` functions. This also requires adding a `case customEncodingFailed(error: Error)` to `MultipartEncodingFailureReason` for capturing any errors that are thrown via the closure.

## Value to Alamofire

I think this is something missing from Alamofire, as the other request / upload functions support `throws` via `ParameterEncoder`'s `encode` function. However, there is no equivalent mechanism for `MultipartFormData`.

evnik:
### Issue Link :link:
Fixes #3493

### Goals :soccer:
Allow error throwing when creating a multipart form data request in the `Session.upload` methods. Check #3493 for a usage example and more details on why does it make sense.

- Errors thrown by `multipartFormData` builder will be re-thrown by the `Session.upload` methods.
   *  This way allows to avoid any `try` statements if the builder doesn't actually throw, so this change shouldn't introduce any incompatibilities with existing code.
   * That means the error is not wrapped into the `AFError` as it was mentioned in #3493 
- `@escaping` attribution is removed since builder closure is called directly in the `upload` method.

### DIFF
--- a/Source/Session.swift
+++ b/Source/Session.swift
@@ -798,21 +798,21 @@ open class Session {
     ///                              provided parameters. `nil` by default.
     ///
     /// - Returns:                   The created `UploadRequest`.
-    open func upload(multipartFormData: @escaping (MultipartFormData) -> Void,
+    open func upload(multipartFormData: (MultipartFormData) throws -> Void,
                      to url: URLConvertible,
                      usingThreshold encodingMemoryThreshold: UInt64 = MultipartFormData.encodingMemoryThreshold,
                      method: HTTPMethod = .post,
                      headers: HTTPHeaders? = nil,
                      interceptor: RequestInterceptor? = nil,
                      fileManager: FileManager = .default,
-                     requestModifier: RequestModifier? = nil) -> UploadRequest {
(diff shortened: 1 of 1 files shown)


jshier:
Sorry, I misunderstood the original feature request as allowing the methods on `MultipartFormData` itself to throw. I don't see the value of this change, as you can just build your `MultipartFormData` outside the closure and make the request, which would actually be cleaner overall since you're not inside the closure anymore. Alamofire has specifically avoided top level methods which throw and provides the non-closure forms of multipart upload for exactly this reasons, so I don't think this change is needed at this time. Thanks anyway! 🍻

evnik:
@jshier Should #3493 be closed as well?

PR #3514, opened 2021-11-01
Sample app: display the initial loading state for each Request

TiagoMaiaL:
### Goal :soccer:
To display the initial loading state for each `Request` object in the `DetailViewController` of the sample application.

This change simply displays the `refreshIndicator` of the `UITableView`. For some reason the tint color isn't applied.
As an alternative, a `backgroundView` with an `UIActivityIndicator` can be added to the `UITableView` in order to display this initial loading state, with the proper color and layout position (centered).

### DIFF
--- a/Example/Source/DetailViewController.swift
+++ b/Example/Source/DetailViewController.swift
@@ -76,6 +76,7 @@ class DetailViewController: UITableViewController {
             return
         }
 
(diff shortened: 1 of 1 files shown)


jshier (approved):
👍 Makes sense, thanks!

PR #3516, opened 2021-11-05
Add .indexInBrackets array serialization option for URL parameters

TiagoMaiaL:
### Issue Link :link:
- #1839 
- #2431

### Goals :soccer:

There is no real specification for how to encode URL parameters like arrays and dictionaries.

For `["foo": ["a", 1, true, ["bar": 2], ["qux": 3], ["quy": ["quz": 3]]]]`

1. default `.brackets` encode is:
`foo[]=a&foo[]=1&foo[]=1&foo[][bar]=2&foo[][qux]=3&foo[][quy][quz]=3`

2. `.noBrackets` encode is:
`foo=a&foo=1&foo=1&foo[bar]=2&foo[qux]=3&foo[quy][quz]=3`

3. [jQuery](https://api.jquery.com/jQuery.param/#example-1) >=1.4 and Node.js package [query-string](https://www.npmjs.com/package/query-string) provide index style:
`foo[0]=a&foo[1]=1&foo[2]=1&foo[3][bar]=2&foo[4][qux]=3&foo[5][quy][quz]=3`

A new `ArrayEncoding` enumeration case is added to both `URLEncoding` and `URLEncodedFormEncoder`. The cases now are `.brackets` and `.noBrackets` and `.indexInBrackets`. The `arrayEncoding` option passed to the initializer is used to control the array key serialization in `URLEncodedFormEncoder` and `URLEncoding`.

### Testing Details :mag:

New tests were added to the `ParameterEncoderTests` and `ParameterEncodingTests` suites. They follow the same naming convention but with an  appended `WithIndexInBrackets` suffix.

### Note:

This PR continues the work started at #3015.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -85,13 +85,17 @@ public struct URLEncoding: ParameterEncoding {
         case brackets
         /// No brackets are appended. The key is encoded as is.
         case noBrackets
+        /// Brackets containing the item index appended. This matches the jQuery and Node.js behavior.
+        case indexInBrackets
 
-        func encode(key: String) -> String {
+        func encode(key: String, index: Int) -> String {
             switch self {
             case .brackets:
                 return "\(key)[]"
             case .noBrackets:
                 return key
+            case .indexInBrackets:
+                return "\(key)[\(index)]"
             }
         }
     }
@@ -193,8 +197,8 @@ public struct URLEncoding: ParameterEncoding {
                 components += queryComponents(fromKey: "\(key)[\(nestedKey)]", value: value)
             }
         case let array as [Any]:
-            for value in array {
-                components += queryComponents(fromKey: arrayEncoding.encode(key: key), value: value)
+            for (index, value) in array.enumerated() {
+                components += queryComponents(fromKey: arrayEncoding.encode(key: key, index: index), value: value)
             }
         case let number as NSNumber:
             if number.isBool {
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -85,10 +85,10 @@ public struct URLEncoding: ParameterEncoding {
         case brackets
         /// No brackets are appended. The key is encoded as is.
         case noBrackets
-        /// Brackets containing the item index appended. This matches the jQuery and Node.js behavior.
+        /// Brackets containing the item index are appended. This matches the jQuery and Node.js behavior.
         case indexInBrackets
 
-        func encode(key: String, index: Int) -> String {
+        func encode(key: String, atIndex index: Int) -> String {
             switch self {
             case .brackets:
                 return "\(key)[]"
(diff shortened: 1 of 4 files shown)


Source/ParameterEncoding.swift
```diff
@@ -85,13 +85,17 @@ public struct URLEncoding: ParameterEncoding {
         case brackets
         /// No brackets are appended. The key is encoded as is.
         case noBrackets
+        /// Brackets containing the item index appended. This matches the jQuery and Node.js behavior.
```
jshier, line 88:
```suggestion
        /// Brackets containing the item index are appended. This matches the jQuery and Node.js behavior.
```

Source/ParameterEncoding.swift
```diff
@@ -85,13 +85,17 @@ public struct URLEncoding: ParameterEncoding {
...
+        case indexInBrackets
 
-        func encode(key: String) -> String {
+        func encode(key: String, index: Int) -> String {
```
jshier, line 91: I'd suggest something like `encode(key: String, atIndex index: Int)` to better match Swift's naming guidelines to make the usage a bit more fluent.

Source/URLEncodedFormEncoder.swift
```diff
@@ -48,15 +48,19 @@ public final class URLEncodedFormEncoder {
...
         case noBrackets
+        /// Brackets containing the item index appended. This matches the jQuery and Node.js behavior.
+        case indexInBrackets
 
         /// Encodes the key according to the encoding.
         ///
-        /// - Parameter key: The `key` to encode.
-        /// - Returns:       The encoded key.
-        func encode(_ key: String) -> String {
+        /// - Parameter key:    The `key` to encode.
```
jshier, line 56: Now that there are multiple parameters, you can regenerate these docs using `Command - Option - /` and get a `Parameters` section instead. Also, though it was missing before, our general style is to put an empty line between the parameters and the return.

Source/URLEncodedFormEncoder.swift
```diff
@@ -48,15 +48,19 @@ public final class URLEncodedFormEncoder {
...
+        /// - Parameter index:  When this enum instance is `.indexInBrackets`, the `index` to encode.
+        /// - Returns:          The encoded key.
+        func encode(_ key: String, index: Int) -> String {
```
jshier, line 59: Like the other `encode`, `index` would be more fluently expressed as `atIndex`.

jshier (changes requested):
Thanks for the PR, this looks great! Just a few doc formatting suggestions and some naming and this should be all set.

Source/URLEncodedFormEncoder.swift
```diff
@@ -930,8 +936,8 @@ final class URLEncodedFormSerializer {
     }
 
     func serialize(_ array: [URLEncodedFormComponent], forKey key: String) -> String {
-        var segments: [String] = array.map { component in
-            let keyPath = arrayEncoding.encode(key)
+        var segments: [String] = array.enumerated().map { index, component in
+            let keyPath = arrayEncoding.encode(key, index: index)
```
jshier, line 940: Seems you've missed the use site for the updated name.

jshier (changes requested):
I think you missed one use site of the updated naming. Once that's updated I can trigger the test suite.

Source/URLEncodedFormEncoder.swift
TiagoMaiaL: Thanks for catching it, I forgot to build  🤦🏽‍♂️.

jshier (approved):
👍

kleverson:
Great work!

PR #3517, opened 2021-11-05
Deprecate publishDecodable variants with incorrect parameter name

jshier:
### Goals :soccer:
This PR deprecates the forms of `publishDecodable` which used a `emptyResponseMethods` parameter name, as it was supposed to be `emptyRequestMethods`.

Typical deprecation but can't use the `rename` variant because it was a parameter, not a method, that was renamed. Also used `@_disfavoredOverload` so that existing uses continue to compile.

### Testing Details :mag:
No testing changes, as they picked up the new overload correctly.

### DIFF
--- a/Source/Combine.swift
+++ b/Source/Combine.swift
@@ -197,22 +197,9 @@ extension DataRequest {
                         on: queue)
     }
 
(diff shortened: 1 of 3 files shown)


jshier:
Test failure likely due to httpbin.org flakiness (digest auth test redirects to httpbin because Firewalk doesn't support it natively).

PR #3518, opened 2021-11-06
Sort acceptable content-types in error messages

Usipov:
### Goals :soccer:

To make serialization error texts less random which comes in handy when you log those errors from you mobile apps. 

Currently I get errors like:
```
Request failed due to an underlying Alamofire error: Response Content-Type "image/webp" does not match any acceptable types: application/octet-stream,image/bmp,image/gif,image/heic,image/heif,image/ico,image/jp2,image/jpeg,image/jpg,image/png,image/tiff...
```
where acceptable types are listed in a random order. By making error texts more deterministic we get better compression on the errors storage and also a more reliable error counter

### DIFF
--- a/Source/Validation.swift
+++ b/Source/Validation.swift
@@ -117,7 +117,7 @@ extension Request {
             }
 
             let error: AFError = {
-                let reason: ErrorReason = .missingContentType(acceptableContentTypes: Array(acceptableContentTypes))
+                let reason: ErrorReason = .missingContentType(acceptableContentTypes: acceptableContentTypes.sorted())
                 return AFError.responseValidationFailed(reason: reason)
             }()
 
@@ -131,7 +131,7 @@ extension Request {
         }
 
         let error: AFError = {
-            let reason: ErrorReason = .unacceptableContentType(acceptableContentTypes: Array(acceptableContentTypes),
+            let reason: ErrorReason = .unacceptableContentType(acceptableContentTypes: acceptableContentTypes.sorted(),
                                                                responseContentType: responseContentType)
 
             return AFError.responseValidationFailed(reason: reason)
--- a/Tests/ValidationTests.swift
+++ b/Tests/ValidationTests.swift
@@ -247,6 +247,78 @@ final class ContentTypeValidationTestCase: BaseTestCase {
         }
     }
 
+    func testThatContentTypeValidationFailureSortsPossibleContentTypes() {
+        // Given
+        let endpoint = Endpoint.xml
+
+        let requestDidCompleteExpectation = expectation(description: "request should succeed and return xml")
+        let downloadDidCompleteExpectation = expectation(description: "download should succeed and return xml")
+
+        var requestError: AFError?
+        var downloadError: AFError?
+
+        let acceptableContentTypes = [ // Sorted in a random order, not alphabetically
+            "application/octet-stream",
+            "image/gif",
+            "image/x-xbitmap",
+            "image/tiff",
+            "image/jpg",
+            "image/x-bmp",
+            "image/jpeg",
+            "image/x-icon",
+            "image/jp2",
+            "image/png",
+            "image/ico",
+            "image/bmp",
+            "image/x-ms-bmp",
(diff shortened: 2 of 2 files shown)


Tests/ValidationTests.swift
```diff
@@ -221,14 +221,14 @@ final class ContentTypeValidationTestCase: BaseTestCase {
 
         // When
         AF.request(endpoint)
-            .validate(contentType: ["application/octet-stream"])
+            .validate(contentType: ["application/octet-stream", "application/json"])
```
jshier, line 224: A single test with two items is too likely to succeed randomly. Can you add a specific test, or update this test, to use many content types, like the image content types you mentioned? That should ensure the whole list is being alphabetized.

jshier (changes requested):
Thanks for the PR! Just a few notes, then I think this is good to go.

jshier:
> By making error texts more deterministic we get better compression on the errors storage and also a more reliable error counter

I'm not sure what you mean here. There's no compression on error storage (the strings are generated when you print their description) and I'm not sure what you mean by "error counter". In any case, the sorting should help with readability and figuring out whether your particular type is in the list or not.

Source/Validation.swift
```diff
@@ -131,7 +131,7 @@ extension Request {
         }
 
         let error: AFError = {
-            let reason: ErrorReason = .unacceptableContentType(acceptableContentTypes: Array(acceptableContentTypes),
+            let reason: ErrorReason = .unacceptableContentType(acceptableContentTypes: Array(acceptableContentTypes).sorted(),
```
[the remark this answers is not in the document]
Usipov: Good point!

Usipov:
> I'm not sure what you mean here. There's no compression on error storage (the strings are generated when you print their description) and I'm not sure what you mean by "error counter". In any case, the sorting should help with readability and figuring out whether your particular type is in the list or not.

We store all error texts in a database, group them by uniq text and then count. So, having lots of kinda different errors affects both the storage and the counter

Tests/ValidationTests.swift
jshier: I'm not sure why you marked this resolved, you didn't add any tests or make this test more complex.

Usipov:
@jshier, could you please revisit the PR? 🙏 
Do I understand it right, that after merging the PR you (not me) will release a new version?

Usipov:
> Workflow runs completed with no jobs
That's strange 🤔

Tests/ValidationTests.swift
```diff
@@ -247,6 +247,78 @@ final class ContentTypeValidationTestCase: BaseTestCase {
         }
     }
 
+    func testThatValidationForRequestWithUnacceptableContentTypeResponseFailsWhenMultipleContentTypesAreAcceptable() {
```
jshier, line 250: This test title doesn't really make sense, since this test is about ensuring the possible content types are sorted, not that the failure happens in the first place. I'd suggest `testThatContentTypeValidationFailureSortsPossibleContentTypes()`.

Tests/ValidationTests.swift
```diff
@@ -247,6 +247,78 @@ final class ContentTypeValidationTestCase: BaseTestCase {
...
+        // Given
+        let endpoint = Endpoint.xml
+
+        let expectation1 = expectation(description: "request should succeed and return xml")
```
jshier, line 254: These can be given better names. Perhaps `request1DidComplete` and `request2DidComplete`? And update the `description`s to indicate which request it's talking about.

jshier (changes requested):
Thanks for the better test. I have some naming notes, otherwise it looks good.

Usipov:
@jshier 🙏

jshier (approved):
👍 Looks good, thanks!

Usipov:
@jshier, thanks for the review!
Do I understand it right, that you (not me) will release a new version?

jshier:
Yes, this will go out with out next release. I'm aiming to do it around the time Apple releases Xcode 13.2.

Issue #3537, opened 2022-01-07
Concurrency is not compatible with Mac App Store / AppStoreConnect


frogg:
Uploading an app with version 5.5 is impossible (Error Message Attached).
On 5.4.4 it works fine.

I assume this is related to async / await.

Error Message:
```
This bundle is invalid. The executable code in [ICE Buddy.app/Contents/Frameworks/libswift_Concurrency.dylib] contain unsupported content. With error code STATE ERROR.VALIDATION ERROR.90000 for id 45011870-c31a-4c30-93e7-9c7d2014dbe0
```

![image](https://user-images.githubusercontent.com/5204169/148531399-40023507-0307-4dd4-9931-9f2e58c3b680.png)

jshier:
Yep, looks like Apple is still having issues with concurrency back deployment. This is very similar to #3529 and has the same fix: keep using 5.4.4 until Apple fixes it. You'll have to keep using older version of every library that updates to support concurrency.

jshier:
I heard word that the macOS deployment should [now be fixed](https://forums.swift.org/t/app-store-connect-does-not-accept-back-deploy-concurrency-macos-app-that-build-by-xcode-13-2-rc/53913/22). Can anyone confirm?

MariusDeReus:
@jshier, not sure if you mean this: but last week I still had issues when using Alamofire 5.5 and compiling on Xcode 13.2.1. Those builds crashed on macOS 10.14 and before, because the concurrency library cannot be found. 
I am using Cocoapods and have the app deployment target set to macOS 10.13. Apparently there is no way that Cocoapods can see that it needs to take Alamofire without the concurrency extension.

jshier:
@MariusDeReus No, this is a different issue. Your issue is #3529, and the only solution there is to keep using Xcode 13.1 or Alamofire 5.4.4 until either Apple issues a fix or we can separate out the concurrency feature into a subspec for Alamofire. This issue was simply about being able to submit to the App Store. It looks like that's fixed, so I'm closing this issue.

Issue #3529, opened 2021-12-20
Concurrency Crashes Using Xcode 13.2/.1 and Bitcode Issue on Xcode 13.3


LuckyPia:
Launch times for models below iOS13 ` libswift_ Concurrency.dylib (which was built for iOS 13.0) `, because '5.5.0' supports Concurrency, but only supports models above iOS 13. Please limit the usage environment of Concurrency

jshier:
It seems that iTunes Connect doesn't strip the concurrency binary out of app's deployed back to iOS 12. Stay on Alamofire 5.4.4 until that is fixed.

qinqi777:
I have run into the same issue, solved by pod 'Alamofire', '5.4.4'

jshier:
Pinning this until Apple ships a fix.

x2web:
Build with Xcode 13.1 to avoid crash on iOS 12  and earlier.

jshier:
To clarify: if you are deploying to a 2018 Apple OS or earlier (iOS 12, macOS 10.14, etc.) you have two options: remain on Alamofire 5.4.4 or Xcode 13.1. If you're targeting the 2019 OSes or newer, you shouldn't encounter this issue at all.

As for a fix, we're working to see if using CocoaPods subspecs and a separate SPM module may allow users to avoid the concurrency feature if they aren't using it. It's also possible Apple may fix this, we just can't know when.

jshier:
Looks like Xcode 13.3 beta 1 fixes multiple concurrency issues, including runtime issues on iOS 13 and 14, as well as this issue on older OSes. Hypothetically you could copy the concurrency compatibility library from Xcode 13.3 to your app built with Xcode 13.2.1 to also fix the issue.

Given that fact, I don't think we want to support a separate concurrency module or other workaround long term, so until Xcode 13.3 ships, everyone should keep using the existing workarounds. We may also want to consider moving the concurrency requirement to 5.6 when Xcode 13.3 ships so that users of the latest Alamofire version always avoid the broken version.

mharlowfod:
Confirmed 5.4.4 fixes issues on iOS 12 devices experienced with Xcode 13.2 and AF 5.5. Looking forward to an xcode fix soon if it really happens.

jshier:
This should be fixed in Xcode 13.3, which will likely be released next week after the Apple event.

kambala-decapitator:
[...]

PR #3544, opened 2022-01-16
Add Explicit Concurrency Subspec / Module / Build Flag

jshier:
### Issue Link :link:
#3529 and #3537

### Goals :soccer:
Due to Apple's ongoing issues around deploying the concurrency compatibility library to iOS 12 and below, this PR moves the recently added concurrency features to a separate CocoaPods subspec, SPM module, and Carthage build flag. (Not yet implemented.)

Subspec is easy, just removes the `Concurrency.swift` file from the default spec and adds it to the subspec.

This PR creates the `AlamofireConcurrency` submodule for SPM. Users must `import AlamofireConcurrency` to use the concurrency features.

Carthage build flag TBD.

### Testing Details :mag:
This PR also expands full testing to SPM so that it can test everything.


MariusDeReus:
Not sure, but is it possible to make this subspec depend on a minimum OS version. 
So for instance it will not be possible to add this dependency to your project when the deployment target of your project is below macOS 10.15?

jshier:
The subspec is the easy part, regardless of whether we can set a deployment target. If we can't, users will just have to avoid deploying back too far manually and it's an easy fix if they do. Right now the problem is SPM. There's no easy way to do what we need using SPM without exposing additional `public` API, so we'll have to experiment with some non-standard methods.

jshier:
Closing this, as Apple has fixed the issue in Xcode 13.3, so we don't want to introduce this just to not need it in a month.

PR #3548, opened 2022-01-20
Add Concurrency Example and Fix Documentation Link

### LINKED ISSUE #3540: "Session Configurations" link in "Usage" document points to non-existing part of the "Advanced Usage" document

## What did you do?

I'm trying to navigate to the "Session Configurations" section explaining the concept of "global" HTTP headers - there is a quote in the "Headers" section of `Usage.md`  document:

> For HTTP headers that do not change, it is recommended to set them on the `URLSessionConfiguration` so they are automatically applied to any `URLSessionTask` created by the underlying `URLSession`. For more information, see the [Session Configurations](AdvancedUsage.md#session-manager) section.

After clicking on the "Session Configurations" link, I'm navigated to the "Session Configurations" documentation

## What happened instead?

After clicking on the "Session Configurations" link, I'm navigated to the top of the `AdvancedUsage.md` document (the `#session-manager` anchor doesn't exist in that document...)

### LINKED ISSUE #3547: Allow multiple response handling with concurrency

## Problem

In the past, a common pattern used for debugging looks like this:

```swift
AF.request("url").responseDecodable(of: CodableValue.self) { response in
  [handle response, normal app code]
}
.responseString { response in print(response.value!) }
```

Note the trailing responseString, which would perform my normal response handling code but also print out the raw string from the server to the console. This is very useful when first implementing an API, so you can take the actual output from the server and read it, paste it into a linter, etc. When everything is working, you just comment out that last line. Then if something stops working, uncomment that line temporarily and you'll see the output again.

## Feature Request

In the new concurrency model, the code looks like this:

```swift
let outputValue = try await AF.request("url").serializingDecodable(CodableValue.self).value
[handle response, normal app code]
```

I see there is a `serializingString` function but it appears by using it, that would cause the request to run twice, which isn't desired.

## Value to Alamofire

This was a really useful function in Alamofire and it would be great to utilize it still while taking advantage of the new features of concurrency.

jshier:
### Issue Link :link:
Fixes #3547 and fixes #3540.

### Goals :soccer:
Adds an additional concurrency example and fixes a broken link.

### DIFF
--- a/Documentation/AdvancedUsage.md
+++ b/Documentation/AdvancedUsage.md
@@ -1249,7 +1249,7 @@ let request = AF.request(...)
 // Later...
 let stringResponse = await request.serializingString().response
 // Elsewhere...
-let decodableResponse = await request.serializingDecodable(TestResponse.self).self
+let decodableResponse = await request.serializingDecodable(TestResponse.self).response
 ```
 
 Finally, like all Swift Concurrency APIs, these `await`able properties can be used to `await` multiple requests issued in parallel. For example:
--- a/Documentation/AdvancedUsage.md
+++ b/Documentation/AdvancedUsage.md
@@ -1239,10 +1239,20 @@ let response = await task.response // Returns full DataResponse<TestResponse, AF
 // Elsewhere...
 let result = await task.result // Returns Result<TestResponse, AFError>
 // And...
-let value = try await task.value // Returns the TestResponse or throws the AFError
+let value = try await task.value // Returns the TestResponse or throws the AFError as an Error
 ```
 
-Like all Swift Concurrency APIs, these `await`able properties can be used to `await` multiple requests issued in parallel. For example:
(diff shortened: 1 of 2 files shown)


Documentation/AdvancedUsage.md
```diff
@@ -1239,10 +1239,20 @@ let response = await task.response // Returns full DataResponse<TestResponse, AF
...
 
-Like all Swift Concurrency APIs, these `await`able properties can be used to `await` multiple requests issued in parallel. For example:
+Similarly, and like Alamofire's existing closure and publisher-based response handlers, each request can produce multiple tasks that perform the same or different serializations.
+
+```swift
+let request = AF.request(...)
+// Later...
+let stringResponse = await request.serializingString().response
+// Elsewhere...
+let decodableResponse = await request.serializingDecodable(TestResponse.self).self
```
Vortec4800, line 1252: Is the `.self` on the end intended, or should it be `.response`?

jshier: Fixed, thanks.

PR #3562, opened 2022-02-21
corrected DataTask example

emanuelef:
### Issue Link :link:
DataTask example code uses task instead of the declared dataTask

### Goals :soccer:
Fix the DataTask example code

### Testing Details :mag:
None


jshier (approved):
Thanks.

PR #3573, opened 2022-03-10
Update `extension URLRequest` method in Document

Huang-Libo:
Refer: [URLRequest+Alamofire.swift](https://github.com/Alamofire/Alamofire/blob/master/Source/URLRequest%2BAlamofire.swift#L27-L39)


jshier (approved):

jshier:
Thanks!

PR #3589, opened 2022-04-08
Explicit Concurrency Module

jshier:
### Issue Link :link:
#3529

### Goals :soccer:
Due to ongoing Apple bugs, this PR moves the concurrency features into their own CocoaPods subspec or SPM module.

For CocoaPods, simply add the `Alamofire/Concurrency` subspec to the Podfile alongside or instead of the existing `Alamofire` pod.

For SPM, add the `AlamofireConcurrency` module as a dependency and add `import AlamofireConcurrency` wherever the async APIs are used.

This PR adds an SPM module by moving sources around, and adds the CP subspec by simply including the `Concurrency.swift` file.

### Testing Details :mag:
Testing in Xcode is unchanged, but SPM testing now excludes the concurrency work, as setting up that test target was too complex.


jshier:
Closing in favor of moving the concurrency features to Swift 5.6 now that Xcode 13.3.1 fixes the back deployment issues.

PR #3623, opened 2022-06-10
Fix typo in AdvancedUsage.md

ElonPark:
### Goals :soccer:

- Fix typo in AdvancedUsage.md
  - The description says that the default value is `false`, but the sample code is `true`.
스크린샷 2022-06-10 오후 6 43 21

### DIFF
--- a/Documentation/AdvancedUsage.md
+++ b/Documentation/AdvancedUsage.md
@@ -1285,7 +1285,7 @@ By default, `DataTask` and `DownloadTask` values do not cancel the underlying re
 ```swift
 let request = AF.request(...) // Creates the DataRequest.
 let task = Task { // Produces a `Task<DataResponse<TestResponse, AFError>, Never> value.
-    await request.serializingDecodable(TestResponse.self, automaticallyCancelling: true).response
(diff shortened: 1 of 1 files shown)


Documentation/AdvancedUsage.md
```diff
@@ -1285,7 +1285,7 @@ By default, `DataTask` and `DownloadTask` values do not cancel the underlying re
 ```swift
 let request = AF.request(...) // Creates the DataRequest.
 let task = Task { // Produces a `Task<DataResponse<TestResponse, AFError>, Never> value.
-    await request.serializingDecodable(TestResponse.self, automaticallyCancelling: true).response
+    await request.serializingDecodable(TestResponse.self, automaticallyCancelling: false).response
```
jshier, line 1288:
```suggestion
    await request.serializingDecodable(TestResponse.self).response
```

It's probably better to just remove the explicit parameter so readers don't think it's required.

ElonPark: Thanks!

jshier:
Thanks!

PR #3645, opened 2022-09-08
get activeRequests from Session

knottx:
### Goals :soccer:
because i need to sent log `activeRequests` when process in 
`func refresh(_ credential: Credential, for session: Session, completion: @escaping (Result<Credential, Error>) -> Void)` 
but in `Session` the `activeRequests` is an internal var.
e.g. when refresh complete or fail i need to log this event.


jshier:
There is already a `withAllRequests` closure, so this feature is unnecessary. Thanks anyway!

PR #3653, opened 2022-10-09
Move Swift Package to 5.7.

jshier:
### Goals :soccer:
This PR moves the Swift package to Swift 5.7 and adds a specific Swift 5.6 package file.


Kaspik:
Is this causing Xcode to throw new warnings because new Xcode does not support older versions supported by Alamofire?

Screenshot 2022-11-21 at 15 32 42

PR #3661, opened 2022-11-16
Refactor Unit tests inside the `RequestInterceptorTests.swift` File.

iHackSubhodip:
### Goals :soccer:
Reducing code repetition by introducing a helper method `makeSUT()` inside the `RequestInterceptorTests.swift` file.

### DIFF
--- a/Tests/RequestInterceptorTests.swift
+++ b/Tests/RequestInterceptorTests.swift
@@ -34,10 +34,7 @@ private struct RetryError: Error {}
 final class RetryResultTestCase: BaseTestCase {
     func testRetryRequiredProperty() {
         // Given, When
-        let retry = RetryResult.retry
-        let retryWithDelay = RetryResult.retryWithDelay(1.0)
-        let doNotRetry = RetryResult.doNotRetry
-        let doNotRetryWithError = RetryResult.doNotRetryWithError(MockError())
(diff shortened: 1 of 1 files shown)


jshier:
Thanks, but this isn't something we're interested in. We keep our tests a bit verbose to increase clarity when we revisit them or if we see failures. This sort of refactor doesn't achieve that goal.

iHackSubhodip:
@jshier Thanks for the insights. I'll be glad to contribute over the repo, so to get started, raised a small pr to understand the expectations. Please let me know, if there is any area where I can focus and contribute.

Also, Thanks for reviewing it 🙂

iHackSubhodip:
@jshier Thanks for sharing. For me, all of the above looks interesting. 🙂

As I see the topic no.1 is largely there, Should I go ahead and try to get it completed? Also, feel free to suggest if you've any other thought on picking a different topic.

One question, if I want to open up a discussion, where should I start, it's in github only or do we have any other platform i.e. Alamofire discussion section. Please let me know.

Best,
Subhodip

jshier:
A GitHub Discussion would be fine. If you'd like to look at websocket support, start a discussion with your summary of what you think the feature needs to be completed and I'll add my own thoughts and we can go from there.

iHackSubhodip:
@jshier sure, I'll have a look and give a try!

PR #3691, opened 2023-02-16
Fix Sendability Warnings in Xcode 14.3

jshier:
### Goals :soccer:
This PR fixes new `Sendable` warnings in Xcode 14.3.

This PR marks many Alamofire types as `@unchecked Sendable`, as they should be thread-safe but can't be determined so by the compiler for a variety of reasons.

### Testing Details :mag:
This is a compile time issue, so no tests. Adding these attributes should be a source compatible change.

### DIFF
--- a/Source/Concurrency.swift
+++ b/Source/Concurrency.swift
@@ -701,6 +701,7 @@ public struct StreamOf<Element>: AsyncSequence {
     }
 }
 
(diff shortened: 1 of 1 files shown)


jshier:
This was a bug in Xcode 14.3b1, fixed in beta 2, so closing this. We'll take another look at sendability once it's more required.

PR #3699, opened 2023-03-15
Add ArrayEncoding.custom to URLEncodedFormEncoder and ParameterEncoding.

naquin:
### Goals :soccer:
Let the user define a custom key formatting for arrays.

This helps for uncommon formats such as in the AWS API where it expects Array.1.Key or Array.member.1.Key. Rather than add specifics for those, providing a simple customization helps overcome this.

This follows other `.custom` conventions in the encoding options.

### Testing Details :mag:
Added a test under `ParameterEncodingTests.testThatArraysCanBeEncodedWithCustomClosure` that covers the change consistent with the other array encoding tests.

### DIFF
--- a/Source/ParameterEncoding.swift
+++ b/Source/ParameterEncoding.swift
@@ -87,6 +87,8 @@ public struct URLEncoding: ParameterEncoding {
         case noBrackets
         /// Brackets containing the item index are appended. This matches the jQuery and Node.js behavior.
         case indexInBrackets
+        /// Provide a custom array key encoding with the given closure.
+        case custom((String, Int) -> String)
 
         func encode(key: String, atIndex index: Int) -> String {
             switch self {
@@ -96,6 +98,7 @@ public struct URLEncoding: ParameterEncoding {
                 return key
             case .indexInBrackets:
                 return "\(key)[\(index)]"
+            case let .custom(encoding): return encoding(key, index)
             }
         }
     }
--- a/Source/URLEncodedFormEncoder.swift
+++ b/Source/URLEncodedFormEncoder.swift
@@ -60,6 +60,8 @@ public final class URLEncodedFormEncoder {
         case noBrackets
         /// Brackets containing the item index are appended. This matches the jQuery and Node.js behavior.
         case indexInBrackets
+        /// Provide a custom array key encoding with the given closure.
+        case custom((String, Int) -> String)
 
         /// Encodes the key according to the encoding.
         ///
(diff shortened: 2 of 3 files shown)


Source/URLEncodedFormEncoder.swift
```diff
@@ -60,6 +60,8 @@ public final class URLEncodedFormEncoder {
         case noBrackets
         /// Brackets containing the item index are appended. This matches the jQuery and Node.js behavior.
         case indexInBrackets
+        /// Provide a custom array key encoding with the given closure.
+        case custom((String, Int) -> String)
```
jshier, line 64:
```suggestion
        case custom((_ key: String, _ index: Int) -> String)
```

Internal labels make this a bit nicer to read.

Source/ParameterEncoding.swift
```diff
@@ -87,6 +87,8 @@ public struct URLEncoding: ParameterEncoding {
         case noBrackets
         /// Brackets containing the item index are appended. This matches the jQuery and Node.js behavior.
         case indexInBrackets
+        /// Provide a custom array key encoding with the given closure.
+        case custom((String, Int) -> String)
```
jshier, line 91:
```suggestion
        case custom((_ key: String, _ index: Int) -> String)
```

Internal labels make this a bit nicer to read.

Source/ParameterEncoding.swift
```diff
@@ -96,6 +98,7 @@ public struct URLEncoding: ParameterEncoding {
                 return key
             case .indexInBrackets:
                 return "\(key)[\(index)]"
+            case let .custom(encoding): return encoding(key, index)
```
jshier, line 101:
```suggestion
            case let .custom(encoding): 
                return encoding(key, index)
```
Follow the existing indentation.

Tests/ParameterEncoderTests.swift
```diff
@@ -624,6 +624,20 @@ final class URLEncodedFormEncoderTests: BaseTestCase {
         XCTAssertEqual(result.success, "array=1&array=2")
     }
 
+    func testThatArraysCanBeEncodedWithCustomClosure() {
+        // Given
+        let encoder = URLEncodedFormEncoder(arrayEncoding: .custom({ key, index in
+          "\(key).\(index+1)"
```
jshier, line 630:
```suggestion
          "\(key).\(index + 1)"
```
A little spacing for the index.

jshier (changes requested):
Looks good, definitely something to have. I've made some suggested edits. Can you add a test for the `ParameterEncoding` version? I'm not necessarily keeping that version in sync with everything, but any new features that are added should be tested.

naquin:
@jshier 
Thanks for the suggested edits, some nice polish there!

Added the unit test as requested.

jshier (approved):

PR #3704, opened 2023-03-30
Change flatMap to map

kati-kms:
According to Apple's Documents,
[map](https://developer.apple.com/documentation/swift/optional/map(_:)-7txtz) : "Use the map method with a closure that returns a non-optional value."
[flatMap](https://developer.apple.com/documentation/swift/optional/flatmap(_:)) : "Use the flatMap method with a closure that returns an optional value."

With the HTTPMethod changed from enum to structure(#2901), the init is no longer optional
So, in this case, the existing way works normally, but I think this way is more intended function utilization. 
(map guarantee non-optional)

Thank you


jshier:
Sure, thanks.

Issue #3719, opened 2023-05-10
5.7.0 Breaking change


DevSean:
* [*] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

ℹ Performed an iOS Unity build as usual.

## What did you expect to happen?

ℹ Build should have succeeded.

## What happened instead?

ℹ Build failed due to iOS Deployment change made within 5.7.0.
Compiling for iOS 10, but module 'Alamofire' has a minimum deployment target of iOS 11.0: 

This occurs due to a dependency from Moya:
https://github.com/CocoaPods/Specs/blob/master/Specs/8/a/7/Moya/14.0.0/Moya.podspec.json#L37

Which in turn is pulled in from our Advertisement sdk:
https://github.com/CocoaPods/Specs/blob/master/Specs/8/a/a/SuperAwesome/8.4.1/SuperAwesome.podspec.json#L39
https://github.com/CocoaPods/Specs/blame/master/Specs/6/3/6/IronSourceSuperAwesomeAdapter/4.1.5.1/IronSourceSuperAwesomeAdapter.podspec.json#L34

## Alamofire Environment

**Alamofire Version:** 5.7.0 (Was previously 5.6.4).
**Dependency Manager:** Cocoapods
**Xcode Version:** 14.3
**Swift Version:** N/A
**Platform(s) Running Alamofire:** iOS
**macOS Version Running Xcode:** 13.3.1 (a)

## Demo Project

ℹ N/A

jshier:
Yes, unfortunately the change was necessary due to Xcode 14.3 no longer compiling with an iOS 10 deployment target (#3716). You can either override your pod deployment targets using a `post_install` hook or lock your Alamofire version to 5.6.4 in your Podfile.

jshier:
Yeah, this is probably too big of a change to do right now. I didn't realize CocoaPods considered deployment targets breaking like that.

jshier:
5.7.1 has been released (I think, there was a CocoaPods error *after* I received the email about the release), which should fix this.

PR #3720, opened 2023-05-10
Revert Deployment Target Changes

jshier:
### Issue Link :link:
#3719

### Goals :soccer:
Unfortunately the deployment changes were more breaking than expected so this PR reverts them. Anyone having issues with libarclite should manually set their deployment target using a `post_install` hook, but that issue is much less common than this deployment target issue.


crsantos:
Hello, 

> Unfortunately the deployment changes were more breaking than expected

I'm afraid that this is fixed with a `post_install` for everyone that uses `Alamofire` as a dependency on a `Podfile`.

But this is not the same as having a pod with a dependency on it, as cocoapods needs to lint the spec, and iOS 10 is one of the deployment targets that is not valid on Xcode 14.3 anymore, every single `pod lib lint` will fail, as it needs iOS 11.0 and beyond, like you can see on this PR discussion: https://github.com/mxcl/PromiseKit/pull/1318#issuecomment-1587238635

So we have a chicken and egg problem here 😅 

Could a major version fix this? And we can finally drop iOS 10?

jshier:
iOS 10 (and that generation) is still valid, it just produces a warning, unlike iOS 8. For integrators, the `post_install` hook can take care of that, and for libraries with dependencies, passing `--allow-warnings` allows linting and pushing to pass normally with no changes to the pod itself. Any dependent library can also set its deployment target to any higher value, so Alamofire being low isn't a limitation.

The simple fact of the matter is that Apple will keep arbitrarily updating the minimum supported deployment target in Xcode whenever they want, and doing a major release to match every time that happens isn't tenable. If there's an actual breaking change, not simply a CocoaPod issue, then a major release may make sense.

crsantos:
Thank you so much for confirming @jshier, that's true, and I confirmed here with a lint, it's indeed a warning. 

Let me get from here 😄

PR #3728, opened 2023-05-23
Fixed the Xcode14.3 warning message for Heterogenous collection in `ParameterEncodingTests.swift`

Mayank-84:
- Fixed the warning in Xcode 14.3 `Heterogeneous collection literal could only be inferred to '[Any]'; add explicit type annotation if this is intentional`

### Goals :soccer:

This PR intends to fix the warning message that Xcode 14.3 generates in `ParameterEncodingTests.swift`

AlamofireSS

In case of heterogenous collection, compiler expects to explicitly specify the type to `[Any]`. So I added the type of parameter explicitly in the definition of the `parameters` variable.

### Testing Details :mag:

This just fixes minor compiler warnings. I don't think this requires any tests.

### DIFF
--- a/Tests/ParameterEncodingTests.swift
+++ b/Tests/ParameterEncodingTests.swift
@@ -153,7 +153,7 @@ final class URLParameterEncodingTestCase: ParameterEncodingTestCase {
 
     func testURLParameterEncodeStringKeyArrayValueParameter() throws {
         // Given
-        let parameters = ["foo": ["a", 1, true]]
+        let parameters: [String: [Any]] = ["foo": ["a", 1, true]]
 
         // When
         let urlRequest = try encoding.encode(urlRequest, with: parameters)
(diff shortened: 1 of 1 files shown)


Mayank-84:
@jshier do you have any suggestions for below image, shall I fix this as compiler suggests or is there any subtle way you think we can do this? 

Screenshot 2023-05-23 at 10 15 03 PM

jshier:
I'm on vacation so it'll be a couple weeks until I can fully review but I held off on these changes since the compiler change is going to be reverted. It doesn't look like that made it into a Swift 5.8 release, so we'll see if 5.9 at WWDC has it. If not, we can probably fix it.

Mayank-84:
Sure jshier.

jshier:
This was apparently [reverted in Xcode 14.3.1](https://forums.swift.org/t/swift-5-8-new-warnings-for-any-collections/63484/17), so this shouldn't be necessary any more. Thanks anyway!

PR #3737, opened 2023-06-17
Delete unused initializer in Protected

Huang-Libo:
### Issue Link :link:

N/A

### Goals :soccer:

Delete unused initializer 

N/A

### Testing Details :mag:

N/A

### DIFF
--- a/Source/Protected.swift
+++ b/Source/Protected.swift
@@ -91,10 +91,6 @@ final class Protected<T> {
     #endif
     private var value: T
 
-    init(_ value: T) {
-        self.value = value
-    }
-
     /// The contained value. Unsafe for anything more than direct read or write.
     var wrappedValue: T {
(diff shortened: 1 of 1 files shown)


Source/Protected.swift
```diff
@@ -91,10 +91,6 @@ final class Protected<T> {
     #endif
     private var value: T
 
-    init(_ value: T) {
-        self.value = value
-    }
```
Huang-Libo, line 96: We only used the `init(wrappedValue: T)` initializer of `Protected`

Huang-Libo:
@jshier 
Understood, but I think we also can use the `init(wrappedValue: T)` initializer if we don't want to use it as a property wrapper. 
Anyway, we can keep the `init(_ value: T)` initializer, but I suggest adding some comments about it, so users can better understand this property wrapper.

PR #3741, opened 2023-06-28
Updated to right usage of #if on Apple platforms

brennobemoura:
### Goals :soccer:

The goals here is to improve the right usage of `#if` statements on Apple platforms.

This #3738 can be improved to guarantee a better compilation conditions by using the correct framework related to each piece of code. After some revision, I found that Alamofire uses `!(os(Linux) || os(Windows))` which is not a great solution. So, depending on the context of each code updated, I change these `#if` to `canImport(Darwin)` or `canImport(Security)`.

Also, it isn't necessary to add on the `@available` and `#available` the `xrOS 1.0`. This is some Swift compiler bug since it chose the right branch. You can check [here](https://github.com/apple/swift-nio-ssl/pull/436#event-9605890831).

### Testing Details :mag:

I didn't implement any new feature.

### DIFF
--- a/Source/MultipartFormData.swift
+++ b/Source/MultipartFormData.swift
@@ -24,10 +24,12 @@
 
 import Foundation
 
+#if canImport(Darwin)
 #if os(iOS) || os(watchOS) || os(tvOS)
 import MobileCoreServices
-#elseif os(macOS)
+#else
 import CoreServices
+#endif
 #endif
 
 /// Constructs `multipart/form-data` for uploads within an HTTP or HTTPS body. There are currently two ways to encode
@@ -583,7 +585,7 @@ extension MultipartFormData {
     // MARK: - Private - Mime Type
 
     private func mimeType(forPathExtension pathExtension: String) -> String {
-        #if !(os(Linux) || os(Windows))
+        #if canImport(Darwin)
         if
             let id = UTTypeCreatePreferredIdentifierForTag(kUTTagClassFilenameExtension, pathExtension as CFString, nil)?.takeRetainedValue(),
             let contentType = UTTypeCopyPreferredTagWithClass(id, kUTTagClassMIMEType)?.takeRetainedValue() {
--- a/Source/Request.swift
+++ b/Source/Request.swift
@@ -1315,7 +1315,7 @@ public final class DataStreamRequest: Request {
 
     func didReceive(data: Data) {
         $streamMutableState.write { state in
-            #if !(os(Linux) || os(Windows))
+            #if canImport(Darwin)
             if let stream = state.outputStream {
                 underlyingQueue.async {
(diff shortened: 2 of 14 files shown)


Source/MultipartFormData.swift
```diff
@@ -24,11 +24,13 @@
 
 import Foundation
 
+#if canImport(Darwin)
```
jshier, line 27: This Darwin checks seems incorrect. Probably better to redo the OS check as a `canImport` for `MobileCoreServices` or `CoreServices.

Source/Request.swift
```diff
@@ -1315,7 +1315,7 @@ public final class DataStreamRequest: Request {
 
     func didReceive(data: Data) {
         $streamMutableState.write { state in
-            #if !(os(Linux) || os(Windows))
+            #if canImport(Darwin)
```
jshier, line 1318: These checks work around limitations in `swift-corelibs-foundation`, so a `Darwin` check isn't appropriate here. In the future these sorts of APIs should be supported across platforms with the `swift-foundation` rewrite.

Tests/AuthenticationInterceptorTests.swift
```diff
@@ -289,7 +289,7 @@ final class AuthenticationInterceptorTestCase: BaseTestCase {
 
     // MARK: - Tests - Retry
 
-    #if !(os(Linux) || os(Windows)) // URLRequest to /invalid/path is a fatal error.
+    #if canImport(Darwin) // URLRequest to /invalid/path is a fatal error.
```
jshier, line 292: Again, not an appropriate `Darwin` check.

Tests/NetworkReachabilityManagerTests.swift
```diff
@@ -22,7 +22,7 @@
 //  THE SOFTWARE.
 //
 
-#if canImport(SytemConfiguration)
+#if canImport(SystemConfiguration)
```
jshier, line 25: Can't tell what changed here.

jshier (changes requested):
Thanks for the PR, sorry it's taken so long to review. I think it will be useful. I agree with most of the framework checks but the `Darwin` checks, while technically matching Apple platforms, don't actually match the conditions we'd like to check for in most cases (missing or broken functionality in `swift-corelibs-foundation`), so I think most of those can be reverted.

Tests/NetworkReachabilityManagerTests.swift
jshier, line 25: Thanks for the typo fix. I think also think `visionOS` needs to be added to the platforms for the WWAN tests at the bottom.

jshier:
Looking to get this merged, so I've created #3756 to make the changes I suggested here. You'll still get credit.

PR #3742, opened 2023-07-02
fix: no need call empty initializer

Shinolr:

remove the explicitly call of empty initializer in HTTPHeaders


jshier:
True, thanks!

PR #3744, opened 2023-07-11
Added Android Support

hggz:
### Goals :soccer:

The goal of this pull request is to add Android support to Alamofire.

The following changes were verified on the swift-5.8 toolchain built for Android.

Alamofire should work flawlessly against an android swift toolchain (tested with version 5.8) with the same limitations and context as the current support on Linux + Windows.

### Testing Details :mag:

No new unit tests were added.

The following context + assumptions are in place for ensuring this works as expected:

- built as a swift package manager dependency on a swift 5.8 or above toolchain
- swift binary artifact is linked dynamically against an NDK (tested on r25c) android (api level 30) application
- the following gnu c libs must also be dynamically linked and loaded at runtime in your MainActivity
  - libcrypto
  - libssl
  - libz
 (feel free to grab these from the termux repo dev snapshots)

Tested on android api level 30, architectures: x86_64, arm64

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -301,10 +301,9 @@ jobs:
           if not exist .build\x86_64-unknown-windows-msvc\debug\Alamofire.swiftmodule exit 1
   CodeQL:
     name: Analyze with CodeQL
-    runs-on: macOS-12
+    runs-on: macOS-13
     env:
-      DEVELOPER_DIR: "/Applications/Xcode_14.2.app/Contents/Developer"
-      CODEQL_ENABLE_EXPERIMENTAL_FEATURES_SWIFT: true
+      DEVELOPER_DIR: "/Applications/Xcode_14.3.1.app/Contents/Developer"
     timeout-minutes: 10
     steps:
       - name: Clone
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -267,6 +267,14 @@ jobs:
       - uses: actions/checkout@v3
       - name: ${{ matrix.image }}
         run: swift build --build-tests -c debug
+  Android:
+    name: Android
+    uses: Swift-Commons/swift-android-sdk/.github/workflows/sdks.yml@ci
+    strategy:
+      fail-fast: false
+    with:
+      target-repo: ${{ github.repository }}
+      checkout-hash: ${{ github.sha }}
   Windows:
     name: Windows
     runs-on: windows-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,7 +273,7 @@ jobs:
       - name: ${{ matrix.image }}
         run: swift build --build-tests -c debug
   Android:
-    name: Android 
+    name: Android
     uses: hggz/swift-android-sdk/.github/workflows/sdks.yml@ci
     strategy:
       fail-fast: false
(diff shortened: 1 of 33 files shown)


jshier (commented):
Thanks for the PR! Is it possible to add an Android builder to our CI matrix? It would be build only for now but it would be good to have it alongside Linux and Windows.

hggz:
> Thanks for the PR! Is it possible to add an Android builder to our CI matrix? It would be build only for now but it would be good to have it alongside Linux and Windows.

sure, ill take a look. For reference I'm using this android toolchain's strategy for building an android friendly swift toolchain. I'm sure we can adopt this easily. https://github.com/finagolfin/swift-android-sdk/actions

hggz:
@jshier I pushed up a ci workflow for the matrix. I had to brush up on my github action knowledge but It should be pretty usable.

I implemented a workflow that builds a 5.8 android toolchain and builds alamofire from it at the current state. 

- The workflow should be pretty portable to use with other swift packages
- The workflow can be easily updated to leverage other builds of the android swift toolchain (currently using 5.8 release for arm64)
- By default, if no package is specified like how I'm doing now, it'll attempt to build the swift-argument-parser package.

I can help optimize it if needed but it should be good enough. If there are any recommendations about potentially storing the pre built toolchain somewhere to be fetched and shave off quite a bit of time of the workflow, I'd welcome them. I tried getting caching setup sort of how the original repo I posted above had but I couldn't get it to work reliably. I'm sure I can learn more about github action best practices to figure it out.

Anyway, let me know what you think

jshier:
This looks pretty good, but I think we want #3741 to get merged first so you don't need to touch so many `#if` checks. I trigger the CI for now to see how it turns out.

jshier:
@hggz Looks like the Android job you added is pretty slow. Can you refactor it to always use precompiled artifacts? It's not useful if it can't run as quickly as the other jobs.

hggz:
Sure thing, I'll update accordingly.

jshier:
Looks like the job was otherwise successful but took 42 minutes, which is much too long.

jshier:
Also looks like I need to update the platform identifier for visionOS. I'll do that in a separate PR.

jshier:
Also, the build actually failed but the job succeeded, so it needs to be configured to fail if any part of the process fails.

hggz:
Thank you, good catch. I'll verify the build once the prebuilt artifact
step is in place which will make that verification process a lot saner :)

jshier:
Sorry, I guess the build did succeed, just with many warnings. (Also seeing these on the nightlies on other branches, so I'll take a look.)

jshier:
I've merged the visionOS changes and some concurrency fixes, so you can rebase to pick up the changes.

hggz:
I fixed the CI pipeline. It now builds in sub 1 min 30 seconds instead of 40 min :) I cached the android toolchain I built as a release artifact and can use it readily now in the github action. 

I'll support it to make sure its up to date. Let me know if this works whenever you have a moment. 

Cheers

jshier:
@hggz I've updated this with the latest conditional changes and everything passes, so I'm going to merge it. Moving forward I'd appreciate it if you could build out additional versions of the Android builder for Swift 5.8, 5.9, etc, similar to how the Windows builders work. Additionally, it looks like your builder has some deprecation warnings from GitHub, so you'll probably want to fix those.

hggz:
Thanks Jon, apologies for the delay in getting those changes merged, I'll
be working on newer Android swift toolchains shortly.

PR #3754, opened 2023-08-01
feat: MultipartFormData ContentHeaderEncoding UTF-8 & iso-8859-1

apps4everyone:


jshier:
Thanks for the PR! Please follow the PR template to inform why you'd like to see this change and how applicable it is for all Alamofire users.

apps4everyone:
> Thanks for the PR! Please follow the PR template to inform why you'd like to see this change and how applicable it is for all Alamofire users.

Thx for the quick reply, just opened a issues https://github.com/Alamofire/Alamofire/issues/3755 to describe the problem

apps4everyone:
> I guess I don't understand the need here. UTF-8 is already used for the filename encoding we get the from the system, so why would a less capable encoding be necessary? And I'm not sure `cString` is an appropriate API to use here, as we aren't dealing with C strings in the first place.
> 
> Additionally, a change like this requires both updated documentation as well as tests. I can write the documentation if you can't, but tests are required as part of this change.

Hey @jshier thx for the feedback and sorry for the late reply. Currently it look's like we use use iso-8859-1 by default and not UTF-8 for the  MultipartFormData ContentHeaderEncoding. Maybe everywhere else but not for MultipartFormData. 🤔

Source/MultipartFormData.swift
```diff
@@ -510,9 +558,23 @@ open class MultipartFormData {
...
+        mimeType: String? = nil,
+        encoding: ContentHeaderEncoding?
+    ) -> HTTPHeaders {
         var disposition = "form-data; name=\"\(name)\""
-        if let fileName = fileName { disposition += "; filename=\"\(fileName)\"" }
+
+        if let fileName = fileName {
+            let encodingPrefix: String
+            if let encoding = encoding {
+                encodingPrefix = "*=\(encoding.rawValue)"
```
apps4everyone, line 572: GIST, optional encoding Info with one of:
*=UTF-8
*=iso-8859-1
default is without only
=

apps4everyone: it look's like its not the client bit the server, the server falls back to iso-8859-1 if there is only `=`.

apps4everyone: so the server uses the UTF-8 String bit only with iso-8859-1 char set so there are unknown characters. So so with *=UTF-8 we can tell the server explicitly that we are using *=UTF-8

PR #3756, opened 2023-08-09
Replace Some Platform Checks With canImport Checks

jshier:
### Issue Link :link:
This PR replaces #3741.

### Goals :soccer:
This PR builds on #3741 to replace many platform `#if os` checks with `#if canImport` for the relevant frameworks.


Mhudzmhudz:
Hello why I cannot commet

PR #3758, opened 2023-08-10
added comments for encode method

Rahul-Thengadi:


jshier:
This documentation should be inherited from the protocol itself and so doesn't need to be duplicated at each implementation point.

PR #3763, opened 2023-08-20
Update Windows CI

compnerd:
Adjust the command for building on Windows.  The installer should take care of the setup of the environment.  The actions setup already refresh the environment on behalf of the user.  This should allow us to avoid requiring resetting the environment.  The `cd` is implicit in the GHA execution.


jshier:
Nice, thanks!

compnerd:
@jshier I figure that you wouldn't mind a bit of simplification on the windows rules :)

PR #3779, opened 2023-09-19
Add encodeIfPresent Support to URLEncodedFormEncoder

### LINKED ISSUE #3778: NilEncoding always dropKey when encoding an Encodable object with optional properties

## What did you do?

I am using `URLEncodedFormEncoder` with the option `NilEncoding.dropValue` to encode an object that has optional string properties.

Assuming the following struct : 

```swift
struct MyStruct: Encodable {
    let p1: String
    let p2: String?
}
```

If I encode myStruct I expect to get the following query `p1=string&p2=`.

```swift
do {
    let myStruct = MyStruct(p1: "string", p2: nil)
    let formEncoder = URLEncodedFormEncoder(nilEncoding: .dropValue)
    let query: String = try formEncoder.encode(myStruct)
    print(query) // prints "p1=string"
} catch {
    print(error)
}
```

## What happened instead?

Got `p1=string`.

## Alamofire Environment

**Alamofire Version:** 5.7.1
**Dependency Manager:** Cocoapods 1.12.1
**Xcode Version:** Xcode 14.3.1
**Swift Version:** 5.7
**Platform(s) Running Alamofire:** iOS
**macOS Version Running Xcode:** Ventura 13.3.1

## Demo Project

Code above is sufficient to reproduce the issue.

## Suggestion

Digging into Alamofire I think the issue might by in `extension _URLEncodedFormEncoder.KeyedContainer: KeyedEncodingContainerProtocol {`.

In fact when the encoder found a nil value it seems to prefer using the `encodeIfPresent` method instead of `encodeNil`.

Using this rather naive implementation I managed to get my Struct encoded properly :

```swift
func encodeIfPresent(_ value: String?, forKey key: Key) throws {
    if let value {
        try encode(value, forKey: key)
    } else {
        guard let nilValue = nilEncoding.encodeNil() else { return }
        try encode(nilValue, forKey: key)
    }
}
```

jshier:
### Issue Link :link:
Fixes #3778

### Goals :soccer:
The encoding of synthesized `Encodable` values is different than `[String: String?]` so we were missing coverage for `encodeIfPresent`, causing `Optional` values to be dropped. This PR adds the missing support.

Added all of the `encodeIfPresent` overloads to `_URLEncodedFormEncoder.KeyedContainer` so that the protocol uses the custom implementation.

### Testing Details :mag:
Tests added for `Encodable` values with optional properties.

### DIFF
--- a/Source/URLEncodedFormEncoder.swift
+++ b/Source/URLEncodedFormEncoder.swift
@@ -695,6 +695,74 @@ extension _URLEncodedFormEncoder.KeyedContainer: KeyedEncodingContainerProtocol 
         try encode(nilValue, forKey: key)
     }
 
+    func encodeIfPresent(_ value: Bool?, forKey key: Key) throws {
+        try _encodeIfPresent(value, forKey: key)
+    }
+
+    func encodeIfPresent(_ value: String?, forKey key: Key) throws {
+        try _encodeIfPresent(value, forKey: key)
+    }
+
+    func encodeIfPresent(_ value: Double?, forKey key: Key) throws {
+        try _encodeIfPresent(value, forKey: key)
+    }
+
+    func encodeIfPresent(_ value: Float?, forKey key: Key) throws {
+        try _encodeIfPresent(value, forKey: key)
+    }
+
+    func encodeIfPresent(_ value: Int?, forKey key: Key) throws {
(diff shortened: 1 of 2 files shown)


sergheibilozor (commented):
😇

PR #3788, opened 2023-10-25
Fix visionOS Target Build Settings

Sai:
### Goals :soccer:
The build settings "Skip Install" is missing in visionOS target, which cloud cause the archive invalid when a project linked to the visionOS framework.

Set the "Skip Install" in visionOS target to YES

### Testing Details :mag:
No Test change.

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -2218,6 +2218,7 @@
 					"@loader_path/Frameworks",
 				);
 				SDKROOT = xros;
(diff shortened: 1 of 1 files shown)


jshier:
Thanks, this matches all the other targets. You'd think Apple's templates would be correct but I guess not.

wjacker:
Hi @jshier , seems 5.8.1 tag version name still is 5.8.0

jshier:
Dang it. I don't think that breaks anything, it's just a value for the User-Agent.

wjacker:
yep, minor effect but that will be fine.

Issue #3805, opened 2023-12-19
override AuthenticationInterceptor


cdoky:
> ℹ Please fill out this template when filing a feature request.
> All lines beginning with an ℹ symbol instruct you with what info we expect. You can delete those lines once you've filled in the info.
>
> Feature requests should code examples whenever possible.
>
> Per our [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md), we use GitHub for
> bugs and feature requests, not general support. Other issues should be opened on Stack Overflow with the tag `alamofire`.
>
> Please remove this line and everything above it before submitting.

* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## Problem

- This Interceptor is great, but there is a problem, it intercepts all the requests, so that when I set this blocker to global, as soon as 401 appears, there is no way to process any other requests. I found that by rewriting the adapt method without authorization checking, I could solve the problem

## Feature Request

- I just want to set the access so that I can override the corresponding method

## Value to Alamofire
```swift
public class OAuthenticationInterceptor: AuthenticationInterceptor<OAuthAuthenticator> {
    public override func adapt(_ urlRequest: URLRequest, for session: Session, completion: @escaping (Result<URLRequest, Error>) -> Void) {
        var authenticatedRequest = urlRequest
        if let credential = self.credential {
            authenticator.apply(credential, to: &authenticatedRequest)
        }
        completion(.success(authenticatedRequest))
    }
}
```
If we do not support the override, so, I have to make a copy of AuthenticationInterceptor, modify adapt method separately
```swift
// MARK: -

/// The `AuthenticationInterceptor` class manages the queuing and threading complexity of authenticating requests.
/// It relies on an `Authenticator` type to handle the actual `URLRequest` authentication and `Credential` refresh.
public class OAuthenticationInterceptor<AuthenticatorType>: RequestInterceptor where AuthenticatorType: Authenticator {
    // MARK: Typealiases

    /// Type of credential used to authenticate requests.
    public typealias Credential = AuthenticatorType.Credential

    // MARK: Helper Types

    /// Type that defines a time window used to identify excessive refresh calls. When enabled, prior to executing a
    /// refresh, the `AuthenticationInterceptor` compares the timestamp history of previous refresh calls against the
[...]

PR #3806, opened 2023-12-19
Feature: override AuthenticationInterceptor

cdoky:
You can override the adapt and retry method based on your business scenario

### Issue Link :link:

https://github.com/Alamofire/Alamofire/issues/3805#issue-2048551904

### DIFF
--- a/Source/AuthenticationInterceptor.swift
+++ b/Source/AuthenticationInterceptor.swift
@@ -157,7 +157,7 @@ public enum AuthenticationError: Error {
 
 /// The `AuthenticationInterceptor` class manages the queuing and threading complexity of authenticating requests.
 /// It relies on an `Authenticator` type to handle the actual `URLRequest` authentication and `Credential` refresh.
-public class AuthenticationInterceptor<AuthenticatorType>: RequestInterceptor where AuthenticatorType: Authenticator {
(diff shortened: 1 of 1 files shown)


cdoky:
@jshier Can you help for this PR?

Source/AuthenticationInterceptor.swift
```diff
@@ -221,7 +221,7 @@ public class AuthenticationInterceptor<AuthenticatorType>: RequestInterceptor wh
         set { mutableState.credential = newValue }
     }
 
-    let authenticator: AuthenticatorType
+    public let authenticator: AuthenticatorType
     let queue = DispatchQueue(label: "org.alamofire.authentication.inspector")
```
jshier, line 225:
```suggestion
    let queue = DispatchQueue(label: "org.alamofire.authenticationInterceptor")
```

jshier:
I was initially okay with this change, but properly subclassing `AuthenticationInterceptor` requires more than just opening up the method, it would need expose a way to safely access and mutate the internal state, which would then expose the `Protected` type we use for thread safety. That seems too much exposure for functionality that can also be accomplished by composing the existing `AuthenticationInterceptor` into a new `RequestInterceptor` that customizes the behavior you need, or by simply copying code. So I won't be accepting this PR, sorry.

PR #3822, opened 2024-01-29
Fix: uploadTask request body warning

siamakrostami:
### Goals :soccer:
- Fix a warning for uploadTask request body

- Just created a variable and copy the original request in it, then set the httpBody to nil

### Testing Details :mag:
- Tested on my personal project

### DIFF
--- a/Source/Core/UploadRequest.swift
+++ b/Source/Core/UploadRequest.swift
@@ -24,8 +24,48 @@
 
 import Foundation
 
+// MARK: - UploadRequest
+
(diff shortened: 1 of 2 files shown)


siamakrostami:
Screenshot 2024-01-29 at 17 43 01

jshier:
It doesn't seems like a good idea for Alamofire to silently rewrite user's requests to fix this error when the issue is with their code. The correct solution is for users to fix the issue on their side and ensure they aren't manually adding a body the `URLRequest` they're uploading. Thanks anyway!

PR #3823, opened 2024-01-29
fix: iOS v11 deprecation for SPM with tools version >=5.9

apps4everyone:
iOS v11 deprecation https://developer.apple.com/documentation/packagedescription/supportedplatform/iosversion/v11

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -27,8 +27,8 @@ import PackageDescription
 
 let package = Package(name: "Alamofire",
                       platforms: [.macOS(.v10_13),
-                                  .iOS(.v11),
-                                  .tvOS(.v11),
+                                  .iOS(.v12),
+                                  .tvOS(.v12),
                                   .watchOS(.v4)],
                       products: [
                           .library(name: "Alamofire", targets: ["Alamofire"]),
--- a/Package@swift-5.9.swift
+++ b/Package@swift-5.9.swift
@@ -27,8 +27,8 @@ import PackageDescription
 
 let package = Package(name: "Alamofire",
                       platforms: [.macOS(.v10_13),
-                                  .iOS(.v11),
-                                  .tvOS(.v11),
+                                  .iOS(.v12),
+                                  .tvOS(.v12),
                                   .watchOS(.v4)],
                       products: [
                           .library(name: "Alamofire", targets: ["Alamofire"]),


jshier:
This is only a warning when you're opening the package directly, not when integrating it into a project, so I'd like the package not to differ from the CocoaPod or project support even more than it already does. This is technically a major breaking change, as any package which depends on Alamofire would have to make this change as well or fail to build due to differing deployment targets. So thanks for the PR but we won't be accepting it.

apps4everyone:
Thank you but [Tuist](https://github.com/tuist/tuist) is using the package directly. Therefore I see those warnings.

apps4everyone:
Can you even build for iOS 11 when using Xcode 15 (tools version 5.9) ???

jshier:
You're going to have to be more specific; Tuist is a project generator, why would that interact with the Package any more directly than a normal project? If it's a warning from Tuist during project generation that's one thing, if it's a warning from the generated project that's another.

jshier:
> Can you even build for iOS 11 when using Xcode 15 (tools version 5.9) ???

Yes, of course, it's just a deployment value. Alamofire's project still supports iOS 10, and there are other projects out that that use the same deployment target, hence not bumping the value.

apps4everyone:
Xcode 15 is not even supporting iOS 11.
Therefore this is non breaking, as Xcode not even support's 11 anymore:
The Xcode 15 release supports on-device debugging in iOS 12 and later, tvOS 12 and later, and watchOS 4 and later. 
https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes

jshier:
Which is correct, but Alamofire supports more versions of Xcode than that, and code can still be deployed to those older OS versions. This is a breaking change, as other libraries depend on Alamofire and match its deployment target. I [tried it before](https://github.com/Alamofire/Alamofire/pull/3720).

apps4everyone:
but the Swift.package and Package@swift-5.9.swift only support's Xcode 15+ anyway

jshier:
Yes, but like I said, other packages may depend on Alamofire and moving the deployment target would break those packages. Doing so simply to fix a warning that most users will never see (you still haven't clarified where you're seeing it) is a bad idea. Breaking changes like this would need to be part of a major (v6) release, and I'm not going to do a major release every time Xcode feels like warning about a version of Xcode.

jshier:
That's incorrect. Like Alamofire, those packages can keep using the older deployment targets just fine, they may simply see the warning. Now, if you have any evidence that code built with Xcode 15 literally *can't run* on iOS 10 or 11, let me know, but I've never seen any evidence of that.

jshier:
So far we have a warning versus the guarantee that bumping the deployment target will break other packages. That's an easy choice for this library.

apps4everyone:
Thx, will try to find a prove and understand your point. But I still think it's technically not possible if you use SPM as they get resolved by the tools versions. That's why you keep the older with @swiftx.x.swift

apps4everyone:
seeing this https://developer.apple.com/forums/thread/735634

apps4everyone:
also here https://github.com/realm/realm-swift/issues/8368#issuecomment-1737604011

jshier:
Your first link is just a question. Your second isn't based on anything, it's just the same assertion you've been repeating. We know that Xcode 15 produces a warning with older deployment targets, but is there any evidence that it actually *doesn't work* when run on those older OSes? If so, I'd hope Xcode would produce an error instead of just letting you build for a platform that won't work.

jshier:
I am investigating whether SPM considers mismatching deployment targets a build error. If not we may be able to make the change anyway.

jshier:
It appears SPM implicitly promotes all older deployment targets to the lowest supported version anyway, so while target version mismatches for new targets produce an error, it doesn't matter for deprecated versions, so this would be a breaking change. My mistake, sorry about all the back and forth.

jshier:
Can you also bump the macOS and watchOS versions to match the iOS and tvOS generations?

apps4everyone:
> Can you also bump the macOS and watchOS versions to match the iOS and tvOS generations?

Thx and now worries, and sorry that I dit not add more info in the begining. My bad, sorry.

I think it's only iOS, others are looking ok:
https://developer.apple.com/documentation/packagedescription/supportedplatform/macosversion/v10_13
https://developer.apple.com/documentation/packagedescription/supportedplatform/watchosversion/v4

apps4everyone:
thx @jshier, one question: would it possible to make a new release with this fix the next week? thx

jshier:
I'm on vacation next week, so probably not. If I have downtime I may work on finalizing the `WebSocketRequest` documentation and tests I need, as I'd like to finally get that into 5.9, but I don't know how much time I'll have.

apps4everyone:
enjoy the vacation

PR #3832, opened 2024-02-16
chore: Maintain consistency in the doc

hugo-syn:
Maintain consistency in the doc like here: 

https://github.com/Alamofire/Alamofire/blob/5c68811d8b1eb4ac312a3968fa54e88a31a33abc/Source/Features/AuthenticationInterceptor.swift#L27


jshier (approved):
Thanks!

PR #3816, opened 2024-01-18
Make Empty conform to `Sendable`

AvdLee:
### Goals :soccer:
Make the `Empty` struct conform to `Sendable` to prepare for strict concurrency.

### Implementation Details :construction:
I was running into this warning while migrating our codebase for strict concurrency, related to `Empty` not being `Sendable`:
CleanShot 2024-01-18 at 10 57 09@2x

I could fix it for now using:

```swift
import Alamofire

extension Empty: @unchecked Sendable { }
```

But I'd obviously love to see this being added to the core library.

PR #3848, opened 2024-03-18
Make `HTTPMethod` Sendable

bestwnh:
### Goals :soccer:
Just like the https://github.com/Alamofire/Alamofire/pull/3816 but for `HTTPMethod`.
image

### DIFF
--- a/Source/Core/HTTPMethod.swift
+++ b/Source/Core/HTTPMethod.swift
@@ -26,7 +26,7 @@
 /// `HTTPMethod.get != HTTPMethod(rawValue: "get")`.
 ///
 /// See https://tools.ietf.org/html/rfc7231#section-4.3
-public struct HTTPMethod: RawRepresentable, Equatable, Hashable {
+public struct HTTPMethod: RawRepresentable, Equatable, Hashable, Sendable {
     /// `CONNECT` method.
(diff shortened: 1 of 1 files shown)


jshier:
Thanks. I'll audit the other simple types.

bestwnh:
@jshier Maybe it's time to begin the process to fully support concurrency? 👀

jshier:
Full support is impossible until Apple's frameworks themselves fully support it, but Alamofire is generally thread-safe, even if it can't always provide `Sendable` conformances. For the most part I can add `Sendable` support to Alamofire's simple types, and I already declare such support as `@unchecked` for the `Request` types (which are manually thread-safe through locks and controlled queue usage), but complex types like `AFError` can't truly be `Sendable` until the various system types it captures are themselves `Sendable`. In the meantime, an audit of the simple public types should be possible.

jshier (approved):

bestwnh:
Thanks for your explanation. I used to think that only UIKit is lack of concurrency support. Alamofire only use Foundation so I though it would be easier. I believe Alamofire is thread-safe. But I still want to kill all warning.

jshier:
Sure. Feel free to report any others but remember, you can use `@preconcurrency import Alamofire` to silence them in general.

PR #3858, opened 2024-04-05
Fix testThatDatesCanBeEncodedAsFormatted test fails due to current calendar mismatches with expected result

kittisak-phetrungnapha:
### Goals :soccer:
Currently test function `testThatDatesCanBeEncodedAsFormatted` fails when trying to run tests from simulator that is not using Gregorian calendar while the assertion is expecting date in Gregorian format. The goal should be running tests passed for all simulator calendar settings.

Force `DateFormatter` in the test function `testThatDatesCanBeEncodedAsFormatted` to be Gregorian so that it will be run tests passed for all simulator calendar settings.


jshier (approved):
Nice catch!

Issue #3766, opened 2023-08-24
Memory leak from RequestInterceptor.swift file


thinhnguyenext:
* [x] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?
I utilized the leak instrument and detected the memory leak originating from the function `adapt` in the file `RequestInterceptor.swift`. In fact, this function failed to add a weak reference in its closure.

![image](https://github.com/Alamofire/Alamofire/assets/143064266/8811e728-63fe-4c87-b1bd-6e799753f071)
![image](https://github.com/Alamofire/Alamofire/assets/143064266/98cca7ad-06e1-49a4-b55b-bb21c5f31d9d)

## What did you expect to happen?

I expected to add weak or unowned into the `adapt` function's closure, as this would prevent the memory leak from occurring.

## What happened instead?

It still retained the reference that caused the memory leak.

## Alamofire Environment

Alamofire Version: 5.7.1
Dependency Manager: SPM
Xcode Version: 14.3.1 (14E300c)
Swift Version: 5
Platform(s) Running Alamofire: iOS Simulator
macOS Version Running Xcode: 13.5.1 Ventura

jshier:
Can you provide an example project that reproduces this leak? There really shouldn't be a leak here, as all of the captures should be released once the nested adapt calls are complete.

kittisak-phetrungnapha:
@thinhnguyenext I add memory leak test function and apply it to `InterceptorTests`. It turn out that no memory leak has been found.
More detail: https://github.com/Alamofire/Alamofire/pull/3859

PR #3859, opened 2024-04-05
Add a test function to detect a memory leak if it it happened 

kittisak-phetrungnapha:
### Goals :soccer:
Follow up from [#3766](https://github.com/Alamofire/Alamofire/issues/3766), I create a test function to check that is `RequestInterceptor` has a memory leak or not. It's becoming this pull request because it could be used with other places as well.

- Add `checkMemoryLeaks` in `BaseTestCase` so that all subclassed test classes can use this function.
- Add checking memory leaks testing for all tested functions in `InterceptorTests` class.

Example of memory leaks (It's just fake from me):
leak

### DIFF
--- a/Tests/BaseTestCase.swift
+++ b/Tests/BaseTestCase.swift
@@ -133,7 +133,7 @@ class BaseTestCase: XCTestCase {
     ///   - object: The testing object for each class.
     ///   - file: Which file that failure test happens.
     ///   - line: Which line that failure test happens.
-    @available(iOS 13.0, *)
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
     func checkMemoryLeaks(_ object: AnyObject, file: StaticString = #file, line: UInt = #line) {
         addTeardownBlock { [weak object] in
             XCTAssertNil(object, "The object should be deallocated. Memory leaks may happen in the object.", file: file, line: line)
--- a/Tests/BaseTestCase.swift
+++ b/Tests/BaseTestCase.swift
@@ -126,4 +126,17 @@ class BaseTestCase: XCTestCase {
 
         waitForExpectations(timeout: timeout)
     }
+    
+    /// Check memory leaks for an object when a test operation has been finished.
+    ///
+    /// - Parameters:
+    ///   - object: The testing object for each class.
+    ///   - file: Which file that failure test happens.
+    ///   - line: Which line that failure test happens.
+    @available(iOS 13.0, *)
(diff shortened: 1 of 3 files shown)


Tests/BaseTestCase.swift
```diff
@@ -126,4 +126,17 @@ class BaseTestCase: XCTestCase {
...
         waitForExpectations(timeout: timeout)
     }
+    
+    /// Check memory leaks for an object when a test operation has been finished.
+    ///
+    /// - Parameters:
+    ///   - object: The testing object for each class.
+    ///   - file: Which file that failure test happens.
+    ///   - line: Which line that failure test happens.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
```
jshier, line 136: Can you add explicit watchOS and visionOS support as well? I think the other OSes are on their own.

Tests/RequestInterceptorTests.swift
```diff
@@ -563,6 +578,13 @@ final class InterceptorTests: BaseTestCase {
         XCTAssertEqual(result, RetryResult.doNotRetryWithError(RetryError()))
         XCTAssertTrue(result.error is RetryError)
         XCTAssertFalse(retrier2Called)
+        checkMemoryLeaksIfApplicable(interceptor)
+    }
+    
+    private func checkMemoryLeaksIfApplicable(_ object: AnyObject, file: StaticString = #file, line: UInt = #line) {
```
jshier, line 584: Seems like this should just live in the `BaseTestCase` to allow for all tests to use it. And given we can't test back before iOS 13 anyway, I'm not sure the `IfApplicable` is really necessary here.

jshier (changes requested):
Thanks, this seems like it could be useful. Just a couple small notes.

Tests/BaseTestCase.swift
kittisak-phetrungnapha: I have added watchOS but for visionOS I think there is no explicit need to be added.

```
@available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
@preconcurrency public func addTeardownBlock(_ block: @escaping @MainActor @Sendable () throws -> Void)
```

Tests/RequestInterceptorTests.swift
kittisak-phetrungnapha: Fixed

Vivi0074 (commented):
Please excuse my review! A1 overview

PR #3880, opened 2024-07-11
Swift 6 Native Builds

jshier:
### Goals :soccer:
This PR tracks the implementation of full concurrency checking when building in Swift 6 mode. It's unknown whether all of these changes will be fully backward compatible.

This PR largely consists of `Sendable` annotations, with a few logic changes to fix actual safety issues.

### Testing Details :mag:
Tests updated with needed annotations, but mutation warnings need a larger overall refactor.


gianfilippocisternino:
Any news on Swift 6 support?

scottasoutherland:
> Alamofire doesn't need to do anything to "support" Swift 6 or be integrated with Swift 6 projects. This PR just tracks the changes to make it build with Swift 6 internally. There are only a few things left in the library itself, mainly interactions with Apple types that aren't `Sendable`. (There are also warnings in the test suite, but I can deal with that later.) Feel free to try the branch and report any issues you find.

The problem i've run into, is that when integrating in my own project, many of the types provided are not Sendable. Usually you can get around this with @preconcurrency import. However for closure types, @preconcurrency import has no effect, and there is no way to silence the warnings. I can see in this branch you have fixed that, but full release of this branch is something that is blocking us from fully supporting complete concurrency/swift 6.

jshier:
Yes, this branch should fix that, in addition to making and requiring more things as `Sendable`. It also marks many APIs as `@preconcurrency` so these requirements don't break anything.

bbrk24:
Will AlamofireImage also get Swift 6 builds in the near future? I noticed Alamofire/AlamofireImage#478 (not quite the same, since that's about `@retroactive`, not Swift 6 _per se_), but no PR.

jshier:
Yes, AlamofireImage will get an update once I ship the Alamofire update.

jiyoungparki:
@jshier  Could you tell me when the version of Alamofire that supports Swift 6 will be released?

PR #3881, opened 2024-07-13
Fix existential-any for Swift6

kitwtnb:
### Goals :soccer:
Resolve `existential-any` build errors.

Added `any` to each protocols.  
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0335-existential-any.md


jshier:
Thanks! Though not strictly necessary (it's up to the project choose to enable this setting), I was still going to make the change as part of a Swift modernization pass.

PR #3886, opened 2024-08-07
Use MPTCP by default if available

Aperence:
### Goals :soccer:

Improve the network reliability of apps using Alamofire by enabling Multipath TCP by default, in Handover Mode. This
TCP extension allows to use multiple network interfaces in case of deterioration of the connection on one of them
Sources: 
- https://www.mptcp.dev
- https://developer.apple.com/documentation/foundation/urlsessionconfiguration/improving_network_reliability_using_multipath_tcp

Set the property `multipathServiceType` of the `URLSessionConfiguration` objects `default` and `ephemeral`

### Testing Details :mag:

Ran again the tests to make sure that everything was still working fine. No additional tests seems needed


jshier:
I don't think this is something we want, as it requires connection and server support to actually improve anything. There's a reason Apple doesn't enable this by default. It also requires an entitlement to work, so Alamofire enabling it by default does nothing if developers aren't entitled. Thanks anyway!

PR #3888, opened 2024-08-12
Do not read (potentially) large file during content type validation

jcavar:
### Goals :soccer:
- Avoid memory spikes during large file downloads
- Avoid failed downloads due to insufficient memory

Currently, the implementation of content type validation for `DownloadRequest` reads the whole file in memory to pass the data to a generic `Request` `validate`.

This can result in a huge memory spike or, even worse, an inability to read the file due to insufficient memory and failing requests.

There are a few options here:

- Use memory mapped file - this doesn't work in case of compressed files (zip)
- Read file metadata and check the file size
- Since the `Request` validator only checks if the data is empty, we could read a single byte from the file and pass it to the `Request` validator.

However, even if the response is empty, it seems incorrect to skip content type validation, so I  to simplify and remove the check completely.

Am I missing something here?

### DIFF
--- a/Source/Features/Validation.swift
+++ b/Source/Features/Validation.swift
@@ -94,10 +94,11 @@ extension Request {
 
     fileprivate func validate<S: Sequence>(contentType acceptableContentTypes: S,
                                            response: HTTPURLResponse,
-                                           data: Data?)
+                                           data: Data?,
+                                           skipIfEmpty: Bool = true)
         -> ValidationResult
         where S.Iterator.Element == String {
-        guard let data, !data.isEmpty else { return .success(()) }
+        if (data == nil || data!.isEmpty) && skipIfEmpty { return .success(()) }
 
         return validate(contentType: acceptableContentTypes, response: response)
     }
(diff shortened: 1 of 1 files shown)


jshier:
Thanks for the PR! Thinking about this I think we need to keep the same logic given existing users of the library. In the future we may reevaluate that choice, but for now we should maintain the behavior. We can always remove it later. Can you see if checking the URL attributes maintains the performance gain? You can use `try fileURL.resourceValues(forKeys: [.fileSizeKey])` and see if that gives you the information we need.

jcavar:
Hi @jshier, thank you for your response.

I understand concerns about backward compatibility. However, it still seems to me that empty responses should be validated.

How do you feel about the `skipIfEmpty` flag, which will default to `true` and will allow us (and anyone else) to override the existing behavior?

jshier:
Sure, though I'd still like to see if the `URL` check will for empty files in the first place.

jshier:
@jcavar Thanks for the update, but that's not really what I meant. Can you replace the `Data` initialization with the `URL` check I pointed out to see if it accomplishes your original goal? I'd like to make that optimization while still preserving the size check.

jcavar:
Hi @jshier, thank you for your comment. I meant to post an update, but I had to complete the testing.

As I see it, there are two issues:
1. Content type validation is skipped for empty files
2. Due to issue number 1, file size needs to be checked and that operation is inefficient

The goal of this PR is to address the first issue. And in such a case it also avoids reading file contents (since we don't need an empty file check anymore).

For the second issue, we don't have a use case where we are receiving empty files. So I can't test the change that you are proposing. And unfortunately, I don't have capacity to create reproducible example.

I believe this PR is a step in the right direction that maintains backward compatibility, but I am open to your thoughts.

jshier:
I've put up #3899, which reads the file attributes rather than the file itself to check the size, while keeping the other behaviors intact. That includes not validating the content type if the file is empty (no `fileURL` or `fileSize` is zero).

jcavar:
Thank you!

PR #3892, opened 2024-09-06
Added Response function documentation with example usage

herlandro:


jshier:
Thanks for the PR, but we don't generally document our private APIs, as it introduces a maintenance burden when we want to change them. Thanks anyway!

herlandro:
Ok, thanks for your replay. There is another place that I can help?

PR #3901, opened 2024-10-14
Set `@Sendable` on refresh method in Authenticator

rono23:
### Goals :soccer:

After upgrading to version 5.10.0, the following error occurred, so I tried to fix it with this PR.

image

### Environments

- Xcode: 16.0
- Swift: 6.0
- Strict Concurrency Checking: Complete


jshier:
Thanks!

jshier:
This has shipped with Alamofire 5.10.1.

PR #3920, opened 2024-11-24
Fix Validator Sendability

### LINKED ISSUE #3919: DataRequest content type validation for async request causes crash in Swift 6 mode
## What did you do?

Call of an asynchronous session request in a Swift concurrency task with content type validation via `DataRequest.validate(contentType:)`.

Code was compiled in Swift 6 language mode.
(See demo code below)

That the call does not cause a crash and works both when compiled in Swift 6 mode and when compiled with Swift 5.

## What happened instead?

`DataRequest.validate(contentType:)` causes a crash due to a Swift 6 concurrency assertion failure.

The stack trace is:
```
#0	in _dispatch_assert_queue_fail ()
#1	in dispatch_assert_queue ()
#2	in swift_task_isCurrentExecutorImpl ()
#3	in implicit closure #1 in closure #1 in ContentViewModel.makeRequest() ()
#4	in closure #1 in DataRequest.validate<[String]>(contentType:) at Alamofire/Source/Features/Validation.swift:175
```

Note that the exact same code compiled with the Swift 5 language mode does not cause a crash.
The underlying issue seems to be the use of non-sendable closures in the `DataRequest` implementations.

## Alamofire Environment

**Alamofire Version:** 5.10.1
**Dependency Manager:** Swift Package Manager
**Xcode Version:** 16.1 (16B40)
**Swift Version:** 6.0.2 (swiftlang-6.0.2.1.2 clang-1600.0.26.4)
**Platform(s) Running Alamofire:** iOS 18.1
**macOS Version Running Xcode:** macOS Sequoia 15.1

## Demo Project

```swift
@MainActor
func makeRequest() {
    Task {
        let session = Session()

        let urlRequest = URLRequest(
            url: URL(string: "https://raw.githubusercontent.com/Alamofire/Alamofire/refs/heads/master/Documentation/Usage.md")!
        )

        do {
            let result = try await session.request(urlRequest)
                .validate(contentType: ["application/json"]) // <-- this line causes a crash, removing it resolves the issue
                .serializingString()
                .response
                .result
                .get()

            print("Result: \(result)")
        } catch {
            print("Error: \(error)")
        }
    }
}
```

jshier:
### Issue Link :link:
Fixes #3919

### Goals :soccer:
This PR makes the validation system `Sendable`, fixing a runtime crash when the consuming app is built in Swift 6 mode.

Adds `@Sendable` to the validation closures, and ensures the returned `Error` is actually `Error & Sendable`.

### Testing Details :mag:
No additional tests.

### DIFF
--- a/Source/Core/DataRequest.swift
+++ b/Source/Core/DataRequest.swift
@@ -143,7 +143,7 @@ public class DataRequest: Request, @unchecked Sendable {
     @preconcurrency
     @discardableResult
     public func validate(_ validation: @escaping Validation) -> Self {
-        let validator: () -> Void = { [unowned self] in
+        let validator: @Sendable () -> Void = { [unowned self] in
             guard error == nil, let response else { return }
 
             let result = validation(request, response, data)
--- a/Source/Core/DataStreamRequest.swift
+++ b/Source/Core/DataStreamRequest.swift
@@ -199,7 +199,7 @@ public final class DataStreamRequest: Request, @unchecked Sendable {
     /// - Returns:              The `DataStreamRequest`.
     @discardableResult
     public func validate(_ validation: @escaping Validation) -> Self {
-        let validator: () -> Void = { [unowned self] in
+        let validator: @Sendable () -> Void = { [unowned self] in
             guard error == nil, let response else { return }
 
             let result = validation(request, response)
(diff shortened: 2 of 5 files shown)


dav3yblaz3 (commented):
looks like this could solve lots of issues thank you

PR #3929, opened 2025-01-04
Restore Linux and Android CI builds

marcprux:
### Goals :soccer:

This PR restores the Linux and Android build actions, which were disabled in https://github.com/Alamofire/Alamofire/pull/3920. It also updates the Android build action to use [swift-android-action@v2](https://github.com/marketplace/actions/swift-android-action), which enables it to run on Linux rather than requiring macOS.

Linux and Android toolchains cannot build test methods annotated with `@MainActor`, as discussed at https://github.com/swiftlang/swift-package-manager/pull/5525. Rather than disabling building the tests, we instead trim out the `@MainActor` declarations with a simple `sed` command prior to performing the build on those platforms.

### Testing Details :mag:
I didn't add or remove any tests. The builds at passing in my fork for [Android](https://github.com/marcprux/Alamofire/actions/runs/12614003719/job/35152459756) and [Linux](https://github.com/marcprux/Alamofire/actions/runs/12614003719/job/35152460366).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -330,6 +330,8 @@ jobs:
         run: swift build --build-tests -c debug
   Android:
     name: Android
+    strategy:
+      fail-fast: false
     runs-on: ubuntu-latest
     steps:
       - name: "Checkout"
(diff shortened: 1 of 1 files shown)


jshier:
Interesting, thanks. Not a huge fan of modifying the tests on other platforms, but probably fine until they're refactored to Swift Testing.

Issue #3937, opened 2025-02-18
URL conversion fails for URLs containing non-ASCII characters (e.g., Japanese characters)


SlavaPerryAyeKruchkovenko:
* [ true] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?
I tried to convert a string containing non-ASCII characters (Japanese characters) into a URL using the asURL() method provided by the URLConvertible protocol in Alamofire. The string I used was:

```let urlString = "https://www.testexample.com/サリーサルマガンディ"```
## What did you expect to happen?

I expected the asURL() method to successfully convert the string into a URL object, as the string is a valid URL when percent-encoded.

## What happened instead?

The asURL() method threw an ```AFError.invalidURL``` error because the string contains non-ASCII characters (Japanese characters), which are not valid in a URL without percent-encoding.

## Alamofire Environment

**Alamofire Version: 6.0**
**Dependency Manager: CocoaPods **
**Xcode Version: xcode 15.2**
**Swift Version: 5.9.2**
**Platform(s) Running Alamofire: iOS**
**macOS Version Running Xcode: ventura 13.6.7**

## Demo Project

```swift
try{
   let urlString = try "https://www.testexample.com/サリーサルマガンディ".asUrl()
} catch{
    print(error)
}
```

## Proposed Solution

I was able to resolve the issue by adding percent-encoding to the string before converting it to a URL. Here is the modified asURL() method:

```swift
extension String: URLConvertible {
    public func asURL() throws -> URL {
        guard let encodedString = self.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed),
              let url = URL(string: encodedString) else {
            throw AFError.invalidURL(url: self)
        }
        
        return url
    }
}
```
This solution works because it ensures that non-ASCII characters are correctly encoded before attempting to create a URL. In this case, the ASCII characters will remain unchanged.

jshier:
Thanks for the report! Feel free to put up a PR with this fix. However, instead of `addingPercentEncoding`, which applies the encoding to all parts of the string, `URLComponents` can be used, since it's `URL` aware. 

```swift
guard let components = URLComponents(string: self), let url = components.url else {
  throw AFError.invalidURL(url: self)
}

return url
```

If you would like to do that I can review and merge it.

SlavaPerryAyeKruchkovenko:
[...]

PR #3939, opened 2025-02-20
Fix URL conversion for URLs containing non-ASCII characters

SlavaPerryAyeKruchkovenko:
### Issue Link :link:
Fixes https://github.com/Alamofire/Alamofire/issues/3937

### Goals :soccer:
Add encode url with ```URLComponents``` for encode non ANSII characters 

Adds ```URLComponents``` into string extension method asUrl for encode url.
Also added a blank line check to pass the auto tests.

### Testing Details :mag:
Adds empty string check for complete test for session with incorrect url
Adds new test for convert string to url


jshier:
Thinking more about this issue, I'm pretty sure this fix would break existing code, as anyone who had percent escaped their `String` before converting it to a `URL` would see it doubly escaped. So we'll have to find a way to do this that's more backward compatible, or move the breaking change to Alamofire 6 (whenever that happens).

jshier:
Unfortunately, this would be a breaking behavioral change, and so can't be made in Alamofire 5. Once Alamofire 6 is underway we can make the escaping automatic.

PR #3947, opened 2025-03-23
Deprecate `NetworkReachabilityManager`

jshier:
### Issue Link :link:
Fixes #3943

### Goals :soccer:
Apple's `SCNetworkReachability` types were deprecated in iOS 17.4 related OS versions. This PR deprecates `NetworkReachabilityManager` in favor of `NWPathMonitor` from Network.framework, which has been the better solution for many years now.

This PR also updates various infrastructure bits, and deletes the `.swiftpm` directory to prevent Xcode from generating schemes when integrating the package.

### Implementation Details :construction:
Class deprecated.

### Testing Details :mag:
Tests deprecated.

PR #3948, opened 2025-03-24
Request caching when offline

gouravkmar:
Issue Link :link:
No specific issue was raised for this feature; this PR introduces a new functionality for caching failed requests and retrying them when the network becomes available.

Goals :soccer:
Automatically cache failed network requests and retry them once the network is reachable.
Improve handling of network connectivity issues, reducing the need for manual retries.

Implementation Details :construction:
A new CachedRequestManager class is introduced to cache failed requests and retry them when the network is reachable.
An extension on DataRequest adds a cacheAndRetry() method to handle caching and automatic retries of failed requests.

Usage :rocket:
Simply use the cacheAndRetry() method when making requests with Alamofire:

AF.request("https://example.com/api").cacheAndRetry()

Testing Details :mag:
Unit tests ensure failed requests are cached, retried, and cleared after successful retries.
Manual testing verifies that the retry mechanism works when toggling network availability.

### DIFF
--- a/Alamofire.xcodeproj/project.pbxproj
+++ b/Alamofire.xcodeproj/project.pbxproj
@@ -569,6 +569,11 @@
 		4CFB030E1D7D2FA20056F249 /* utf8_string.txt in Resources */ = {isa = PBXBuildFile; fileRef = 4CFB02F41D7D2FA20056F249 /* utf8_string.txt */; };
 		4CFB030F1D7D2FA20056F249 /* utf8_string.txt in Resources */ = {isa = PBXBuildFile; fileRef = 4CFB02F41D7D2FA20056F249 /* utf8_string.txt */; };
 		4DD67C251A5C590000ED2280 /* Alamofire.swift in Sources */ = {isa = PBXBuildFile; fileRef = F897FF4019AA800700AB5182 /* Alamofire.swift */; };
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR, we've wanted something like this! However, there's no need for custom caching here, Alamofire's retry feature already handles that. All you need to do here is create a custom `RequestRetrier`-conforming type and handle the retry logic in there, Alamofire can asynchronously wait to know whether to retry. 

Additionally, given the [imminent deprecation of `NetworkReachabilityMonitor`](https://github.com/Alamofire/Alamofire/pull/3947), you should rebuild on top of `NWPathMonitor`. That will ensure it works well into the future.

With those changes I can do a more detailed review and we can get this merged.

gouravkmar:
> Thanks for the PR, we've wanted something like this! However, there's no need for custom caching here, Alamofire's retry feature already handles that. All you need to do here is create a custom `RequestRetrier`-conforming type and handle the retry logic in there, Alamofire can asynchronously wait to know whether to retry.
> 
> Additionally, given the [imminent deprecation of `NetworkReachabilityManager`](https://github.com/Alamofire/Alamofire/pull/3947), you should rebuild on top of `NWPathMonitor`. That will ensure it works well into the future.
> 
> With those changes I can do a more detailed review and we can get this merged.

I've added the support of multiple failed requests in case of session wide retrier and removed the custom caching flow, please do check

gouravkmar:
@marcprux can you please help merge this in case this is fine?

PR #3970, opened 2025-08-01
feat:Add URL modification support for download resume data

voshill:
### Goals :soccer: 
Add URL modification support for download resume data

Parse the downloaded resumeData as a plist and modify URLs in the array corresponding to the $objects key in the plist. If oldURL is provided, it will modify exact matching strings; otherwise, it will modify all valid URLs.

### Testing Details :mag:
I'll add unit tests once the interface design is approved.


jshier:
Thanks for the PR, this is something we've thought about for a long time. However, before we get to specific APIs for modifying the resume data, we need to better model and test the parsing of the resume data blob. In this case, we need to write a `Codable` model which can be used to parse the blob, be modified, and written back into a blob. Additionally, extensive tests for each of the parsed properties, their mutation, and resuming downloads with modified resume data must be in place before more specific APIs are built. If you can get that done I can give a more thorough review before we consider what sort of convenience API we should build.

Issue #3870, opened 2024-05-15
EncodeIfPresent Doesn't Honor If An Optional is Present for URLEncodedFormEncoder


phoney:
* [ X] I've read, understood, and done my best to follow the [*CONTRIBUTING guidelines](https://github.com/Alamofire/Alamofire/blob/master/CONTRIBUTING.md).

## What did you do?

I expect that ```encodeIfPresent()``` when used with ```URLEncodedFormEncoder``` will NOT emit an optional value that is ```nil```. I expect that a simple ```encode()``` WILL include an optional value, even if it's nil, based on the ```nilEncoding``` value. I believe that's how JSONEncoder works but I don't think that's how ```URLEncodedFormEncoder``` works.

I have these two unit tests that I expect to pass. The ```JSONEncoder``` test passes but the URLEncodedFormEncoder test doesn't pass.

```
    private struct OptionalEncodableStruct2: Encodable {
        let one = "one"
        let two: String? = nil
        let three: String? = nil

        func encode(to encoder: any Encoder) throws {
            var container = encoder.container(keyedBy: CodingKeys.self)

            try container.encode(one, forKey: .one)
            try container.encode(two, forKey: .two)
            try container.encodeIfPresent(three, forKey: .three)
        }

        private enum CodingKeys: String, CodingKey {
            case one, two, three
        }
    }

    func testThatEncodeIfPresentDoesntEncodeNil() {
        // Given
        let encoder = URLEncodedFormEncoder(nilEncoding: .null)

        // When
        let result = Result<String, Error> { try encoder.encode(OptionalEncodableStruct2()) }

        // Then
        // XCTAssertEqual(result.success, "one=one")   // Cannot convert to string...
        switch result {
        case .success(let value):
            // two should be emitted, three should not be emitted
            XCTAssertEqual(value, "one=one&two=null")
            // XCTAssertEqual failed: ("one=one&three=null&two=null") is not equal to ("one=one&two=null").null
            // XCTAssertEqual failed: ("one=one&three=&two=") is not equal to ("one=one&two=") .dropValue
            // XCTAssertEqual failed: ("one=one") is not equal to ("one=one&two=null") .dropKey
        case .failure(let error):
            print(error)
            XCTFail()
        }
    }

    func testThatEncodeIfPresentDoesntEncodeNilJSONEncoder() throws {
        // Given
        let encoder = JSONEncoder()

        // When
        let result = try encoder.encode(OptionalEncodableStruct2())
[...]

PR #3977, opened 2025-09-10
feat: Add intentional nil encoding strategy to URLEncodedFormEncoder

darkbrewx:
## Issue Link 🔗

#3870 

## Goals ⚽

Enable different nil handling behaviors for `encode()` vs `encodeIfPresent()` in `URLEncodedFormEncoder`. Adds `.intentionalOnly` strategy that matches JSONEncoder semantics while maintaining full backward compatibility.

## Implementation Details 🚧

**Core Changes:**
- Extended `NilEncoding` with dual strategy support via `init(encode:encodeIfPresent:)` 
- Added `encodeNilIfPresent()` method and updated `KeyedContainer._encodeIfPresent()` to use it
- New `.intentionalOnly` strategy: encodes nil as "null" for `encode()`, skips for `encodeIfPresent()`

**Usage:**
```swift
let encoder = URLEncodedFormEncoder(nilEncoding: .intentionalOnly)

struct UserData: Encodable {
    let name: String? = nil
    let age: Int? = nil
    
    func encode(to encoder: any Encoder) throws {
        var container = encoder.container(keyedBy: CodingKeys.self)
        try container.encode(name, forKey: .name)        // → "name=null"  
        try container.encodeIfPresent(age, forKey: .age) // → skipped
    }
    
    enum CodingKeys: String, CodingKey { case name, age }
}
// Result: "name=null"
```

**Backward compatibility:** Existing `init(encoding:)` API unchanged.

## Testing Details 🔍

**New Tests:** 10 test cases in `IntentionalNilEncodingTests`
- ✅ `.intentionalOnly` strategy behavior
- ✅ Custom dual strategies  
- ✅ Array recursion and edge cases
- ✅ Backward compatibility

**Results:** All tests pass, existing URLEncodedFormEncoder tests unaffected.


Tests/IntentionalNilEncodingTests.swift
```diff
@@ -0,0 +1,247 @@
+//
+//  IntentionalNilEncodingTests.swift
+//
+//  Copyright (c) 2014-2018 Alamofire Software Foundation (http://alamofire.org/)
```
jshier, line 4: All new files should use the current year for copyright.

Tests/IntentionalNilEncodingTests.swift
```diff
@@ -0,0 +1,247 @@
...
+//
+
+import Alamofire
+import Foundation
+import XCTest
+
+final class IntentionalNilEncodingTests: BaseTestCase {
+    // MARK: - Test Helper Structures
+
+    struct TestStruct: Encodable {
```
jshier, line 32: Feel free to consolidate all of the test structs as `fileprivate`s at the bottom of this file.

Tests/IntentionalNilEncodingTests.swift
jshier: Really, though, these tests should be consolidated into the existing encoding tests.

PR #3988, opened 2025-11-04
[MultipartFormData] Allow append(fileURL:withName:) to infer extension from name parameter when URL lacks one

Rohan-cod-intouch:
### Goals :soccer:
- Allow files stored without extensions (e.g., in temporary directories) to be uploaded successfully if a valid filename with extension is provided through the name parameter.

- Maintain current behavior and compatibility for URLs that already include an extension.

Updated append(_ fileURL: URL, withName name: String) to:

- Attempt to infer MIME type from the file URL’s pathExtension, as before.

- If the URL lacks an extension, extract one from the name parameter using (name as NSString).pathExtension.

- Use that extension to derive the MIME type and proceed with file append.

- If neither contain an extension, preserve existing error behavior with .bodyPartFilenameInvalid(in:).

- Updated the doc comment to clarify the new fallback behavior and document the MIME type inference logic.

### Testing Details :mag:
None.

### DIFF
--- a/Source/Features/MultipartFormData.swift
+++ b/Source/Features/MultipartFormData.swift
@@ -166,7 +166,7 @@ open class MultipartFormData {
     /// `fileURL`. The `Content-Type` header MIME type is derived by mapping the file extension of the `fileURL` to its
     /// system-associated MIME type.
     ///
-    /// If the `fileURL`’s path does not contain a file extension, this method now attempts to extract an extension from the
+    /// If the `fileURL` does not contain a file extension, this method attempts to extract the extension from the
     /// provided `name` parameter instead. This allows uploading files stored without an extension when a filename with a
     /// valid extension is available in the request context.
     ///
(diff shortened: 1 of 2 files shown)


Source/Features/MultipartFormData.swift
```diff
@@ -162,9 +162,16 @@ open class MultipartFormData {
...
     /// - Multipart form boundary
     ///
-    /// The filename in the `Content-Disposition` HTTP header is generated from the last path component of the
-    /// `fileURL`. The `Content-Type` HTTP header MIME type is generated by mapping the `fileURL` extension to the
-    /// system associated MIME type.
+    /// The `filename` in the `Content-Disposition` HTTP header is generated from the last path component of the provided
+    /// `fileURL`. The `Content-Type` header MIME type is derived by mapping the file extension of the `fileURL` to its
+    /// system-associated MIME type.
+    ///
+    /// If the `fileURL`’s path does not contain a file extension, this method now attempts to extract an extension from the
```
jshier:
```suggestion
    /// If the `fileURL` does not contain a file extension, this method attempts to extract the extension from the
```

PR #4027, opened 2026-04-09
feat: add RequestDeduplicator to coalesce concurrent identical DataRequests

### LINKED ISSUE #2239: Request Deduplication
Some ability to deduplicate requests for the same resource would be quite powerful and could lead to some nice abilities. AlamofireImage currently does this, as well as Moya. We should be able to adapt the feature to something more general to be included with the base library.

VDurocher:
Closes #2239

## Problem

When multiple parts of an app simultaneously request the same resource — a common scenario on screens with several independent components — Alamofire creates a separate `URLSessionDataTask` for each call. The network is hit N times for identical data.

AlamofireImage has solved this for image downloads via its `ImageDownloader`. This PR brings the same capability to the base library in a general, type-safe way.

## Solution

`RequestDeduplicator` is a lightweight coordinator (~80 lines) that sits in front of `Session`. It tracks in-flight `DataRequest`s by a deduplication key and returns the **same `DataRequest` object** to every concurrent caller with the same key.

Because all callers hold a reference to the same object, each can attach its own independent response handlers — they all fire when the single network task completes. No extra queuing, no shared mutable response state.

```swift
let deduplicator = RequestDeduplicator()

// Only one URLSessionDataTask is created; both handlers receive the same response.
let req1 = deduplicator.request(using: session, "https://api.example.com/feed")
    .responseDecodable(of: Feed.self) { updateHomeScreen($0) }

let req2 = deduplicator.request(using: session, "https://api.example.com/feed")
    .responseDecodable(of: Feed.self) { updateWidgetScreen($0) }
```

## Design decisions

**No new protocol / no subclassing.** `RequestDeduplicator` is a standalone coordinator, not a `RequestInterceptor`. Deduplication is a *creation-time* concern — it must decide before a `DataRequest` exists whether to create one at all. The interceptor pipeline runs *after* creation, so it cannot serve this role.

**Thread safety via `Protected<State>` / `os_unfair_lock`.** Identical to `AuthenticationInterceptor` and `OfflineRetrier`. The `session.request()` call inside the write closure is safe because Alamofire dispatches its setup work asynchronously onto `Session.rootQueue`; the lock is never held across a blocking call.

**Automatic cleanup via a response handler.** A `.response { }` handler registered on each new request removes the in-flight entry when the request finishes or fails. This handler is registered *before* any caller-added handlers (which are added after `deduplicator.request()` returns), so the map is guaranteed to be cleared before callers' completions fire.

**Customisable key.** The default strategy is `"METHOD:absoluteURL"`. A `KeyProvider` closure lets callers widen (e.g. ignore query parameters) or narrow (e.g. include a header value) the deduplication boundary. Returning `nil` opts a specific request out entirely.

**Known limitation — cancellation.** Cancelling any one of the returned `DataRequest`s cancels the underlying task for all callers, since they all hold a reference to the same object. This matches AlamofireImage's behaviour and is documented in the API. Per-caller reference counting is left as a future enhancement.

## Files changed

| File | Change |
|------|--------|
| `Source/Features/RequestDeduplicator.swift` | New — implementation (~130 lines with doc comments) |
| `Tests/RequestDeduplicatorTests.swift` | New — 9 tests using Swift Testing, Given/When/Then |

## Tests

```
✔ defaultKeyProviderEncodesMethodAndAbsoluteURL
✔ concurrentRequestsForTheSameURLReturnTheSameDataRequest
✔ bothCallersReceiveASuccessfulResponseViaTheSharedRequest
✔ requestsForDifferentURLsAreNotDeduplicated
✔ requestsWithDifferentHTTPMethodsAreNotDeduplicated
✔ inflightEntryIsClearedAfterCompletionAllowingFreshDeduplication
✔ returningNilFromKeyProviderBypassesDeduplication
✔ customKeyProviderCanWidenDeduplicationBoundaryAcrossMethods
✔ urlResolutionFailureBypassesDeduplication
```

### DIFF
--- a/Source/Features/RequestDeduplicator.swift
+++ b/Source/Features/RequestDeduplicator.swift
@@ -1,0 +1,167 @@
+//
+//  RequestDeduplicator.swift
+//
+//  Copyright (c) 2026 Alamofire Software Foundation (http://alamofire.org/)
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+/// Coordinates concurrent `DataRequest`s for identical resources to avoid redundant network calls.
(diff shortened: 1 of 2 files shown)


jshier:
Thanks for the PR, but I'm afraid this just isn't the right approach. If I had wanted to port the AlamofireImage `DownloadManager`, I could've done so years ago. Instead, this needs to be something integrated into the request pipeline in some way.

If you're interested in talking through the design requirements of this feature, please open a GitHub discussion.

PR #4029, opened 2026-04-13
Add FreeBSD compilation support

hoanga:
### Issue Link :link:
Fix compile on FreeBSD

### Goals :soccer:
Fix compile on FreeBSD

### Testing Details :mag:

Ran swift build on FreeBSD

### DIFF
--- a/Source/Features/Combine.swift
+++ b/Source/Features/Combine.swift
@@ -22,7 +22,7 @@
 //  THE SOFTWARE.
 //
 
-#if !((os(iOS) && (arch(i386) || arch(arm))) || os(Windows) || os(Linux) || os(Android))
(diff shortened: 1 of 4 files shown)


hoanga:
I have tried an initial investigation for the build failure in 52aed37. While I do not think the extra case statements for adding os support are the trigger for this, I am also unable to completely rule out it is not. But one clue might be that I attempted to run the exact same test locally on the mac with this branch and the tests did not complete even after waiting quite some time which leads me to believe it might be CI environmental related (perhaps the watchOS proxy bootstrap isn't starting in the right timeframe?)

 I have pushed up a change to let the tests run further on FreeBSD but it seems that the tests will not run to completion due to other dependency support still needed in Swift.

PR #4036, opened 2026-05-06
fix: the httpheaders update and append methods in so... in...

orbisai0security:
## Summary
Fix high severity security issue in `Source/Core/HTTPHeaders.swift`.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | V-001 |
| **Severity** | HIGH |
| **Scanner** | multi_agent_ai |
| **Rule** | `V-001` |
| **File** | `Source/Core/HTTPHeaders.swift:46` |

**Description**: The HTTPHeaders update and append methods in Source/Core/HTTPHeaders.swift accept arbitrary string values for header names and values without stripping or rejecting CRLF sequences (\r\n). If application code passes user-controlled input directly into these methods, an attacker can inject additional HTTP headers by embedding newline characters in the header value, a technique known as HTTP header injection or HTTP response splitting.

## Changes
- `Source/Core/HTTPHeaders.swift`

## Verification
- [x] Build passes
- [x] Scanner re-scan confirms fix
- [x] LLM code review passed

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/Source/Core/HTTPHeaders.swift
+++ b/Source/Core/HTTPHeaders.swift
@@ -198,8 +198,8 @@ public struct HTTPHeader: Equatable, Hashable, Sendable {
     ///   - name:  The name of the header.
     ///   - value: The value of the header.
     public init(name: String, value: String) {
-        self.name = name
-        self.value = value
+        self.name = name.filter { $0 != "\r" && $0 != "\n" }
+        self.value = value.filter { $0 != "\r" && $0 != "\n" }
     }
 }
 


jshier:
First, this is NOT a security issue. Since Alamofire is built on top of `URLSession`, it inherits `URLSession`'s mitigations. That includes dropping header values with `\r\n` from `URLRequest.allHTTPHeaderFields`. While `HTTPHeaders` itself keeps the values around, round tripping through `URLRequest` drops the invalid values so they never reach the server. Additionally, `URLSessionConfiguration.httpAdditionalHeaders` also keeps the invalid value around, relying on `URLRequest` to drop the value, so there doesn't seem to be a need to mitigate this at the `HTTPHeaders` level as well.

Second, a report like this, if it was a real security issue, violates Alamofire's security reporting policy. @orbisai0security, if there's a human reading these messages, you CANNOT blindly report suspected security issues in the open like this, it violates all best practices around responsible reporting. In the future, please email reports to `security@alamofire.org`.

Third, the mitigation proposed in the PR is incorrect, or at least incomplete. As seen in `URLRequest`, suspect header values should be dropped entirely, ensuring any incorrect values aren't sent to the server at all.

orbisai0security:
Thanks for the detailed explanation, and apologies for the public security framing here.

I see your point that Alamofire ultimately goes through `URLSession` / `URLRequest`, and if invalid CRLF-containing header values are dropped before being sent on the wire, then this PR does not demonstrate a real header-injection vulnerability in Alamofire.

Also agreed that silently stripping `\r` / `\n` is not the right mitigation anyway, since it mutates caller input rather than rejecting or dropping the invalid header.

Issue #3073, opened 2020-02-23
DataPreprocessor doesn't have it's own failure reason


jshier:
Alamofire 5 added a `DataPreprocessor` protocol to enable the easy filtering of `Data` before processing, mainly to strip various headers in front of JSON. Unfortunately, when the protocol was added, no corresponding `ResponseSerializationFailureReason` was added, meaning the failure of a `DataPreprocessor` will be encapsulated in a `.customSerializationFailed` reason. Luckily, none of the current `DataPreprocessor`s produce errors, so we may never see this issue. 

Fixing this will require a major version change due to the addition of an enum case.

guptamohit1907:
I have added the fix in this PR 
https://github.com/Alamofire/Alamofire/pull/4038

PR #4038, opened 2026-05-22
Fix for DataPreprocessor doesn't have it's own failure reason #3073

guptamohit1907:
### Issue Link :link:
https://github.com/Alamofire/Alamofire/issues/3073

### Goals :soccer:
1. Add a dedicated AFError.ResponseSerializationFailureReason case for DataPreprocessor failures so they are no longer incorrectly surfaced as .customSerializationFailed.
2. Ensure all built-in serializers and stream serializers wrap preprocessing errors in the new specific error case.
3. Improve debuggability for consumers who implement custom DataPreprocessor types that can throw.

1. Added a new case dataPreprocessingFailed(error: any Error) to AFError.ResponseSerializationFailureReason in AFError.swift, along with updates to underlyingError, failedStringEncoding, and localizedDescription to handle it exhaustively.
2. In ResponseSerialization.swift, wrapped the dataPreprocessor.preprocess(data) call in a do-catch block inside all four built-in serializers (DataResponseSerializer, StringResponseSerializer, JSONResponseSerializer, DecodableResponseSerializer) so any thrown error is re-thrown as AFError.responseSerializationFailed(reason: .dataPreprocessingFailed(error: error)).
3. Applied the same fix to DecodableStreamSerializer.serialize(_:) in DataStreamRequest.swift, which had the same unguarded call.

### Testing Details :mag:
Updated all four existing "failing data preprocessor" tests in DataPreprocessorSerializationTests (Tests/ResponseSerializationTests.swift) to assert the error is specifically AFError.responseSerializationFailed(.dataPreprocessingFailed(...)) rather than just any non-nil failure. This covers DataResponseSerializer, StringResponseSerializer, JSONResponseSerializer, and DecodableResponseSerializer.

### DIFF
--- a/Tests/ResponseSerializationTests.swift
+++ b/Tests/ResponseSerializationTests.swift
@@ -1250,6 +1250,13 @@ final class DataPreprocessorSerializationTests: BaseTestCase {
         XCTAssertTrue(result.isFailure)
         XCTAssertNil(result.success)
         XCTAssertNotNil(result.failure)
+        XCTAssertTrue(result.failure?.asAFError?.isResponseSerializationError == true)
+        if case .responseSerializationFailed(let reason) = result.failure?.asAFError,
+           case .dataPreprocessingFailed = reason {
+            // Expected
+        } else {
+            XCTFail("Error should be AFError.responseSerializationFailed(.dataPreprocessingFailed), got: \(String(describing: result.failure))")
+        }
     }
 
     func testThatStringResponseSerializerProperlyCallsSuccessfulDataPreprocessor() {
@@ -1280,6 +1287,13 @@ final class DataPreprocessorSerializationTests: BaseTestCase {
         XCTAssertTrue(result.isFailure)
         XCTAssertNil(result.success)
         XCTAssertNotNil(result.failure)
+        XCTAssertTrue(result.failure?.asAFError?.isResponseSerializationError == true)
+        if case .responseSerializationFailed(let reason) = result.failure?.asAFError,
+           case .dataPreprocessingFailed = reason {
+            // Expected
+        } else {
+            XCTFail("Error should be AFError.responseSerializationFailed(.dataPreprocessingFailed), got: \(String(describing: result.failure))")
+        }
     }
 
     @available(*, deprecated)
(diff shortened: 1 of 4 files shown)


Tests/ResponseSerializationTests.swift
```diff
@@ -1312,6 +1326,20 @@ final class DataPreprocessorSerializationTests: BaseTestCase {
...
         XCTAssertNil(result.success)
         XCTAssertNotNil(result.failure)
+        XCTAssertTrue(result.failure?.asAFError?.isResponseSerializationError == true)
+        if case .responseSerializationFailed(let reason) = result.failure?.asAFError,
+           case .dataPreprocessingFailed = reason {
+            // Expected
+        } else {
+            XCTFail("Error should be AFError.responseSerializationFailed(.dataPreprocessingFailed), got: \(String(describing: result.failure))")
+        }
+        XCTAssertTrue(result.failure?.asAFError?.isResponseSerializationError == true)
```
jsingh6, line 1336: **[MEDIUM] Duplicate assertion block in JSON serializer preprocessor failure test**

The entire `XCTAssertTrue` + `if case .responseSerializationFailed` check block that validates `dataPreprocessingFailed` is pasted twice starting at this line. The first copy (lines 1329–1335) already validates the condition; this second copy is dead code that will never produce a different result. It is almost certainly a copy-paste mistake and should be removed.

_Category: logic error_

jsingh6 (commented):
This PR adds a dedicated `AFError.ResponseSerializationFailureReason.dataPreprocessingFailed(error:)` case so that `DataPreprocessor` failures are no longer incorrectly surfaced as `.customSerializationFailed`. The new case is wired up exhaustively in `AFError.swift` (`underlyingError`, `failedStringEncoding`, `localizedDescription`) and the do-catch wrapping is applied consistently across all four built-in serializers in `ResponseSerialization.swift` and `DecodableStreamSerializer` in `DataStreamRequest.swift`.

**Findings: 1 medium-severity issue (duplicate test assertions in the JSON serializer test case).**

jshier:
This change isn't possible without a major version break. When that occurs `AFError` will be fully refactored, so there's really nothing to do now.

